PR #38, opened 2022-03-04
Add support for Vala

lw64:


benpocalypse:
I know this PR has been open for awhile, and I don't want to add any noise for no reason - but I would _love_ to see Vala support added to Workbench. This app is **great** for doing quick Gtk work, and Vala support would really help me with my app development. That's it - cheers all!

lw64:
> I know this PR has been open for awhile, and I don't want to add any noise for no reason - but I would _love_ to see Vala support added to Workbench. This app is **great** for doing quick Gtk work, and Vala support would really help me with my app development. That's it - cheers all!

When it is finished it would be great to have the demos from the Library translated to vala :) Everyone who wants can help then.

benpocalypse:
> > I know this PR has been open for awhile, and I don't want to add any noise for no reason - but I would _love_ to see Vala support added to Workbench. This app is **great** for doing quick Gtk work, and Vala support would really help me with my app development. That's it - cheers all!
> 
> When it is finished it would be great to have the demos from the Library translated to vala :) Everyone who wants can help then.

I'll get to work on porting the demos from the Library ASAP. Great idea.

sonnyp (commented):
Things to do later:

* Adjust bookmarks for Vala and multi-lang future @sonnyp 
* Support opening `.vala` files @sonnyp 
* Translate demos @benpocalypse
* Probably refactor to have something like `Code` base class and `CodeJavaScript` / `CodeVala` - with `code.run`, `code.preview`, ... @sonnyp 
* make `compiler.compile` async @sonnyp 
* make dbus async @sonnyp 
* split icons into a separate resource - or move them to the demo resource @sonnyp 
* Disable dropdown option if example isn't available in the corresponding lang
* Edit .desktop file / readme / metainfo / github tags for Vala

src/demos/Welcome/main.vala
```diff
@@ -0,0 +1,11 @@
+Gtk.Builder builder;
+
+public void set_builder (Gtk.Builder b) {
+	builder = b;
+}
```
sonnyp, line 5: I'm ok with using `builder` instead of `workbench.builder` but is it possible to make it so that this isn't needed?

it's going to be confusing to beginners 

---

EDIT

We might also want to expose the preview window so actually

`workbench.builder` and `workbench.window` would be nice IMHO

src/Previewer.js
```diff
@@ -28,9 +30,12 @@ export default function Preview({
 
   let css_provider = null;
   let object_root = null;
+  let language = null;
+  let dbus_proxy = null;
+  let subprocess = null;
 
   preview_window_button.connect("clicked", () => {
-    if (!object_root) return;
+    if (!object_root || language !== "JavaScript") return;
```
sonnyp, line 38: wdyt of not opening the preview directly and instead showing the same thing as when previewing `GtkWindow` ?

image

This would solve 2 things:

1. Opening the window when loading the demo or clicking on run is a bit abrupt and confusing
2. There are currently 2 live previews

src/demos/Welcome/main.vala
```diff
@@ -0,0 +1,11 @@
...
+
+public void run () {
+	print("Hello World from Vala code!\n");
+	var label = builder.get_object ("text") as Gtk.Label;
+	//label.label = "Hi!";
+}
```
sonnyp, line 11: is it possible to pass an argument to run?

if so something like this would look great

```suggestion
public void run (workbench: Workbench) {
	print("Hello World from Vala code!\n");
	var label = workbench.builder.get_object ("text") as Gtk.Label;
	//label.label = "Hi!";
}
```

src/demos/Welcome/main.vala
```diff
@@ -0,0 +1,11 @@
+Gtk.Builder builder;
```
sonnyp, line 1: please translate the Welcome demo fully 

others can be done later

src/vala-previewer/previewer.vala
```diff
@@ -0,0 +1,111 @@
...
+
+        [CCode (has_target=false)]
+        private delegate void RunFunction ();
+
+        [CCode (has_target=false)]
+        private delegate void BuilderFunction (Gtk.Builder builder);
+
+        private Gtk.Window window = new Gtk.Window () {
+            hide_on_close = true
+        };
```
sonnyp, line 85: Please make the window for the preview default to 800x600 and a AdwWindow

src/vala-previewer/previewer.vala
```diff
@@ -0,0 +1,111 @@
...
+
+        public void update_ui (string content, string target_id) {
+            this.builder = new Gtk.Builder.from_string (content, content.length);
+            var target = this.builder.get_object (target_id) as Gtk.Widget;
+            if (target == null) {
+                stderr.printf (@"Widget with target_id='$target_id' could not be found.\n");
+                return;
+            }
+            this.window.child = target;
+        }
```
sonnyp, line 14: this doesn't support previewing GtkRoot object

if you prefer we can take care of it after this PR but please make sure you will take care of it or help out :pray:

sonnyp (changes requested):
Very cool ! 

I merged `main`, fixed the conflicts and made some changes – please have a look at them.

I just have a few requests and questions

src/vala-previewer/previewer.vala
```diff
@@ -0,0 +1,111 @@
...
+                              connection.register_object ("/re/sonny/workbench/vala_previewer", new Previewer ());
+                          } catch (IOError e) {
+                              stderr.printf ("Could not register service\n");
+                          }
+                      },
+                      (connection, name) => { stderr.printf ("Couldn't obtain the bus name.\n"); });
+
+        yield;
+    }
+}
```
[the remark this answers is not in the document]
lw64: My preferred coding style would be with 4 spaces, but 2 are totally fine. gnome-contacts does it too for example.

src/demos/Welcome/main.vala
lw64: I am planning to put that confusing boilerplate into a separate file, that gets compiled together, so you still can easily use them.

src/Previewer.js
lw64: I have not solved the case where you have a GtkRoot in the xml. At the moment that does not work, I think.

Though I think having a preview in-window is pretty convenient, even if you still have the external window open, but somewhere in the background. The in-window preview could also only appear when closing the external window though.

And what should "show preview window" do? running can only happen when the preview is visible.

src/demos/Welcome/main.vala
lw64: I think that cannot work, because run() is a dbus method and how would you pass a reference to workbench over dbus?

src/vala-previewer/previewer.vala
lw64: That would bean having a AdwHeaderbar in the demo, right?

src/vala-previewer/previewer.vala
lw64: A problem I see is that AdwWIndow can set the content only .content, and will crash otherwise.

src/Previewer.js
sonnyp: > I have not solved the case where you have a GtkRoot in the xml. At the moment that does not work, I think.

That's independent - my suggestion would apply for any kind of object

> Though I think having a preview in-window is pretty convenient

It's confusing and error-prone as the code will only affect one of them. It's really not an option as is.

> The in-window preview could also only appear when closing the external window though.

That's an interesting alternative.

Basically – opening the preview as window would only happen when clicking on "Run". When closiing the preview window - you can resume editing the XML and click on Run again to see the "end result".

I like it because it would simplify the implementation, as you wouldn't need to support real time updates on the out-of-process preview.

Is there a downside to it?

src/demos/Welcome/main.vala
sonnyp: workbench here wouldn't be the JavasScript object - it would be a Vala object with `builder` assigned to the Vala builder and `Window` to the vala winow

The idea is just to have the same "API" as JavaScript

src/vala-previewer/previewer.vala
sonnyp: AdwWindow supports window.title

src/vala-previewer/previewer.vala
sonnyp: When you get a GtkWindow to preview - you need to use it directly

src/vala-previewer/previewer.vala
lw64: But when you cannot specify the root, because it gets added by the previewer?

sonnyp (commented):
As discussed – please also add support for dark/light mode

Ho and if you could register `workbench.gresource.xml` in the vala proceess to avoid missing icons that would be great (I will split this into a separate resource later)

image

src/Previewer.js
sonnyp: > And since you have to click run after updating the preview anyway – is there a downside to it?

Yeah - when previewing a window - you now have 2 windows instead of 1.

That might get confusing

lw64:
For the missing icons, I think it would be best to put them into a shared library, so they don't need to be compiled as well every time.

src/Previewer.js
```diff
@@ -97,20 +118,31 @@ export default function Previewer({
       return;
     }
 
+    dbus_proxy.DarkMode = style_manager.dark;
+
```
sonnyp: this doesn't work - dbus_proxy is null at this point in time.

I'll proceed with merging anyway - appreciate if you can send a follow up PR :pray:

sonnyp (approved):
Well done!

I have minor additional changes https://github.com/sonnyp/Workbench/pull/38/commits/ccce44a7230813a97e134e3a792414bddffcca51

Celebration: https://floss.social/@sonny/108319817839890501
Some follow ups: https://github.com/sonnyp/Workbench/pull/38#pullrequestreview-973010088

PR #76, opened 2022-05-19
Fix up Vala support

lw64:


src/Library/demos/Welcome/main.js
```diff
@@ -1,6 +1,7 @@
 import Gtk from "gi://Gtk?version=4.0";
 
 Gtk.init();
+Adw.init();
```
sonnyp, line 4: `Adw.init()` already calls `Gtk.init()` so it should be one or the other.

This particular demo doesn't use `Adw`

lw64: That is right. But the style is different then.

sonnyp: Should we always use `Adw.init()` ?

sonnyp:
https://github.com/sonnyp/Workbench/commit/4d029ab2d65d1aca4b234d946c659ef16564ede6

PR #92, opened 2022-05-21
Use template from Code

sonnyp:
Fixes https://github.com/sonnyp/Workbench/issues/100

https://jwestman.pages.gitlab.gnome.org/blueprint-compiler/examples.html#referencing-a-template-object

```
using Gtk 4.0;


template MyAppWindow : ApplicationWindow { }

Gtk.Label {
  visible: bind MyAppWindow.visible;
}
```


src/Previewer/Previewer.js
```diff
@@ -90,11 +90,13 @@ export default function Previewer({
     let text = panel_ui.xml.trim();
     let target_id;
     let tree;
+    // let template;
 
     try {
       tree = ltx.parse(text);
-      [target_id, text] = targetBuildable(tree);
+      [target_id, text /* template */] = targetBuildable(tree);
```
sonnyp, line 97: we need to expose `workbench.template` for `Internal.js`

for `External.js` we probably need to write this to disk

Issue #93, opened 2022-05-21
Support connecting Builder signal handlers


sonnyp:
UI definitions with signal handlers can now be previewed - see https://github.com/sonnyp/Workbench/pull/94

But users should be able to define handlers in `Code` that get automatically connected.

PR #94, opened 2022-05-21
Add support for previewing with signals

sonnyp:
Related https://github.com/sonnyp/Workbench/issues/93


src/Previewer/External.js
```diff
@@ -21,10 +21,6 @@ export default function Previewer({ builder, onWindowChange }) {
   })();
 
   const dbus_proxy = DBusPreviewer();
-  dbus_proxy.connectSignal("Ready", () => {
-    updateColorScheme();
-  });
-
```
sonnyp, line 27: not used

PR #101, opened 2022-05-24
Translate library examples to Vala

benpocalypse:
There are some shortcomings due to the Vala API not having access to the `application` class, but I think these translations are otherwise fairly faithful to the originals. Feel free to comment and criticize though.


src/Library/demos/Toasts/main.vala
```diff
@@ -0,0 +1,37 @@
...
+    toast.dismissed.connect (() => {
+      button_simple.set_sensitive (true);
+    });
+    overlay.add_toast (toast);
+    button_simple.set_sensitive (false);
+  });
+
+  var button_advanced = workbench.builder.get_object("button_advanced") as Gtk.Button;
+  button_advanced.clicked.connect(() => {
+    var message_id = "42";
```
lw64, line 19: please use ```int``` instead of ```var``` here.

src/Library/demos/Toasts/main.vala
```diff
@@ -0,0 +1,37 @@
...
+  var overlay = workbench.builder.get_object ("overlay") as Adw.ToastOverlay;
+
+  var button_simple = workbench.builder.get_object ("button_simple") as Gtk.Button;
+  button_simple.clicked.connect (() => {
+    var toast = new Adw.Toast ("Toasts are delicious!") {
+      timeout = 1
+    };
+
+    toast.dismissed.connect (() => {
+      button_simple.set_sensitive (true);
```
lw64, line 11: ```button_simple.sensitive = true;```

src/Library/demos/Toasts/main.vala
```diff
@@ -0,0 +1,37 @@
...
+      action_name = "win.undo",
+      action_target = new GLib.Variant.string (message_id),
+      priority = Adw.ToastPriority.HIGH,
+    };
+    overlay.add_toast (advanced_toast);
+  });
+
+  var action_console = new SimpleAction ("undo", new GLib.VariantType ("s"));
+  action_console.activate.connect ((self, target) => {
+      var val = target.get_string ();
```
lw64, line 31: here the type should be also written out, and not just ```var```.

lw64 (changes requested):

lw64:
About the missing Application instance, I think a SimpleActionGroup could be used instead. It should be probably provided by the previewer in the ```workbench``` interface.

Also I think the websocket demo could be object oriented, so that you have a ```WorkbenchWebSocketClient``` or similar that handles the libsoup stuff.

src/Library/demos/Toasts/main.vala
benpocalypse: This is actually a `string` although I could change it to an `int` and then use an `int.parse ()` on line 42, but that seems a tad redundant. Ok to just change it to `string` on line 19 here?

src/Library/demos/Toasts/main.vala
```diff
@@ -0,0 +1,37 @@
...
+      stdout.printf ("undo" + val);
```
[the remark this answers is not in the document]
benpocalypse: Indeed there is. Good catch.

src/Library/demos/Toasts/main.vala
lw64: Yeah. I lookedntonquickly over it :?/)

sonnyp (commented):
the Adw.Application is now exposed to `workbench.application` thanks to @lw64 https://github.com/sonnyp/Workbench/pull/102

sonnyp:
Also - please feel free to add yourself to the list of contributors in `src/about.js`

benpocalypse:
Yay!
Thanks...I'll get to the changes today.

benpocalypse:
I updated my branch to make use of the newly exposed `application` but the end effect still isn't what is desired. The notifications demo exposes an assertion that fails...not sure what the problem is. Feel free to bonk me on the head.

lw64:
The notification demo is not working currently. Can you split it into a separate commit, that can be later merged, when notifications will work? The erst can be merged I think.

benpocalypse:
> The notification demo is not working currently. Can you split it into a separate commit, that can be later merged, when notifications will work? The rest can be merged I think.

Sure. I'll back out the notifications demo and then we'll see if we can't get this PR merged.

sonnyp (approved):
Thank you !

I have filed a bug about the issue with actions https://github.com/sonnyp/Workbench/issues/110

PR #106, opened 2022-05-26
Prevent dbusproxy from taking down Workbench

### LINKED ISSUE #95: Vala signal handlers crash Workbench
Follow up from https://github.com/sonnyp/Workbench/pull/94

Unfortunately out-of-process signal handlers completely crashes Workbench at the moment.

I couldn't find a way to define signal handlers for non-template builder in Vala.
If it's possible then I think we can rely on the user doing that and let the previewer process crash. If it's not possible we should fallback similarly to what I did for JS in https://github.com/sonnyp/Workbench/pull/94

Regardless - missing signal handlers shouldn't crash Workbench itself - rest can be implemented later as part of https://github.com/sonnyp/Workbench/issues/93

**Code** (Vala)

```
[GtkCallback]
public void on_button_clicked () { 
}
```

**UI** (Blueprint)

```
using Gtk 4.0;

Box welcome {
  orientation: vertical;
  valign: center;
  halign: center;
  
  Button {
    label: "Hello";
    clicked => on_button_clicked();
  }
}
```

sonnyp:
Closes #95 

image

### DIFF
--- a/src/Previewer/vala-previewer/previewer.vala
+++ b/src/Previewer/vala-previewer/previewer.vala
@@ -4,13 +4,12 @@ namespace Workbench {
   public class Previewer : Object {
     construct {
       this.window = new Gtk.Window () {
+        // Ensure the header bar has the same height as Workbench
+        titlebar = new Gtk.HeaderBar(),
         title = "Preview",
         hide_on_close = true,
         default_width = 600,
         default_height = 800
-      };
-      // Ensure the header bar has the same height as Workbench
-      this.window.titlebar = new Gtk.HeaderBar() {
       };
       this.window.close_request.connect (() => {
         this.window_open (false);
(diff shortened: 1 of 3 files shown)


src/Previewer/vala-previewer/previewer.vala
```diff
@@ -4,6 +4,9 @@ namespace Workbench {
   public class Previewer : Object {
     construct {
       this.window = new Gtk.Window () {
+        // Ensure the header bar has the same height as Workbench
+        titlebar = new Gtk.HeaderBar(),
```
lw64, line 8: There is a space missing before the brackets.
Apart from that you were faster with fixing than I could comment ;)

lw64 (changes requested):

Issue #60, opened 2022-05-10
Preview root object property changes


sonnyp:
Similar to https://github.com/sonnyp/Workbench/issues/59

When previewing a GtkWindow for example – updates to the properties of the window will not be picked up by the live preview.

We could just set all properties on change but what if a property gets removed? Is there a way to reset a GObject to default properties?

* [x] JavaScript https://github.com/sonnyp/Workbench/pull/115
* [ ] Vala

PR #115, opened 2022-06-12
Support root property updates for JS

sonnyp:
Fixes https://github.com/sonnyp/Workbench/issues/60

Needs the Vala equivalent and ready to go (as far as I can tell)

### DIFF
--- a/src/Library/demos/Application window/main.blp
+++ b/src/Library/demos/Application window/main.blp
@@ -1,47 +1,39 @@
 using Gtk 4.0;
 using Adw 1;
 
-Adw.ApplicationWindow window {
+Gtk.ApplicationWindow window {
   default-width: "800";
   default-height: "600";
-  content: Box {
-    orientation: vertical;
-
-    Adw.HeaderBar {
-      [title]
-      Label {
-        label: "My Awesome App";
-      }
-
-      [end]
-      MenuButton button_menu {
-        menu-model: menu_app;
-        icon-name: "open-menu-symbolic";
-        primary: true;
-      }
-    }
-
-    Adw.StatusPage {
-      title: "My App";
-      description: "My App is awesome";
-      vexpand: true;
-      icon-name: "re.sonny.Workbench-symbolic";
-
-      Box {
-        orientation: vertical;
-
-        LinkButton {
-          label: "API Reference";
-          uri: "https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.1/class.ApplicationWindow.html";
-        }
-
-        LinkButton {
-          label: "Human Interface Guidelines";
-          uri: "https://developer.gnome.org/hig/patterns/containers/windows.html";
-        }
-      }
+  title: "My App";
+  titlebar: Gtk.HeaderBar {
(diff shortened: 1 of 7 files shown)


src/Previewer/Internal.js
```diff
@@ -64,14 +65,48 @@ export default function Internal({
...
+          object_preview[prop_name] = null;
+        }
+
+        const old_value = object_root[prop_name];
+        if (!old_value !== new_value) {
+          object_root[prop_name] = new_value;
+        }
+      }
     }
-    adoptChild(object_preview, object_root);
```
sonnyp, line 74: since `child` and `content` are properties, and we now update all properties, we don't need this anymore

src/Library/demos/Application window/main.blp
```diff
@@ -1,47 +1,39 @@
...
-    Adw.StatusPage {
-      title: "My App";
-      description: "My App is awesome";
-      vexpand: true;
-      icon-name: "re.sonny.Workbench-symbolic";
+  Adw.StatusPage {
+    title: "My App";
+    description: "My App is awesome";
+    vexpand: true;
+    icon-name: "applications-science-symbolic";
```
sonnyp, line 21: I prefer using a "standard" icon

src/Library/demos/Application window/main.blp
```diff
@@ -1,47 +1,39 @@
 using Gtk 4.0;
 using Adw 1;
 
-Adw.ApplicationWindow window {
+Gtk.ApplicationWindow window {
```
sonnyp, line 4: no reason to use Adw.Window / Adw.ApplicationWindow if we don't want a freeform window

src/Library/demos/Application window/main.blp
```diff
@@ -1,47 +1,39 @@
...
-        label: "My Awesome App";
-      }
-
-      [end]
-      MenuButton button_menu {
-        menu-model: menu_app;
-        icon-name: "open-menu-symbolic";
-        primary: true;
-      }
+  title: "My App";
```
sonnyp, line 7: updating a root property, such as `title` now works as expected

src/Library/demos/Application window/main.blp
```diff
@@ -1,47 +1,39 @@
...
+  titlebar: Gtk.HeaderBar {
```
sonnyp, line 8: updating a root property widget such as `title` now also works as expected - try removing it

sonnyp:
@lw64 could you test/review this and once satisfied—implement the same algorithm on the Vala side  in this branch ?

lw64:
Instead of assigning the properties once when rebuilding, and figuring all the property flags out manually, I think it would be way easier to just bind the properties. Only the actually not visible GtkRoot needs to be held alive until the next preview.
```js
object_root.bind(
  prop_name,
  object_preview,
  prop_name,
  GObject.BindingFlags.BIDIRECTIONAL | GObject.BindingFlags.SYNC_CREATE
);
```

lw64 (approved):
Seems all good to me.

PR #119, opened 2022-06-26
Support update of root object type

### LINKED ISSUE #59: Preview change of object root type
When the root object (window) is changed from an AdwWindow to an GtkWindow for example, it doesn't get changed in the preview at the moment.

A possible solution is to always use AdwWindow and imitate a GtkWindow when it is used with putting a headerbar into AdwWindow.

sonnyp:
Fixes https://github.com/sonnyp/Workbench/issues/59

When the new UI code would change the type of the root object - simply destroy the previous one and set the new root object.

From a UX perspective

If the previous object was a GtkRoot - the window is closed
If the new object isn't a GtkRoot - the new object is directly previewed in Workbench window 
If the new object is a GTKRoot - user simply has to click on the primary button "Show Preview Window"

### DIFF
--- a/src/Previewer/Internal.js
+++ b/src/Previewer/Internal.js
@@ -62,18 +62,30 @@ export default function Internal({
     } else {
       obj = updateBuilderNonRoot(object_preview);
     }
-    builder.expose_object(original_id, obj);
+
+    if (original_id) {
+      builder.expose_object(original_id, obj);
+    }
+  }
+
+  function setObjectRoot(object) {
+    object_root = object;
+    object_root.set_hide_on_close(true);
+    object_root.connect("close-request", () => {
+      onWindowChange(false);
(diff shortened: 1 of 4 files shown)


src/Previewer/Internal.js
```diff
@@ -65,15 +65,24 @@ export default function Internal({
...
-      object_root.set_hide_on_close(true);
-      object_root.connect("close-request", () => {
-        onWindowChange(false);
-      });
+      setObjectRoot(object_preview);
+    } else if (
+      object_root.constructor.$gtype !== object_preview.constructor.$gtype
+    ) {
+      object_root.destroy();
+      setObjectRoot(object_preview);
```
sonnyp, line 85: @lw64 that's all there is to do

PR #143, opened 2022-08-17
Replace XML formatter and add format shortcut

sonnyp:
The shortcut is `<Ctrl><Shift>Return`.

There were a few issues with our previous XML formatter, the biggest one is that it would behave differently than Blueprint formatter. So I wrote a custom XML formatter that behaves exactly like Blueprint xml_emitter.py .

https://gitlab.gnome.org/jwestman/blueprint-compiler/-/blob/59283a76adc8d270ff5f67b630b7dfa905ec34a9/blueprintcompiler/xml_emitter.py

@melix99 could you test this quickly? Building this branch in GNOME Builder should work just fine.

Feel free to review as well but that's not required .


src/lib/ltx.js
```diff
@@ -939,10 +939,10 @@ class SaxLtx extends EventEmitter {
       if (!endTag) {
         this.emit("startElement", tagName, attrs);
         if (selfClosing) {
-          this.emit("endElement", tagName);
+          this.emit("endElement", tagName, true);
```
sonnyp, line 942: todo: upstream this

src/window.js
```diff
@@ -202,6 +202,29 @@ export default function Window({ application, version }) {
...
+      return prettier.format(text, {
+        parser: "css",
+        plugins: [prettier_postcss],
+      });
+    });
+
+    format(langs.xml.document.buffer, (text) => {
+      return xml.format(text, 2);
+    });
+  }
```
sonnyp, line 226: todo: only format source views which are visible

melix99:
There's something wrong while formatting XMLs like this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<interface>
  <object class="GtkBox"></object>
</interface>
```

It deletes code every time I press the shortcut.

sonnyp:
@melix99 nice catch — it's fixed

melix99:
Another thing I discovered is that the formatter is a bit too destructive when when the xml is not valid:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<interface>
  <object class="GtkBox" hey/>
</interface>
```

Is formatted to:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<interface>
```

I think it's better to just do nothing or showing an error when the xml is invalid instead of deleting code.

src/lib/ltx.js
sonnyp: https://github.com/xmppjs/ltx/pull/151

sonnyp:
@melix99 fixed as well

Thanks a lot for your help testing this. I will merge this but please do let me know if you encounter an issue or annoyance.

FYI this is a known issue https://github.com/sonnyp/Workbench/issues/144

Issue #148, opened 2022-09-05
Frequent hangs when editing a somewhat complex .blp


YaLTeR:
I'm experimenting with a UI I want to make and my .blp got somewhat lengthy. For some reason, quite frequently now, editing something will cause Workbench to hang forever. I saw it happen even on simple things like editing a string property. Here's the .blp:
```
using Gtk 4.0;
using GtkSource 5;
using Adw 1;

Adw.ApplicationWindow {
  Box {
    orientation: vertical;
    
    Adw.HeaderBar {
      centering-policy: strict;
      
      [title]
      Adw.ViewSwitcherTitle view_switcher_title {
        stack: view_stack;
      }
      
      [end]
      ToggleButton {
        icon-name: "branch-symbolic";
        active: bind branch_flap.reveal-flap bidirectional;
      }
    }

    Adw.ToastOverlay overlay {
      Adw.Flap branch_flap {
        flap-position: end;
        
        [content]
        Box {
          orientation: vertical;

          Adw.ViewStack view_stack {
            vexpand: true;
            
            Adw.ViewStackPage {
              title: "Frame Bulk";
              icon-name: "columns-symbolic";
              child: ScrolledWindow {
                hscrollbar-policy: never;

                Box {
                  orientation: vertical;
                  spacing: 24;
                  
                  Label {
                    styles ["monospace", "card"]
                    name: "frame-bulk-string";
                    halign: center;
                    selectable: true;
                    label: "s03-j--c--|------|--u---|0.004|97.49886|-|2";
                  }
                  
                  Box {
                    spacing: 24;
                    halign: center;
                    
                    Box {
                      orientation: vertical;
                      spacing: 24;
                      
                      ListBox {
                        styles ["boxed-list"]
                      
                        Adw.ActionRow {
                          title: "Strafe";
                          activatable-widget: strafe_switch;
                          
                          Switch strafe_switch {
                            valign: center;
                          }
                        }
                        
                        Adw.ComboRow {
                          title: "Type";
                          model: StringList {
                            strings [
                              "Speed Increasing",
                              "Quick Turn",
                              "Slow Down",
                              "Constant Speed",
                            ]
                          };
                        }
                        
                        Adw.ComboRow {
                          title: "Direction";
                          model: StringList {
                            strings [
                              "Turn Left",
                              "Turn Right",
                              "Best for Type",
                              "Towards Yaw",
                              "Towards Point",
                              "Wiggle Left",
                              "Wiggle Right",
[...]

PR #151, opened 2022-09-11
Throttle Blueprint updates

sonnyp:
Closes https://github.com/sonnyp/Workbench/issues/148

The current implementation (in `main`) is fairly naive; on each text update, Workbench sends the whole text to the Blueprint LSP, waits for an event and then update the Builder.

The whole thing is asynchronous, and it takes longer to complete than the time it takes to get the next "update" signal.
This PR fixes that by waiting for the latest compile before starting the next one.

Here are a couple of ideas on how to make Blueprint updates even faster but we I'll need to measure impact first

1. Extract the event to be its own routine (outside of  `compileBlueprint`) 
2. Implement LSP text range updates
3. Embed/use pyjobject 
4. Use https://github.com/Prince781/lsp-glib

### DIFF
--- a/re.sonny.Workbench.json
+++ b/re.sonny.Workbench.json
@@ -1,7 +1,7 @@
 {
   "id": "re.sonny.Workbench",
   "runtime": "org.gnome.Sdk",
-  "runtime-version": "43beta",
+  "runtime-version": "42",
   "sdk": "org.gnome.Sdk",
   "command": "workbench",
   "finish-args": [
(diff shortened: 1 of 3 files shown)


YaLTeR:
Haven't been able to hang it over my (very brief) testing. Also it seems to perform much better because before (and right now with the Blueprint VSCode extension) you would see the Blueprint LSP slowly struggling to highlight errors character by character when you type, but with this PR it's near-instant for the whole change, because it skips those intermediate steps.

PR #152, opened 2022-09-14
Inline CSS errors

sonnyp:
Todo:

* [x] Remove duplicate entries
* [x] Add inline CSS errors for out-of-process preview

### DIFF
--- a/src/Previewer/Internal.js
+++ b/src/Previewer/Internal.js
@@ -17,6 +17,7 @@ export default function Internal({
   window,
   application,
   dropdown_preview_align,
+  panel_style,
 }) {
   const stack = builder.get_object("stack_preview");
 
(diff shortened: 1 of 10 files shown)


src/Previewer/Internal.js
```diff
@@ -189,6 +190,10 @@ export default function Internal({
     }
 
     css_provider = new Gtk.CssProvider();
+    panel_style.reset();
+    css_provider.connect("parsing-error", (self, section, error) => {
+      panel_style.handleCssParserError(section, error);
+    });
```
sonnyp, line 196: @lw64 do you have thoughts on what's the best way to implement this in `updateCss` in `previewer.vala` ?

My idea would be to serialize section/error and send it on the bus but I'm not sure if it's possible for a client and how.

lw64: It could be implemented using a dbus signal.

sonnyp: :+1: 

 can you review what I did? Specially the vala part

PR #156, opened 2022-09-25
wip: Add support for vala-language-server

lw64:


src/Library/demos/Toasts/main.vala
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/vala workbench.vala --pkg gtk4 --pkg libadwaita-1
```
sonnyp, line 1: Let's use `env` unless there is a good reason not too.

```
#!/usr/bin/env -S vala workbench.vala --pkg gtk4 --pkg libadwaita-1
```

src/Compiler.js
```diff
@@ -37,11 +44,11 @@ export default function Compiler(data_dir) {
       ).catch(() => {}),
     ]);
 
+    GLib.chdir(data_dir);
```
sonnyp, line 47: this changes the current directory for the whole process - please use replace `GSbprocess` with  `GSubprocessLauncher` instead which supports setting cwd

https://docs.gtk.org/gio/method.SubprocessLauncher.set_cwd.html

src/Compiler.js
```diff
@@ -17,6 +17,13 @@ export default function Compiler(data_dir) {
   );
 
   async function compile(code) {
+    let args;
+    try {
+      args = code.split("\n")[0].split("/vala ")[1].split(" ");
```
sonnyp, line 22: can you move this to its own function (in the same file) and document what it does

sonnyp (commented):
GitHub is showing an outdated diff `LSPClient.js` - can you merge `main` or rebase onto it?

src/Compiler.js
```diff
@@ -12,11 +12,15 @@ export default function Compiler(data_dir) {
   const module_file = Gio.File.new_for_path(
     GLib.build_filenamev([data_dir, "libworkbenchcode.so"])
   );
-  const api_file = Gio.File.new_for_path(
-    GLib.build_filenamev([pkg.pkgdatadir, "workbench-api.vala"])
-  );
```
sonnyp, line 17: @lw64 this wasn't used

it's no needed?

sonnyp: its used in `PanelCode.js`

sonnyp (approved):
Well done

PR #160, opened 2022-10-01
Use gjspack

sonnyp:


src/meson.build
```diff
@@ -58,5 +54,24 @@ configure_file(
   install_dir: get_option('bindir')
 )
 
-install_data('workbench-api.vala', install_dir: datadir)
+install_data('workbench-api.vala', install_dir: pkgdatadir)
```
sonnyp, line 57: @lw64 FYI

lw64: the command to run `valac` needs to be adjusted, if they are different paths.

lw64: here: https://github.com/sonnyp/Workbench/blob/main/src/Compiler.js#L16

src/Compiler.js
```diff
@@ -13,7 +13,7 @@ export default function Compiler(data_dir) {
     GLib.build_filenamev([data_dir, "libworkbenchcode.so"])
   );
   const api_file = Gio.File.new_for_path(
-    GLib.build_filenamev(["/app/share", "workbench-api.vala"])
+    GLib.build_filenamev([pkg.pkgdatadir, "workbench-api.vala"])
```
lw64, line 16: perfect! didnt saw this :)

PR #169, opened 2022-11-05
Fix Vala external preview

sonnyp:
If there was no original id - Vala preview would fail with

> No UI definition loaded yet.

because of

> Error: Expected type string for argument 'string' but got type
undefined

### DIFF
--- a/src/Previewer/Previewer.js
+++ b/src/Previewer/Previewer.js
@@ -378,7 +378,7 @@ function targetBuildable(tree) {
 
   const child = findPreviewable(tree);
   if (!child) {
-    return [null, ""];
+    return {};
   }
 
   const original_id = child.attrs.id;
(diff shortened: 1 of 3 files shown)


src/Previewer/Previewer.js
```diff
@@ -378,7 +378,7 @@ function targetBuildable(tree) {
 
   const child = findPreviewable(tree);
   if (!child) {
-    return [null, ""];
+    return {};
```
sonnyp, line 381: Unrelated but the return type was wrong

src/Previewer/vala-previewer/previewer.vala
```diff
@@ -20,14 +20,16 @@ namespace Workbench {
       });
     }
 
-    public void update_ui (string content, string target_id, string original_id) {
+    public void update_ui (string content, string target_id, string original_id = "") {
```
lw64, line 23: Checking for undefined in the js side and vala side is twice, but I guess it doesn't hurt.

lw64 (approved):
Nothing to complain :)

Issue #133, opened 2022-07-10
Destroy window preview when closed


sonnyp:
Going from `modal: true` to `modal: false` does not work.

PR #171, opened 2022-11-07
Destroy preview window on close

sonnyp:
Fixes https://github.com/sonnyp/Workbench/issues/133

### DIFF
--- a/src/Previewer/Internal.js
+++ b/src/Previewer/Internal.js
@@ -19,6 +19,7 @@ export default function Internal({
   application,
   dropdown_preview_align,
   panel_style,
+  panel_ui,
 }) {
   const stack = builder.get_object("stack_preview");
 
@@ -29,8 +30,20 @@ export default function Internal({
     builder.get_object("button_screenshot").visible = true;
   }
 
-  function open() {
-    if (!object_root) return;
+  async function open() {
+    // The flow for internal preview is complicated after the window has been destroyed (on close)
+    // Internal cannot rebuild it on its own so it calls panel_ui.update which will endup trigering internal.updateXML.
+    // instead - Internal could work like the Vala previwer and take a XML string that it can parse whenever it needs
+    if (!object_root) {
+      try {
+        await panel_ui.update();
+      } catch (err) {
+        logError(err);
+        return;
+      }
+      if (!object_root) return;
+    }
+
     object_root.present_with_time(Gdk.CURRENT_TIME);
     onWindowChange(true);
   }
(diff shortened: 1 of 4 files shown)


src/Previewer/Internal.js
```diff
@@ -29,8 +30,16 @@ export default function Internal({
     builder.get_object("button_screenshot").visible = true;
   }
 
-  function open() {
-    if (!object_root) return;
+  async function open() {
+    if (!object_root) {
+      try {
+        await panel_ui.update();
```
sonnyp, line 36: The flow is complicated for internal preview.

After the window is destroyed - `Internal` cannot rebuild it on its own and triggers an update on `panel_ui` which will call `internal.updateXML`. 

I wonder if instead – `Internal` shouldn't work like `External` and get a serializable string that it can feed into its own `Builder` whenever it needs.

src/Previewer/Internal.js
```diff
@@ -89,8 +98,9 @@ export default function Internal({
 
   function setObjectRoot(object) {
     object_root = object;
-    object_root.set_hide_on_close(true);
     object_root.connect("close-request", () => {
+      log("close");
```
sonnyp, line 102:
```suggestion
```

src/Previewer/vala-previewer/previewer.vala
```diff
@@ -3,56 +3,73 @@ namespace Workbench {
...
+
+    private void ensure_window() {
+      if (this.window != null) {
+        return;
+      }
+      var window = new Gtk.Window () {
         // Ensure the header bar has the same height as Workbench
         titlebar = new Gtk.HeaderBar (),
         title = "Preview",
-        hide_on_close = true,
```
sonnyp, line 10: @lw64 can you review/test the Vala / external preview side of this PR?

src/Previewer/Internal.js
lw64: That seems reasonable. Making both previewers work in the same way would make maintainance easier, and I don't think it make it loose features. Only the GtkBuilder from the previewer would need to be handed over to the js code panel.

src/Previewer/vala-previewer/previewer.vala
lw64: It looks all right to me. Only a bit of moving code around into some methods. Except the `ensure_window()`, which is now called on demand instead of on creation of the process.

I see you have your own code style, so going with a prettyfier over it to establish it in the whole file would be good, right? (I used another style in the beginning and it is now a bit inconsistent.)

lw64 (approved):

sonnyp:
linter/prettyfier is always good :+1:

PR #176, opened 2022-11-26
Destroy toplevel objects

sonnyp:


src/Previewer/vala-previewer/previewer.vala
```diff
@@ -75,6 +76,12 @@ namespace Workbench {
         ((Gtk.Window) target).child = null;
         this.window.child = child;
       }
+
+      // FIXME: working for Internal but not here
+      // the list of root objects keeps growing in Inspector as you make changes to UI
+      // Toplevel windows returned by these functions will stay around until the user explicitly destroys them with gtk_window_destroy().
+      // https://docs.gtk.org/gtk4/class.Builder.html
+      target.destroy();
```
sonnyp, line 84: @lw64 could you look into why this doesn't work on the external previewer and global objects keep accumulating when making changes ?

# How to reproduce:

1. Select the `Application window` example
2. Selecting Vala
3.  Entering the following code:

```vala
#!/usr/bin/env -S vala workbench.vala --pkg gtk4 --pkg libadwaita-1
public void main() {}
```

4. Hit `Run`
5. Open the GTK Inspector on the preview window/process
6. Make changes to `UI`
7. Observe the list of global objects grow for each change


Screenshot from 2022-11-25 18-38-14

src/Previewer/Internal.js
```diff
@@ -158,6 +158,12 @@ export default function Internal({
           object_root[prop_name] = new_value;
         }
       }
+
+      // Toplevel windows returned by these functions will stay around until the user explicitly destroys them with gtk_window_destroy().
+      // https://docs.gtk.org/gtk4/class.Builder.html
+      if (object_preview instanceof Gtk.Root) {
+        object_preview.destroy();
+      }
```
sonnyp, line 166: ths works as expected and prevents the list of global objects to grow with every UI change

Issue #69, opened 2022-05-15
Demo/example ideas


sonnyp:
High prio

* [ ] How to use CSS with GTK
* [ ] HTTP requests

Medium prio

* [ ] Mnemonics explained

Low prio (already plenty of resources)

* [ ] properties
* [ ] signals

lw64:
Some others:
- drag and drop
- maps (libshumate)
- webview
- 3d (GLArea)
- 2d, maybe something like turtle (cairo)
- video/audio playback (gstreamer)
- tracker
- clipboard
- adwaita listboxes
- list with search/filer/sorting

sonnyp:
There are some demos here for inspiration https://gitlab.gnome.org/Archive/gnome-devel-docs/-/tree/master/platform-demos/C

sonnyp:
## Ideas

Higher prio

* [ ] How to use CSS with GTK https://github.com/sonnyp/Workbench/issues/58
* [ ] HTTP requests with libsoup
* [ ] Opening and saving files (should probably use libportal)
* [ ] List of common GTK widgets to use
* [ ] List of common Adwaita widgets to use
* [ ] How to use Icons
* [ ] how to use gtk inspector https://gitlab.gnome.org/Teams/documentation/developer-www/-/merge_requests/87
* [ ] fullscreen

Medium prio

* [ ] Mnemonics explained

Low prio (already plenty of resources)

* [ ] properties
* [ ] signals

kakonodia:
Greetings @sonnyp 
Hope you are doing well 
i am K Nodia an outreachy Applicant  while reading through projects i got interest in this 
i followed the contributing Guide and i am done setting up the projects on my PC its now Running 
I would like any advice on what s Next 
Thank you

sonnyp:
Hello, please join the chatroom https://matrix.to/#/%23workbench:gnome.org

PR #193, opened 2023-02-23
Created a New Section How To and added new Demo

SoNiC-HeRE:
Created a section "How To" under which I added How to Use Css with GTK Demo

Issue #69

### DIFF
--- a/src/Library/Library.blp
+++ b/src/Library/Library.blp
@@ -31,6 +31,10 @@ Adw.PreferencesWindow library {
       title: "User Interface";
     }
 
+     Adw.PreferencesGroup library_how_to {
+      title: "How To";
+    }
+
     Adw.PreferencesGroup {
       vexpand: true;
       valign: end;
--- a/src/Library/demos/Use Css with GTK/main.blp
+++ b/src/Library/demos/Use Css with GTK/main.blp
@@ -1,0 +1,29 @@
+using Gtk 4.0;
+using Adw 1;
+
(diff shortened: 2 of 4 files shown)


sonnyp (changes requested):
Well done on figuring out how to add a Library section, unfortunately this entry doesn't fit into Workbench.

From the contributing guide:

https://github.com/sonnyp/Workbench/blob/main/CONTRIBUTING.md#library

> Keep them concise and interactive.

This is too long, hard to read and not interactive.

Workbench goal is to teach by example and teach by doing. Through code, not text instructions. Otherwise, it would be a website, and we already have developer.gnome.org for that.

A better library entry for this would be to show how to use CSS from a code PoV. Selectors, classes etc. Using CSS/Blueprint code in the `Style`, `UI` and `Preview` panels with code comments.

Since you're not familiar with GTK CSS yet and as suggested in the contribution guide – I would recommend starting with something small and accessible

> You can find ideas in [this issue](https://github.com/sonnyp/Workbench/issues/69). Start with something small and accessible.

Here are 2 simple ideas:

# GtkSwitch

Should be pretty straightforward—there is not much to it.

* the `active` property
* the `notify::activate` property

# GtkButton

A bit more work but still accessible.

* the `sensitive` property
* the `clicked` signal
* the `flat` class
* the various classes for it

Screenshot from libadwaita demo:

image

SoNiC-HeRE:
Sure, will work on it

PR #197, opened 2023-02-24
library: Add Avatar entry

AkshayWarrier:
I have kept this demo fairly simple now. I wasn't sure if I should show ``set_custom_image`` or ``set_icon_name``.

Relates to issue #69


sonnyp (changes requested):
Good start, thanks!

There are 3 “main” modes for avatar, see below for what I mean. Feel free to use as is.

The "Image…" button should let the user select a file. I'll let you figure this out but feel free to ask in the room if you struggle.

We should go with the simplest file chooser API. I'd suggest using https://docs.gtk.org/gtk4/class.FileDialog.html 

It's brand new and will only be available in the next GNOME platform release but that doesn't mean we can't use it already.

You will need to set `"runtime-version": "master",` in `re.sonny.Workbench.Devel.json` to make use of it. 

```blp
using Gtk 4.0;
using Adw 1;

Adw.StatusPage {
  title: "Avatar";
  description: "Display a round avatar for image or initials.";

  Box {
    orientation: vertical;
    halign: center;
    
    Adw.Avatar {
      size: 96;
      show-initials: true;
      text: "John Doe";
      margin-bottom: 6;
    }
    
    Label {
      label: "Initials";
      margin-bottom: 30;
    }
    
    Adw.Avatar {
      size: 96;
      margin-bottom:6;
    }
    
    Button {
      label: "Image…";
      margin-bottom: 30;
      styles ["pill"]
    }
    
    Adw.Avatar {
      size:96;
      margin-bottom:6;
    }
    
    Label {
      label: "Fallback";
      margin-bottom: 30;
    }


    LinkButton {
      label: "API Reference";
      uri: "https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.2/class.Avatar.html";
    }
  }
}
```

AkshayWarrier:
Thank you so much for the detailed review! I'll get back to you with the changes as quickly as possible.

src/Library/demos/Avatar/main.js
```diff
@@ -0,0 +1,28 @@
...
+import Gtk from "gi://Gtk";
+import Gdk from "gi://Gdk";
+
+const avatar_image = workbench.builder.get_object("avatar_image");
+const button = workbench.builder.get_object("button");
+
+const dialog = new Gtk.FileDialog();
+
+const filter = new Gtk.FileFilter();
+filter.name = "Image files";
```
sonnyp, line 11: the filter name doesn't seem to work – this is what I see in the bottom right of the dialog

image

could you look into it?

try 

```
new Gtk.FileFilter({
  name: "something"
})
```

also "Images" is fine - no need to be overly specific

src/Library/demos/Avatar/main.js
```diff
@@ -0,0 +1,28 @@
...
+filter.add_pixbuf_formats();
```
sonnyp, line 12: :+1:

src/Library/demos/Avatar/main.js
```diff
@@ -0,0 +1,28 @@
...
+dialog.default_filter = filter;
```
sonnyp, line 13: if we create the filter first we can do

```
new Gtk.FileDialog({
  default_filter: filter
})
```

src/Library/demos/Avatar/main.js
```diff
@@ -0,0 +1,28 @@
...
+    const file = dialog.open_finish(res);
+    const texture = Gdk.Texture.new_from_file(file);
+    avatar_image.set_custom_image(texture);
+  } catch {
+    return;
+  }
+}
+
+button.connect("clicked", () => {
+  dialog.open(workbench.window, null, on_open_cb);
```
sonnyp, line 26: please use `Gio._promisify` like the other entries with async ops

and no need for the extra function

src/Library/demos/Avatar/main.js
```diff
@@ -0,0 +1,28 @@
+import Adw from "gi://Adw";
+import Gtk from "gi://Gtk";
+import Gdk from "gi://Gdk";
+
+const avatar_image = workbench.builder.get_object("avatar_image");
+const button = workbench.builder.get_object("button");
+
+const dialog = new Gtk.FileDialog();
```
sonnyp, line 8: please make this modal and add a title  "Select an Avatar"

sonnyp (changes requested):
:+1:

src/Library/demos/Avatar/main.js
AkshayWarrier:
```
new Gtk.FileFilter({
  name: "something"
})
```
This does not seem to solve the issue. I'll still just leave it in, if it gets fixed in the future.

sonnyp (approved):
Well done.

I have made minor changes before merging https://github.com/sonnyp/Workbench/pull/197/commits/5c147002b46e8a85f6d24763743f4d11ccd4ca07

PR #198, opened 2023-02-25
library: Add Button entry

SoNiC-HeRE:
Created Button Demo under User Interface Section in Library
Issue #69

### DIFF
--- a/src/Library/demos/Button/main.blp
+++ b/src/Library/demos/Button/main.blp
@@ -8,116 +8,107 @@ Adw.StatusPage {
   Box {
     orientation: vertical;
     halign: center;
+    margin-top: 20;
+    margin-bottom: 20;
 
-       Grid buttons {
-         column-spacing: 40;
-         row-spacing: 40;
-         margin-bottom: 50;
+    FlowBox {
+      min-children-per-line:3;
+      margin-bottom: 20;
+
 
          Button Regular {
               name: "regular";
               label: "Regular";
-              layout {
-                column: 1;
-                row: 1;
-              }
+              margin-end: 40;
             }
 
          Button Flat {
               name: "flat";
               label: "Flat";
-              layout {
-                column: 3;
-                row: 3;
-              }
+              margin-end: 40;
               styles ["flat"]
             }
 
          Button Suggested {
               name: "suggested";
               label: "Suggested";
-              layout {
-                column: 3;
-                row: 1;
-              }
               styles ["suggested-action"]
             }
+        }
+
+   FlowBox {
+     min-children-per-line:3;
+     margin-bottom: 20;
 
          Button Destructive {
               name: "destructive";
               label: "Destructive";
-              layout {
-                column: 1;
-                row: 3;
-              }
+              margin-end: 40;
               styles ["destructive-action"]
             }
 
          Button Custom {
               name: "custom";
               label: "Custom";
-              layout {
-                column: 2;
-                row: 2;
-              }
-              styles ["pill","opaque"]
-            }
-
-         Button Plus {
-              name: "plus";
-              icon-name: "plus";
-              halign: center;
-              layout {
-                column: 2;
-                row: 1;
-              }
-              styles ["circular"]
-            }
-
-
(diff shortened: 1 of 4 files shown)


src/Library/demos/Button/main.blp
```diff
@@ -0,0 +1,45 @@
+using Gtk 4.0;
+using Adw 1;
+
+ Adw.StatusPage{
+  title: "Button";
+  description: "A simple way to trigger an event";
```
bertob, line 6: "Allow people to perform actions by clicking"

src/Library/demos/Button/main.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "Button",
+  "category": "user_interface",
+  "description": "A simple way to trigger an event",
```
bertob, line 4: Same as above

bertob (commented):
So in general this should split out the different types of buttons a bit more, it's mixing things too much. I'd not make them all pill buttons, and have separate buttons for showing the main types of buttons. I'd try a grid with normal, suggested, destructive, etc. similar to what's in the style classes demo in libadwaita.

I'd also not show .flat buttons here, because they're a special thing mainly to be used in constrained environments like toolbars/headerbars.

image

src/app.gresource.xml
```diff
@@ -37,5 +37,8 @@
     <file>Library/demos/Custom Widget/main.ui</file>
     <file>Library/demos/Custom Widget/main.js</file>
     <file>Library/demos/Custom Widget/main.json</file>
+    <file>Library/demos/Button/main.blp</file>
+    <file>Library/demos/Button/main.json</file>
+    <file>Library/demos/Button/main.js</file>
```
sonnyp, line 42: this is not needed anymore (you will have to merge main into your branch)

https://github.com/sonnyp/Workbench/pull/200

src/Library/demos/Button/main.blp
```diff
@@ -0,0 +1,45 @@
...
+        label: "API Reference";
+        uri: "https://docs.gtk.org/gtk4/class.Button.html";
+    }
+
+      LinkButton{
+        label: "Human Interface Guidelines";
+        uri: "https://developer.gnome.org/hig/patterns/controls/buttons.html";
+    }
+
+     }
```
sonnyp, line 43: please take care of the indentation until it's automated

it should be 2 spaces

src/Library/demos/Button/main.js
```diff
@@ -0,0 +1,19 @@
...
+    buttons: Gtk.ButtonsType.OK,
+  });
+  dialog.connect("response", () => {
+    dialog.close();
+  });
+
+  dialog.present();
+}
+
+test_button.connect("clicked", handle_click);
```
sonnyp, line 19: the dialog is a bit annoying while testing because the user needs to close the modal to proceed - it is 'blocking'

Also `Gtk.MessageDialog` will be deprecated in GTK 4.10

I don't have a much better idea - I'd say `console.log("Hello, World!");` is enough for now

sonnyp (changes requested):
Agree with Tobias 

There should be

* Regular
* Suggested
* Destrucive
* Custom (1)
* Pill
* Circular
* osd
* Disabled

The challenge is to make the grid look good.

You can take inspiration from libadwaita demo:

image

if you hover one of them - it tells you what CSS class is used

SoNiC-HeRE:
Okay understood. I'll make the necessary changes.

SoNiC-HeRE:
Further to enhance the design i intend to smaller the icon-buttons respective to the class buttons but somehow due to the grid the buttons scale equally.

SoNiC-HeRE:
Applied Changes to match styling as of Adwaita Demo

sonnyp (approved):
Thanks!

I did some cleanups before merging https://github.com/sonnyp/Workbench/pull/198/commits/8f5fc6ebbd5f65349afba79b78c8d63e69dda703

`FlowBox` didn't do anything useful and I simplified a bit the JS code

PR #206, opened 2023-03-05
library: Add Text Colors demo

bragefuglseth:
This demo showcases how you can use Pango attributes to easily apply complex formatting to labels.


sonnyp (approved):
Thanks a lot! Don't hesitate if you make more.

I've made it interactive by adding a GtkEntry.

I think we'll move this kind of entries to a dedicated "Demos" section but for now it's fine

PR #211, opened 2023-03-08
Library: Add Adw.MessageDialog example

halfmexican:
My First PR for this project! I included a .js and .vala file.
If there is any issues or recommendations please let me know!


sonnyp (changes requested):
Great start!

I was planning on waiting for Blueprint support but I guess it makes sense to create them from code anyway.

Following the GNOME HIG, I think we should have at least these 3 examples:

1. Confirmation dialog
2. Error dialog
3. Action dialog

And your password example is good too - maybe under "Advanced error dialog"

src/Library/demos/Dialogs/main.js
```diff
@@ -0,0 +1,69 @@
...
+const button_advanced = workbench.builder.get_object("button_advanced");
+const window = button_simple.get_ancestor(Gtk.Window);
+
+function createSimpleDialog() {
+  let dialog = new Adw.MessageDialog({
+    heading: "Save Changes?",
+    body: "Opened files have unsaved changes. Unsaved changes will be lost forever!",
+    close_response: "cancel",
+    modal: true,
+  });
```
sonnyp, line 14: this is not really a pattern we would use

can you reuse the heading/body of the HIG example https://developer.gnome.org/hig/patterns/feedback/dialogs.html?highlight=dialog#message-dialogs ?

halfmexican: Yes I can. I want to follow HIG and will be adding the error dialog as well. I was wondering if you could clear some things up for me. 

Why wouldn't we use this pattern? I mimicked the message dialog example from the Adwaita Demo. I liked the example as it shows how message dialogs have up to 3 responses and shows off each of the response appearance types.

This example on the hig patterns only shows 2 of the appearance types.
image

src/Library/demos/Dialogs/main.js
```diff
@@ -0,0 +1,69 @@
...
+
+  // Make the dialog transient over the main window
+  dialog.set_transient_for(window);
+
+  //Negative responses like Cancel or Close should use the default appearance.
```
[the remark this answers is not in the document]
halfmexican: I'll remove the comment. I ripped it from the adwaita documentation.
image

src/Library/demos/Dialogs/main.js
halfmexican: Nevermind, I realized I used suggested response for the login example

sonnyp (approved):
Thanks !

Made a few tiny changes https://github.com/sonnyp/Workbench/pull/211/commits/215d07e5131eb0ee2ac6aa35a817955be7c6aea1

(Avoid "!" in user text, they are too aggressive)

sonnyp:
ha - an other thing - please create a new branch before submitting a PR - don't use `main`

otherwise I have to push to your `main` branch when making changes to your PRs before merging and it doesn't work if you have multiple PRs

PR #212, opened 2023-03-08
Fix typo "Information about Blueprint"

imhemish:


imhemish:
Casually saw this while using Workbench

sonnyp (approved):
Nice catch, thank you.

PR #216, opened 2023-03-10
library: Add Box entry

AkshayWarrier:
This PR attempts to add an interactive demo for ``Gtk.Box``

Issue #69

### DIFF
--- a/src/Library/demos/Box/main.blp
+++ b/src/Library/demos/Box/main.blp
@@ -1,0 +1,96 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.Clamp {
+  Box {
+    orientation: vertical;
+
+    Label {
+      label: "Box";
+      margin-top: 12;
+      margin-bottom: 12;
+
+      styles [
+        "title-1",
+      ]
+    }
+
+    Label {
+      label: "A widget that arranges its child widgets into a single row or column";
+      margin-bottom: 18;
+    }
+
+    Box {
+      orientation: vertical;
+      margin-bottom: 24;
+
+      Box {
+        hexpand: true;
+        orientation: vertical;
+        spacing: 12;
+
+        Box {
+          halign: center;
+          spacing: 18;
+          Button button_add{
+            label: "Add Item";
+            styles [
+              "suggested-action",
+              "pill",
+            ]
+          }
+          Button button_remove{
+            label: "Remove Item";
+            styles [
+              "destructive-action",
+              "pill",
+            ]
+          }
+        }
+
+        Box {
+          halign: center;
+          spacing: 18;
+
+          CheckButton highlight {
+            label: "Highlight Box";
+          }
+
+          CheckButton toggle_orient {
+            label: "Toggle Orientation";
+          }
+
+          CheckButton homogeneous {
+            label: "Homogeneous";
+          }
+
+          CheckButton prepend {
+            label: "Prepend Items";
+          }
+        }
+
+        Label {
+          label: "Horizontal Alignment";
+        }
+
+        Scale alignment {
+            round-digits:10;
+            margin-bottom: 18;
+            adjustment: Gtk.Adjustment {
+              lower: 0;
+              upper: 100;
+              value: 0;
+            };
+        }
+      }
+
+      ScrolledWindow {
+        Box interactive_box {
+          spacing: 10;
+          orientation: horizontal;
+          vexpand: true;
+        }
+      }
+    }
+  }
+}
--- a/src/Library/demos/Box/main.css
+++ b/src/Library/demos/Box/main.css
@@ -1,0 +1,7 @@
+#card {
+  padding: 24px;
+}
+
+#border {
+  border: 3px solid @accent_color;
+}
--- a/src/Library/demos/Box/main.js
+++ b/src/Library/demos/Box/main.js
@@ -1,0 +1,76 @@
+import Gtk from "gi://Gtk";
+
+const interactive_box = workbench.builder.get_object("interactive_box");
+const button_add = workbench.builder.get_object("button_add");
+const prepend = workbench.builder.get_object("prepend");
+const button_remove = workbench.builder.get_object("button_remove");
+var count = 0;
+
+button_add.connect("clicked", () => {
+  const label = new Gtk.Label({
+    name: "card",
+    label: `Item ${count + 1}`,
+  });
+  const style = label.get_style_context();
+  style.add_class("card");
+  prepend.active
+    ? interactive_box.prepend(label)
+    : interactive_box.append(label);
+  count += 1;
+});
+
+button_remove.connect("clicked", () => {
+  if (count) {
+    interactive_box.remove(interactive_box.get_last_child());
+    count -= 1;
+  } else {
+    console.log("The box has no child widgets to remove");
+  }
+});
+
+const highlight = workbench.builder.get_object("highlight");
+highlight.connect("toggled", () => {
+  highlight.active
+    ? (interactive_box.name = "border")
+    : (interactive_box.name = "");
+});
+
+const toggle_orient = workbench.builder.get_object("toggle_orient");
+toggle_orient.connect("toggled", () => {
+  interactive_box.orientation === Gtk.Orientation["HORIZONTAL"]
+    ? (interactive_box.orientation = Gtk.Orientation["VERTICAL"])
+    : (interactive_box.orientation = Gtk.Orientation["HORIZONTAL"]);
+});
+
+const homogeneous = workbench.builder.get_object("homogeneous");
+homogeneous.connect("toggled", () => {
+  interactive_box.homogeneous = homogeneous.active;
+});
+
+const alignment = workbench.builder.get_object("alignment");
+
+const marks = {
+  0: "Fill",
+  33: "Start",
+  66: "Center",
+  100: "End",
+};
+
+const label_to_alignment = {
+  Fill: Gtk.Align.FILL,
+  Start: Gtk.Align.START,
+  Center: Gtk.Align.CENTER,
+  End: Gtk.Align.END,
+};
+
+for (const [value, label] of Object.entries(marks)) {
+  alignment.add_mark(value, Gtk.PositionType.RIGHT, label);
+}
+alignment.set_increments(33, 100);
+
+alignment.connect("value-changed", () => {
+  const scale_value = alignment.get_value();
+  const label = marks[scale_value];
+  if (!label) return;
+  interactive_box.halign = label_to_alignment[label];
+});
--- a/src/Library/demos/Box/main.json
+++ b/src/Library/demos/Box/main.json
@@ -1,0 +1,10 @@
+{
+  "name": "Box",
+  "category": "user_interface",
+  "description": "A widget that arranges its child widgets into a single row or column",
+  "panels": [
+    "ui",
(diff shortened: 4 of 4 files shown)


src/Library/demos/Box/main.blp
```diff
@@ -0,0 +1,109 @@
...
+      margin-top: 12;
+      margin-bottom: 12;
+
+      styles [
+        "title-1",
+      ]
+    }
+
+    Label {
+      label: "A widget that arranges its child widgets into a single row or column";
```
andyholmes, line 19: The strings in this file should have gettext translation functions.

```suggestion
      label: _("A widget that arranges its child widgets into a single row or column");
```

src/Library/demos/Box/main.blp
```diff
@@ -0,0 +1,109 @@
...
+          CheckButton prepend {
+            label: "Prepend Items";
+          }
+        }
+
+        Label {
+          label: "Horizontal Alignment";
+        }
+
+        Scale alignment {
```
andyholmes, line 75: The blue fill level here doesn't make much sense, so best just to turn it off.

```suggestion
        Scale alignment {
          has-origin: false;
```

src/Library/demos/Box/main.blp
```diff
@@ -0,0 +1,109 @@
...
+            margin-bottom: 18;
+            adjustment: Gtk.Adjustment {
+              lower: 0;
+              upper: 100;
+              value: 0;
+            };
+        }
+      }
+
+      ScrolledWindow {
```
andyholmes, line 85:
```suggestion
      ScrolledWindow {
        has-frame: true;
```

Add a frame to this, so it's clear how the box is moving in relation to its parent.

There used to be this helpful image in our documentation, with each tile showing the widget's geometry and position in relation to the parent. So as you're flipping the switches in your demo, each combination would show one of the grid tiles below:

widget-hvalign

src/Library/demos/Box/main.css
```diff
@@ -0,0 +1,7 @@
+#card {
```
andyholmes, line 1: This should be a CSS class, since you are using it more than once.

src/Library/demos/Box/main.css
```diff
@@ -0,0 +1,7 @@
...
+  padding: 24px;
```
andyholmes, line 2: Since the `.card` CSS class has a shadow, you should add a bit of margin here (or padding to the parent) to prevent clipping the shadow.

src/Library/demos/Box/main.css
```diff
@@ -0,0 +1,7 @@
...
+}
+
+#border {
```
andyholmes, line 5: This would make more sense as a CSS class, since you are adding and removing it programmatically. You can use `Gtk.Widget.add_css_class()` and `Gtk.Widget.remove_css_class()` for that.

src/Library/demos/Box/main.js
```diff
@@ -0,0 +1,76 @@
+import Gtk from "gi://Gtk";
+
+const interactive_box = workbench.builder.get_object("interactive_box");
+const button_add = workbench.builder.get_object("button_add");
+const prepend = workbench.builder.get_object("prepend");
+const button_remove = workbench.builder.get_object("button_remove");
+var count = 0;
```
andyholmes, line 7:
```suggestion
let count = 0;
```

`var` isn't bad, but it is usually avoided in GJS unless the variable really needs to be hoisted.

src/Library/demos/Box/main.js
```diff
@@ -0,0 +1,76 @@
...
+
+button_add.connect("clicked", () => {
+  const label = new Gtk.Label({
+    name: "card",
+    label: `Item ${count + 1}`,
+  });
+  const style = label.get_style_context();
+  style.add_class("card");
```
andyholmes, line 15: This should just be added as a construct property, especially since style contexts will deprecated sometime soon.

```js
const label = new Gtk.Label({
  name: 'card',
  label: `Item ${count + 1}`,
  css_classes: ['card'],
});
```
`Gtk.Label` doesn't have any CSS classes by default, but for widgets that do you can use `GtkWidget.add_css_class()` to avoid overwriting existing classes.

andyholmes (changes requested):
This looks good and I especially like the UX of the slider to change the alignment.

src/Library/demos/Box/main.blp
```diff
@@ -0,0 +1,109 @@
...
+            label: "Add Item";
+            styles [
+              "suggested-action",
+              "pill",
+            ]
+          }
+          Button button_remove{
+            label: "Remove Item";
+            styles [
+              "destructive-action",
```
sonnyp, line 44: This is not a destructive action.

It doesn't destroy user data.

src/Library/demos/Box/main.blp
```diff
@@ -0,0 +1,109 @@
...
+        orientation: vertical;
+        spacing: 12;
+
+        Box {
+          halign: center;
+          spacing: 18;
+          Button button_add{
+            label: "Add Item";
+            styles [
+              "suggested-action",
```
sonnyp, line 37: I would drop `suggested-action` - it's not more or less important than the second

src/Library/demos/Box/main.blp
```diff
@@ -0,0 +1,109 @@
...
+            label: "Highlight Box";
+          }
+
+          CheckButton toggle_orient {
+            label: "Toggle Orientation";
+          }
+
+          CheckButton homogeneous {
+            label: "Homogeneous";
+          }
```
sonnyp, line 64: This won't do much in this demo since all items already have the same size anyway.

I'd suggest removing it

src/Library/demos/Box/main.blp
```diff
@@ -0,0 +1,109 @@
...
+          halign: center;
+          spacing: 18;
+
+          CheckButton highlight {
+            label: "Highlight Box";
+          }
+
+          CheckButton toggle_orient {
+            label: "Toggle Orientation";
+          }
```
sonnyp, line 60: Please use the the name of the property and the name of the possible values so it clears to users what we are changing here.

orientation: vertical|horizontal

I'd suggest using grouped toggle buttons (see https://github.com/sonnyp/Workbench/pull/219)

src/Library/demos/Box/main.blp
sonnyp: A `Scale` isn't really a good control for this. There are no values possible between the marks.

Same comment as above, I'd suggest using grouped toggle buttons and the exact property name / values (no upper casing)

sonnyp: Also, we need 1 for halign and 1 for valign

src/Library/demos/Box/main.blp
```diff
@@ -0,0 +1,109 @@
...
+
+          CheckButton homogeneous {
+            label: "Homogeneous";
+          }
+
+          CheckButton prepend {
+            label: "Prepend Items";
+          }
```
sonnyp, line 68: nice to have a control for this

I'd suggest using 3 buttons instead `Append Item`, `Prepend Item` and `Remove Item`

If you want - `Append Item` and `Prepend Item` can be merged using a dropdown button to let the user choose the mode but it's not necessary

image

sonnyp (commented):
Awesome to get an interactive Box demo :+1:

src/Library/demos/Box/main.js
AkshayWarrier: I was looking for something like this, thanks!

sonnyp: I had completely missed that in GTK 4

https://docs.gtk.org/gtk4/property.Widget.css-classes.html

TIL - thanks @andyholmes

src/Library/demos/Box/main.blp
AkshayWarrier: When there's space available, it will force the children to take all the available space and share it equally right?

image

image

 If you still think that it's not very useful to have I can remove it 👍🏾  

I understand that homogeneous is generally intended to be used with children that are taking uneven amounts of space and force them to take space equally.

sonnyp: `homogeneous` only guarantees that children will have the same dimensions. Perhaps we could keep it if `Append Item` / `Prepend Item` would create elements of slightly different dimensions 

but it might get confusing

let's leave it out for now and in a second iteration – we can make the children editable – then it will make sense to have a "homogeneous" control

src/Library/demos/Box/main.blp
```diff
@@ -0,0 +1,202 @@
...
+            ]
+          }
+        }
+      }
+
+        Box {
+          halign:center;
+          spacing:18;
+
+          Box{
```
andyholmes, line 79: Just a little layout tweak, to balance things out.

```suggestion
          Box{
            spacing: 18;
```

src/Library/demos/Box/main.blp
```diff
@@ -0,0 +1,202 @@
...
+              halign: center;
+
+              ToggleButton toggle_orientation_horizontal {
+                label: _("horizontal");
+                active: true;
+              }
+              ToggleButton toggle_orientation_vertical {
+                label: _("vertical");
+                group: toggle_orientation_horizontal;
+              }
```
andyholmes, line 96: This is a perfect time to use the `linked` class, which will join the two buttons together as a visibly logical unit :slightly_smiling_face:

src/Library/demos/Box/main.blp
```diff
@@ -0,0 +1,202 @@
...
+            }
+          }
+
+          CheckButton highlight{
+            label: _("Highlight Box");
+          }
+      }
+
+      Grid {
```
andyholmes, line 105: This is definitely the use for a grid, however I recommend you simplify the layout a bit.

I think this would best laid out so that (col, row):

* nothing at `0, 0`
* "halign" box is at `1, 0`
* "valign" box at `0, 1`
* `ScrolledWindow` at `1,1`

And just remove the `column-span`/`row-span` so they default to `1`. You can then move the `margin-end` from `ScrolledWindow` to `Grid` as necessary.

andyholmes (changes requested):
This really great for demonstrating both `GtkBox` and alignment/expansion, well done!

Just a few layout tweaks and I think this is good to go if @sonnyp agrees.

src/Library/demos/Box/main.blp
AkshayWarrier: Ah my bad, I might have accidentally removed it :/

src/Library/demos/Box/main.blp
AkshayWarrier: I actually chose that layout because I was trying to center all the options. But because of it, like you said I had to put a nasty ``margin-end: 114;`` on the ``ScrolledWindow``

image

With this layout
image

I'm not sure how to tackle this otherwise, unless centering that is not very important which is fine 👍🏾

AkshayWarrier: Okay this works 👍🏾

andyholmes (approved):
Great job!

sonnyp (approved):
Fantastic :+1: 

I added a couple of tooltips to clarify what "halign" / "valign" stands for

https://github.com/sonnyp/Workbench/pull/216/commits/6c984a1960efd4c4ad6e8ff24f955f56a3373d13

AkshayWarrier:
Thanks! 😄

PR #224, opened 2023-03-13
IconLibrary: Move the search bar to header

### LINKED ISSUE #207: Icon Library search looks weird
The search bar doesn't have a border, so the content scrolls into nothingness. I'd move it to the headerbar instead.

Capture d’écran du 2023-03-05 19-04-16

vixalien:
fixes #207


sonnyp (approved):
looks way better, thanks

PR #225, opened 2023-03-13
library: Add Date-Picker Entry.

B1TSH3LL:
Added DatePicker Entry/Demo under User Interface section in Library.
Issue https://github.com/sonnyp/Workbench/issues/69


src/Library/demos/Datepicker/index.js
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env gjs
+
+const {Gtk} = imports.gi;
+const DatePickerDemo = require('./main.js').DatePickerDemo;
```
sonnyp, line 4: GJS is not Node.js, there is no such thing as `require`.

This cannot work. Did you test your code at all?

sonnyp (changes requested):
This doesn't work, I cannot start workbench with your changes

> (re.sonny.Workbench.Devel:4): Gjs-CRITICAL **: 12:50:51.088: JS ERROR: Error: Invalid value 'undefined' for property title in object initializer.
_init/Gtk.Widget.prototype._init@resource:///org/gnome/gjs/modules/core/overrides/Gtk.js:55:50
Library/<@resource:///re/sonny/Workbench/src/Library/Library.js:26:20
Library@resource:///re/sonny/Workbench/src/Library/Library.js:25:9
Window@resource:///re/sonny/Workbench/src/window.js:431:10
@resource:///re/sonny/Workbench/src/application.js:28:14

Because your `main.json` is wrong. It looks like you tried to make a package.json file? Workbench/GNOME doesn't use Node.js.

Vala is not required — I suggest you leave it out for now.

Workench demos don't have an `index.js` file.

`main`.js doesn't follow the practices of our demos.

Did you manage to get Workbench running according to the documentation?  https://github.com/sonnyp/Workbench/blob/main/CONTRIBUTING.md

PR #227, opened 2023-03-14
Library: optimized notification

Aryanryn09:
this change is wrt This code uses object shorthand notation to set the properties of the Gio.Notification object, which makes the code more concise and easier to read. It also uses a loop to create and add the Gio.SimpleAction objects, which eliminates some repetition.issue #69

### DIFF
--- a/src/Library/demos/Notification/main.js
+++ b/src/Library/demos/Notification/main.js
@@ -3,37 +3,30 @@ import Gio from "gi://Gio";
 const { application, builder } = workbench;
 
 // https://gjs-docs.gnome.org/gio20/gio.notification
-const notification = new Gio.Notification();
-
-notification.set_title("Lunch is ready");
-notification.set_body(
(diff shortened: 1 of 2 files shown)


sonnyp (changes requested):
This doesn't work. Did you test?

https://github.com/sonnyp/Workbench/blob/main/CONTRIBUTING.md#submitting-a-contribution

> Make sure it's working by running Workbench and launching your entry via the Library.

With your changes and when loading the demo, the following appears in the console:

> (re.sonny.Workbench.Devel:4): Gjs-WARNING **: 13:02:03.231: JS ERROR: Error: No property title on GNotification
@file:///tmp/workbench-0B9V11.js:6:22

Also, the file `window.js` appears to be an error.

Please make sure to review and test your changes before submitting a contribution.

Aryanryn09:
ok i'll surely review it

PR #228, opened 2023-03-14
library: Add File Api Demo

SoNiC-HeRE:
Implemented a File Api Demo under API Section in Library


SoNiC-HeRE:
Planning to Add this demo under a different section in Library. Also will Implement some design changes to make it look somwhat similar to the design attached below. Looking for an early review for the code.

SoNiC-HeRE:
image

sonnyp (commented):
I couldn't run this. Please use a boxed list as discussed. 

https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.3/boxed-lists.html

SoNiC-HeRE:
Should've worked but I'll check what's wrong and yes implement the changes as discussed before

SoNiC-HeRE:
Improved the UI (used Boxed Lists) and made changes as discussed

sonnyp:
Coming back to this, I think we shouldn't think of this as a UI demo but API demo.

I'll create separate issues to add filesystem APIs demo.

SoNiC-HeRE:
> Coming back to this, I think we shouldn't think of this as a UI demo but API demo.
> 
> I'll create separate issues to add filesystem APIs demo.

How about creating a popup to display when the file is uploaded just to minimise ui ?

PR #231, opened 2023-03-15
Library: don't ignore errors from missing "main.json" files

andyholmes:
When loading files for a demo some missing files ignored because they're optional, but the `main.json` file is not.

Throw an error if `Gio.IOErrorEnum.NOT_FOUND` is hit in that particular case.

### DIFF
--- a/src/Library/Library.js
+++ b/src/Library/Library.js
@@ -101,7 +101,7 @@ function readDemoFile(demo_name, file_name) {
   try {
     str = decode(file.load_contents(null)[1]);
   } catch (err) {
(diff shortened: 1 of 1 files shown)


sonnyp:
Thanks - I think this fixes it https://github.com/sonnyp/Workbench/commit/a34ea0e334a2191fe06a0e88f358027f8c51fa09 by making it explicit to newcomers. 

We do want to ignore demo folders without `main.json` - it helps with builddir and switching branches somehow - at least on VSCode with Flatpak.

PR #232, opened 2023-03-15
library: Add Color Dialog entry

AkshayWarrier:
Adds Color Dialog entry.
Issue #69

### DIFF
--- a/src/Library/demos/Color Dialog/main.blp
+++ b/src/Library/demos/Color Dialog/main.blp
@@ -1,0 +1,51 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "Color Dialog";
+  description: "Show a dialog to select a color";
+
+  Box {
+    orientation: vertical;
+    halign: center;
+    spacing:42;
+
+    Box{
+      orientation:vertical;
+
+      ColorDialogButton color_dialog_button {
+        halign: center;
+        margin-bottom:12;
+      }
+      Label {
+        label: "Color Dialog Button";
+        margin-bottom:12 ;
+      }
+
+      LinkButton {
+        label: "API Reference";
+        uri: "https://docs.gtk.org/gtk4/class.ColorDialogButton.html";
+      }
+    }
+
+    Box{
+      orientation: vertical;
+
+      Button custom_button {
+        label: "Select Color…";
+        margin-bottom:12;
+        styles ["pill"]
+      }
+
+      Label {
+        label: "Custom Button";
+        margin-bottom:12 ;
+      }
+
+      LinkButton {
+        label: "API Reference";
+        uri: "https://docs.gtk.org/gtk4/class.ColorDialog.html";
+      }
+    }
+  }
+}
--- a/src/Library/demos/Color Dialog/main.blp
+++ b/src/Library/demos/Color Dialog/main.blp
@@ -3,7 +3,7 @@ using Adw 1;
 
 Adw.StatusPage {
   title: "Color Dialog";
-  description: "Show a dialog to select a color";
+  description: _("Show a dialog to select a color");
 
   Box {
     orientation: vertical;
@@ -18,7 +18,7 @@ Adw.StatusPage {
         margin-bottom:12;
       }
       Label {
-        label: "Color Dialog Button";
+        label: _("Color Dialog Button");
         margin-bottom:12 ;
       }
 
@@ -32,13 +32,13 @@ Adw.StatusPage {
       orientation: vertical;
 
       Button custom_button {
-        label: "Select Color…";
+        label: _("Select Color…");
         margin-bottom:12;
         styles ["pill"]
       }
 
       Label {
-        label: "Custom Button";
+        label: _("Custom Button");
         margin-bottom:12 ;
       }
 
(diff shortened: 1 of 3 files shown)


src/Library/demos/Color Dialog/main.blp
```diff
@@ -0,0 +1,51 @@
...
+
+  Box {
+    orientation: vertical;
+    halign: center;
+    spacing:42;
+
+    Box{
+      orientation:vertical;
+
+      ColorDialogButton color_dialog_button {
```
andyholmes, line 16: I think having the default color here `white` makes it hard to see what it is at a glance. Let's add a solid default color like blue, red, etc.

src/Library/demos/Color Dialog/main.blp
```diff
@@ -0,0 +1,51 @@
...
+        halign: center;
+        margin-bottom:12;
+      }
+      Label {
+        label: _("Color Dialog Button");
```
andyholmes, line 21:
```suggestion
        label: _("Dialog Button");
```

I'd also move these labels above their widgets and add something like the `title-2` class, so its clearer two different things are being demonstrated.

src/Library/demos/Color Dialog/main.blp
```diff
@@ -0,0 +1,51 @@
...
+      orientation: vertical;
+
+      Button custom_button {
+        label: _("Select Color…");
+        margin-bottom:12;
+        styles ["pill"]
+      }
+
+      Label {
+        label: _("Custom Button");
```
andyholmes, line 41:
```suggestion
        label: _("Dialog with Custom Button");
```

src/Library/demos/Color Dialog/main.js
```diff
@@ -0,0 +1,42 @@
...
+  title: "Select a color",
+  modal: true,
+  with_alpha: true,
+});
+
+color_dialog_button.dialog = dialog_standard;
+color_dialog_button.rgba = color;
+
+color_dialog_button.connect("notify::rgba", () => {
+  console.log(
```
andyholmes, line 23: If this is only for debugging use (i.e. you're not printed to `stdout` for another process to read), use `console.debug()` instead.

src/Library/demos/Color Dialog/main.js
```diff
@@ -0,0 +1,42 @@
...
+const dialog_custom = new Gtk.ColorDialog({
+  title: "Select a color",
+  modal: true,
+  with_alpha: false,
+});
+
+custom_button.connect("clicked", async () => {
+  try {
+    const color = await dialog_custom.choose_rgba(workbench.window, null, null);
+    console.log(`Custom Buton: The color selected is ${color.to_string()}`);
```
andyholmes, line 37:
```suggestion
    console.debug(`Custom Button: The color selected is ${color.to_string()}`);
```

src/Library/demos/Color Dialog/main.js
```diff
@@ -0,0 +1,42 @@
...
+  );
+});
+
+const dialog_custom = new Gtk.ColorDialog({
+  title: "Select a color",
+  modal: true,
+  with_alpha: false,
+});
+
+custom_button.connect("clicked", async () => {
```
andyholmes, line 34: Just a piece of advice: be careful about using `async` functions as signal callbacks.

An `async` function will implicitly return a `Promise`, and if the signal handler takes a `boolean` return value the `Promise` will be coerced to `true`.

andyholmes (changes requested):
Great demo, and good job confirming the bug in GJS while working through it.

Just a few cleanups and this will make a great demo of the high-level tools in the platform.

src/Library/demos/Color Dialog/main.js
AkshayWarrier: Okay, I wasn't aware of that. I only used ``console.log`` because that's what all the demos have been using even though they are not actually being read by another process. As a demo, I think it's okay to be more verbose? Since the main purpose of that snippet is to show how to retrieve the color from a dialog.

src/Library/demos/Color Dialog/main.js
AkshayWarrier: Oops :P

src/Library/demos/Color Dialog/main.blp
AkshayWarrier: What do you think of using a ``title-4`` instead of a ``title-2``?

image

image

AkshayWarrier: I think a small widget under a very big title looks a little off?

src/Library/demos/Color Dialog/main.js
andyholmes: Yup, lots of output is good. The different between `console.log()` and `console.debug()` is that debug is usually filtered out for users, but easy to turn on for developers.

No big deal though, you're right this whole application is for developers :)

src/Library/demos/Color Dialog/main.blp
andyholmes: Yeah, that does look better :+1:

andyholmes (approved):
LGTM

PR #233, opened 2023-03-15
library: Add Grid entry

bazylevnik0:


sonnyp (changes requested):
Good start!

We try to make Library entries nice to look at and interactive (but still comprehensible).

What about a (very) simple tic tac toe with a 3x3 grid? Since it will require some JavaScript it's also a good opportunity to showcase the APIs to work with grid.

Don't worry too much about style, we can iterate later.

src/about.js
```diff
@@ -56,6 +56,7 @@ ${getBlueprintVersion()}
     "Hari Rana (TheEvilSkeleton) https://theevilskeleton.gitlab.io",
     "Sriyansh Shivam https://linktr.ee/sonic_here",
     "Angelo Verlain https://www.vixalien.com",
+    "bazylevnik0 https://github.com/AkshayWarrier",
```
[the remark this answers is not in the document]
bazylevnik0: Yep :)

src/Library/demos/Grid/main.blp
```diff
@@ -0,0 +1,71 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "Grid";
+  description: _("Arrange widgets in rows and columns.\n\n*Tic Tac Toe: just click to the '#' for 'x' and click again for 'o':");
```
sonnyp, line 6:
```suggestion
  description: _("Arrange widgets in rows and columns.");
```

hopefully we won't need to explain the UI :)

sonnyp (commented):
I should have been more explicit. With tic tac toe - I meant as a (very simple) game you can play against the computer.

Similar to https://playtictactoe.org/

For the UI please make each cell square.
You can use `circle-outline-thick-symbolic` and `cross-large-symbolic` icons

If you make it look like a tic tac toe - we don't need to explain :)

src/Library/demos/Grid/main.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "Grid",
+  "category": "user_interface",
+  "description": "Allow people to perform using a grids",
+  "panels": [
+    "code",
+    "style",
+    "ui",
+    "preview"
```
sonnyp, line 9: Should be enough to see quickly what Grid is about:

```suggestion
    "ui",
    "preview"
```

sonnyp (commented):
Cool.

Regarding the design:

1. No need for edit mode
2. Have one of the cell by default O or X so that it's clear for the user what they're looking at
3. Make the cells square
4. Only show big  O / X, no background

The icons should be visible and light and dark modes - if you stick to the default color that should work fine

src/Library/demos/Grid/main.css
```diff
@@ -0,0 +1,27 @@
...
+#button01:hover,
+#button02:hover,
+#button10:hover,
+#button11:hover,
+#button12:hover,
+#button20:hover,
+#button21:hover,
+#button22:hover {
+  box-shadow: inset 0px 0px 65px #ff1493;
+}
```
sonnyp, line 26: the default `hover` is fine, you can remove this

```suggestion
```

src/Library/demos/Grid/main.css
```diff
@@ -0,0 +1,27 @@
+#button00,
+#button01,
+#button02,
+#button10,
+#button11,
+#button12,
+#button20,
+#button21,
+#button22 {
```
sonnyp, line 9: in CSS, you should use a class to to make styling reusable between different elements

```suggestion
.cell {
```

You can apply the class to each button using the `styles ['cell']` syntax

And remove `name` on all `button`

src/Library/demos/Grid/main.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "Grid",
+  "category": "user_interface",
+  "description": "Allow people to perform using a grids",
```
sonnyp, line 4:
```suggestion
  "description": "Arrange widgets in rows and columns.",
```

src/Library/demos/Grid/main.js
```diff
@@ -0,0 +1,53 @@
...
+}
+
+let step = 1;
+function onClicked(button) {
+  //check access for user action
+  let clicked = button.get_child();
+  if (
+    clicked.icon_name !== "cross-large-symbolic" &&
+    clicked.icon_name !== "circle-outline-thick-symbolic"
+  ) {
```
sonnyp, line 25: an early return would make the function easier to read and reduce one level of indentation

something like

```suggestion
  if (clicked.icon_name) return;
```

src/Library/demos/Grid/main.js
```diff
@@ -0,0 +1,53 @@
...
+
+for (const id of button_ids) {
+  const button = workbench.builder.get_object(id);
+  button.connect("clicked", onClicked);
+}
+
+let step = 1;
+function onClicked(button) {
+  //check access for user action
+  let clicked = button.get_child();
```
sonnyp, line 21: let's name it after what it is for clarity

```suggestion
  let image = button.get_child();
```

src/Library/demos/Grid/main.js
```diff
@@ -0,0 +1,53 @@
...
+  if (
+    clicked.icon_name !== "cross-large-symbolic" &&
+    clicked.icon_name !== "circle-outline-thick-symbolic"
+  ) {
+    //store and show user action
+    clicked.icon_name = "cross-large-symbolic";
+    //calculate pc reaction
+    let pc_is_thinking = true;
+    let pc_is_thinking_row;
+    let pc_is_thinking_col;
```
sonnyp, line 31: these don't need to be scope outside the `while`

src/Library/demos/Grid/main.js
```diff
@@ -0,0 +1,53 @@
...
+      pc_is_thinking_col = "" + Math.floor(Math.random() * 3);
+      //make pc reaction if possible
+      let temp = workbench.builder.get_object(
+        "button" + pc_is_thinking_row + pc_is_thinking_col,
+      );
+      let searched = temp.get_child();
+      if (
+        searched.icon_name !== "cross-large-symbolic" &&
+        searched.icon_name !== "circle-outline-thick-symbolic"
+      ) {
```
sonnyp, line 43: ditto

src/Library/demos/Grid/main.blp
```diff
@@ -0,0 +1,90 @@
...
+  description: _("Arrange widgets in rows and columns.");
+
+ Box {
+    halign: center;
+    Grid {
+        Button button00 {
+        name: "button00";
+        layout { row: 0; column: 0; }  
+        hexpand: false;
+        Image image00{
```
sonnyp, line 15:
```suggestion
        Image {
```

you can remove all ids on image, they're not used

sonnyp (changes requested):
Looks great now!

Just a few more changes and this is ready to go.

Try to make sure the code is always easy to reason about.

sonnyp (approved):
The point if early return is that you don't need the `else` :)

I fixed it in https://github.com/sonnyp/Workbench/pull/233/commits/810fadfe75bfd6fc5de85e4c7386b465d7c73d49

Thanks for the contribution!

PR #236, opened 2023-03-16
library: Add ViewSwitcher entry

halfmexican:
Simple blueprint demo showing off the ViewSwitcher Pattern

### DIFF
--- a/src/Library/demos/View Switcher/main.blp
+++ b/src/Library/demos/View Switcher/main.blp
@@ -27,33 +27,24 @@ Adw.Window {
         icon-name: "star-filled-rounded-symbolic";
         use-underline: true;
         child:
-        Box{
-          orientation: vertical;
-          valign: center;
+        Adw.StatusPage{
+          title: bind page1.title;
+          icon-name: bind page1.icon-name;
+          child:
+          
+          Box{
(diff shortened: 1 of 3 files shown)


src/Library/demos/View Switcher/main.blp
```diff
@@ -0,0 +1,150 @@
...
+
+      Adw.ViewStackPage page3 {
+        name: "page3";
+        title: _("Notifications");
+        icon-name: "bell-symbolic";
+        badge-number: 3;
+        needs-attention: true;
+        use-underline: true;
+        child:
+        Box{
```
andyholmes, line 94: This demo would be much more helpful with a bit of code showing how to control the `Adw.ViewSwitcher` appearance by modifying the `needs-attention` and `badge-number` properties.

Something like a `Gtk.ListBox` with a number of rows matching the `badge-number`, which is updates as rows are removed (i.e. notifications marked as read).

andyholmes (changes requested):
Looks good so far! Demonstrate a little more how to interact with the `Adw.ViewSwitcher` and its good example of using the platform.

andyholmes (approved):
Very nicely done!

PR #238, opened 2023-03-17
library: Add Level Bars entry

AkshayWarrier:
Adds level bars entry.
- For continuous mode, I've tried to replicate GNOME's battery status widget
- For discrete mode, I have tried my best to eliminate edge cases. I have kept the criteria for password strength fairly loose i.e it's easy to make a strong password. This is mainly because I want all the levels to be discovered quickly. But this can be changed as per discussion.
- The colors for password strength, I haven't done separate styling for light mode/dark mode but I have chosen colors that should be equally visible in both. I can change this if need be.

Issue #69

### DIFF
--- a/src/Library/demos/Level Bars/main.blp
+++ b/src/Library/demos/Level Bars/main.blp
@@ -8,6 +8,7 @@ Adw.StatusPage {
   Box {
     orientation: vertical;
     margin-top: 24;
+    halign: center;
 
     Box {
       orientation: vertical;
@@ -21,12 +22,12 @@ Adw.StatusPage {
 
       Box {
         orientation: vertical;
-        spacing:12;
+        spacing: 12;
 
         Label {
           halign: start;
           margin-start: 12;
-          margin-top:12;
+          margin-top: 12;
           label: _("Battery");
         }
 
@@ -40,6 +41,7 @@ Adw.StatusPage {
             value: 50;
         }
         styles ["card"]
+        margin-bottom: 24;
       }
 
 
@@ -77,4 +79,3 @@ Adw.StatusPage {
     }
   }
 }
-
--- a/src/Library/demos/Level Bars/main.blp
+++ b/src/Library/demos/Level Bars/main.blp
@@ -1,0 +1,130 @@
+using Gtk 4.0;
+using Adw 1;
+Adw.Clamp{
+  Box {
+    orientation: vertical;
+    margin-top: 24;
+
+    Label {
+      label: "Level Bars";
+      margin-top: 6;
+      margin-bottom: 6;
+
+      styles [
+        "title-1",
+      ]
+    }
+
+    Label {
+      label: _("A bar widget used as a level indicator");
+      margin-top: 6;
+      margin-bottom: 24;
+    }
+
+    Box {
+      orientation: vertical;
+      vexpand: true;
+      spacing: 18;
+      margin-start: 18;
+      margin-end: 18;
+
+      Label {
+        label: _("Continuous Level Bar");
+      }
+
+      Box{
+        spacing:18;
+        halign: start;
+
+        Label {
+          label: _("Current Battery Level");
+        }
+
+        SpinButton spin_button{
+          digits:1;
+          adjustment: Gtk.Adjustment {
+            lower: 0;
+            upper: 100;
+            value:100;
+            step-increment: 5;
+          };
+        }
+      }
+
+      Box {
+        orientation: vertical;
+        spacing:12;
+
+        LevelBar bar_continuous{
+            margin-end:12;
+            margin-start:12;
+            margin-top:24;
+            min-value:0;
+            max-value:100;
+            mode: continuous;
+            value: bind spin_button.value;
+        }
+
+        Box {
+          margin-start:12;
+          margin-end:12;
+          margin-bottom:12;
+          Label{
+            label: _("Battery");
+          }
+          Label percentage_label{
+            label: "100%";
+            hexpand: true;
+            halign: end;
+          }
+        }
+        styles ["card"]
+      }
+    }
+
+    Box {
+      orientation: vertical;
+      vexpand: true;
+      spacing: 18;
+      margin-start: 18;
+      margin-end: 18;
+
+      Label {
+        label: _("Discrete Level Bar");
+      }
+
+      Label {
+        halign: start;
+        label: _("A good password uses uppercase and lowercase letters, special characters, digits and\nis atleast 8 characters long.");
+        styles ["dim-label"]
+      }
+      PasswordEntry entry {
+        placeholder-text: "Enter password";
+        show-peek-icon: true;
+      }
+      LevelBar bar_discrete {
+        min-value: 0;
+        max-value: 6;
+        mode: discrete;
+      }
+      Label label_strength{
+
+      }
+    }
+
+    Box {
+      orientation: vertical;
+      margin-bottom:24;
+
+      LinkButton {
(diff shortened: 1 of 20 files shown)


src/Library/demos/Level Bars/main.blp
```diff
@@ -0,0 +1,130 @@
...
+        SpinButton spin_button{
+          digits:1;
+          adjustment: Gtk.Adjustment {
+            lower: 0;
+            upper: 100;
+            value:100;
+            step-increment: 5;
+          };
+        }
+      }
```
sonnyp, line 52: I don't think we need this

it makes the UI harder to read and changing the value of the level is just one property away 

wdyt?

src/Library/demos/Level Bars/main.blp
```diff
@@ -0,0 +1,130 @@
...
+        spacing:12;
+
+        LevelBar bar_continuous{
+            margin-end:12;
+            margin-start:12;
+            margin-top:24;
+            min-value:0;
+            max-value:100;
+            mode: continuous;
+            value: bind spin_button.value;
```
sonnyp, line 65: yes this makes it difficult to grasp for first-timer

and the LevelBar demo should work when copy-pasted somewhere else
use a fixed value like 75%

src/Library/demos/Level Bars/main.blp
```diff
@@ -0,0 +1,130 @@
...
+          Label{
+            label: _("Battery");
+          }
+          Label percentage_label{
+            label: "100%";
+            hexpand: true;
+            halign: end;
+          }
+        }
+        styles ["card"]
```
sonnyp, line 81: nice touch :+1:

src/Library/demos/Level Bars/main.blp
```diff
@@ -0,0 +1,130 @@
...
+
+    Box {
+      orientation: vertical;
+      vexpand: true;
+      spacing: 18;
+      margin-start: 18;
+      margin-end: 18;
+
+      Label {
+        label: _("Continuous Level Bar");
```
sonnyp, line 32:
```suggestion
        label: _("Continuous");
```

src/Library/demos/Level Bars/main.blp
```diff
@@ -0,0 +1,130 @@
...
+
+        LevelBar bar_continuous{
+            margin-end:12;
+            margin-start:12;
+            margin-top:24;
+            min-value:0;
+            max-value:100;
+            mode: continuous;
+            value: bind spin_button.value;
+        }
```
sonnyp, line 66: can you add the mode explicitly? even if it's not required / default

src/Library/demos/Level Bars/main.blp
```diff
@@ -0,0 +1,130 @@
...
+        label: _("Discrete Level Bar");
+      }
+
+      Label {
+        halign: start;
+        label: _("A good password uses uppercase and lowercase letters, special characters, digits and\nis atleast 8 characters long.");
+        styles ["dim-label"]
+      }
+      PasswordEntry entry {
+        placeholder-text: "Enter password";
```
sonnyp, line 102:
```suggestion
        placeholder-text: "Choose a Password";
```

src/Library/demos/Level Bars/main.blp
```diff
@@ -0,0 +1,130 @@
...
+
+      Label {
+        label: _("Discrete Level Bar");
+      }
+
+      Label {
+        halign: start;
+        label: _("A good password uses uppercase and lowercase letters, special characters, digits and\nis atleast 8 characters long.");
+        styles ["dim-label"]
+      }
```
sonnyp, line 100: I think we don't need this either

src/Library/demos/Level Bars/main.js
```diff
@@ -0,0 +1,82 @@
...
+Very Weak- 1-3 characters long, doesn't use all valid characters
+Weak     - >=4 characters long, doesn't use all valid characters
+Moderate - 5-7 characters long, uses all valid characters once
+Strong   - >=8 characters long, uses all valid characters once
+
+Valid characters include uppercase and lowercase letters,
+special characters and digits
+*/
+
+const evaluate_strength = [
```
sonnyp, line 35: This is too complicated for a demo of Gtk.LevelBar
I think using password strength for the demo is a good idea, use the password length only :pray: 

Also https://xkcd.com/936/ :D
In general – better to use ready-made solutions for this kind of security sensitive things. For example https://github.com/dropbox/zxcvbn

sonnyp (changes requested):
Looks very good

src/Library/demos/Level Bars/main.blp
AkshayWarrier: "Choose a Password" or "Choose a password"?

src/Library/demos/Level Bars/main.blp
AkshayWarrier: I have? Unless you mean something else. I can move this up ``mode: continuous`` to make it the first property if you want.

src/Library/demos/Level Bars/main.blp
AkshayWarrier: The only problem is I won't be able to synchronize the ``percentage_label`` and level bar value when the UI is changed. They'll have to run the demo again if that's fine.

src/Library/demos/Level Bars/main.js
AkshayWarrier: I forgot about that xkcd comic :)

src/Library/demos/Level Bars/main.blp
sonnyp: Choose a Password

src/Library/demos/Level Bars/main.blp
sonnyp: Oops, i missed it.

src/Library/demos/Level Bars/main.blp
sonnyp: Do we need the % label?

AkshayWarrier: Well I wanted it to make it look like a battery widget similar to GNOME's to show how it's actually being used. I can remove it if you want.

src/Library/demos/Level Bars/main.blp
sonnyp: please don't mark conversation as resolved if you didn't reply or address the comment

src/Library/demos/Level Bars/main.blp
sonnyp: It's not about what I want - it's about pros and cons

Given the goal of Workbench/the project it's more important for the user to understand and be able to copy paste the Blueprint quickly rather than for the UI to show a spin button and label.

It still looks clear/good IMHO:

image

src/Library/demos/Level Bars/main.blp
sonnyp: Apologies – I forgot to pull, it's all good.

src/Library/demos/Level Bars/main.blp
AkshayWarrier: Hmm I'll keep that in mind 👍🏾 

I guess it's trying to find the balance between trying to make the demo easily copy-pastable and free of bloat while also making sure it looks good.

sonnyp (approved):
Thanks!

I made slight changes to use `Adw.StatusPage`, simplify a bit and warn about the password strength method being insecure to be sincere with users.

AkshayWarrier:
Thanks :)

PR #239, opened 2023-03-18
library: Add Calendar Entry

SoNiC-HeRE:
Added Calendar Demo under User Interface Section in Library
Issue #69

### DIFF
--- a/src/Library/demos/Calendar/main.blp
+++ b/src/Library/demos/Calendar/main.blp
@@ -1,0 +1,30 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "Calendar";
+  description: "Keep track of current or upcoming events";
+
+    Box {
+      orientation: vertical;
+      halign: center;
+
+    Calendar calendar {
+      show-day-names: true;
+      show-week-numbers: true;
+      show-heading: true;
+      margin-bottom: 18;
+    }
+
+    LinkButton {
+      label: "API Refernce";
+      uri: "https://docs.gtk.org/gtk4/class.Calendar.html";
+   }
+  }
+ }
+}
+
+
+
+
+
--- a/src/Library/demos/Calendar/main.blp
+++ b/src/Library/demos/Calendar/main.blp
@@ -3,43 +3,23 @@ using Adw 1;
 
 Adw.StatusPage {
   title: "Calendar";
-  description: "Widget that helps you keep track of events";
+  description: "Display a Gregorian calendar, one month at a time";
 
-    Box {
-      orientation: vertical;
-      halign: center;
+  Box {
+    orientation: vertical;
+    halign: center;
 
     Calendar calendar {
       show-day-names: true;
-      show-week-numbers: true;
+      show-week-numbers: false;
       show-heading: true;
       margin-bottom: 18;
-    }
-
-    Box {
-      spacing: 6;
-      halign: center;
-      margin-bottom: 12;
-
-      Button mark {
-        icon-name: "plus";
-      }
-
-      Button remove_mark {
-        icon-name: "minus";
-      }
+      day: 1;
     }
 
     LinkButton {
-      label: "API Refernce";
+      label: "API Reference";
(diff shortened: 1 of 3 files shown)


src/Library/demos/Calendar/main.blp
```diff
@@ -0,0 +1,30 @@
...
+      margin-bottom: 18;
+    }
+
+    LinkButton {
+      label: "API Refernce";
+      uri: "https://docs.gtk.org/gtk4/class.Calendar.html";
+   }
+  }
+ }
+}
```
SoNiC-HeRE, line 25: will remove this bracket its unused

sonnyp: this is a syntax error - please make sure your contributions run before requesting review

src/Library/demos/Calendar/main.blp
```diff
@@ -0,0 +1,30 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "Calendar";
+  description: "Keep track of current or upcoming events";
```
sonnyp, line 6: that'd make sense as a description for an end user calendar app but not here

I'll let you come up with a new suggestion

src/Library/demos/Calendar/main.js
```diff
@@ -0,0 +1,27 @@
...
+});
+calendar.connect("next-year", () => {
+  console.log("Switched to Next Year");
+});
+calendar.connect("prev-month", () => {
+  console.log("Switched to Previous Month");
+});
+calendar.connect("prev-year", () => {
+  console.log("Switched to Previous Year");
+});
```
sonnyp, line 27: I can't really think of use case for these signals

I'd remove them - @andyholmes wdyt?

sonnyp: maybe we can connect to property changes instead

```
calendar.connect('notify::month', () => {
  console.log(calendar.month);
});
```

src/Library/demos/Calendar/main.js
```diff
@@ -0,0 +1,27 @@
...
+function handleChange() {
+  let year = calendar.year;
+  let month = calendar.month;
+  let day = calendar.day;
+  let monthString = month.toString().padStart(2, "0");
+  let dayString = day.toString().padStart(2, "0");
+
+  console.log(
+    `Selected Date (YYYY-MM-DD): ${year}-${monthString}-${dayString}`,
+  );
```
sonnyp, line 12: please use `getDate` instead - it's the perfect opportunity to showcase the DateTime API :)

https://docs.gtk.org/glib/method.DateTime.format.html

Please use this to display the date:

> \%x: the preferred date representation for the current locale without the time

sonnyp (changes requested):
Could you also put an example of calendar marks?

src/Library/demos/Calendar/main.js
SoNiC-HeRE: but it would return an integer , i don't think that would be relevant to log.

src/Library/demos/Calendar/main.js
SoNiC-HeRE: sure

src/Library/demos/Calendar/main.blp
SoNiC-HeRE: How about : an arrangement of time into days, weeks, months, and years?

src/Library/demos/Calendar/main.js
sonnyp: You're right. What about we log the day/month/year in the user locale?

SoNiC-HeRE:
Sure I'd do that

sonnyp (approved):
Thanks

I made some slight adjustments https://github.com/sonnyp/Workbench/pull/239/commits/40edf52c2d93c9de82e92dc07b58f673fc271def

The mark UI was a bit too complicated and unclear because you had to select a day to mark the day but the day selection style would make the mark invisible for that day.

PR #243, opened 2023-03-21
Library: Add Banner Entry

Kodecheff:
Fixes https://github.com/sonnyp/Workbench/issues/285

### DIFF
--- a/src/Library/demos/Banner/main.blp
+++ b/src/Library/demos/Banner/main.blp
@@ -1,0 +1,24 @@
+using Gtk 4.0;
+
+Box welcome {
+  orientation: vertical;
+  valign: center;
+  halign: center;
+
+  Label {
+    label: "A widget for communicating persistent and important information.";
+    margin-bottom: 20;
+  }
+
+  Box banner_box {
+    orientation: vertical;
+    valign: center;
+    halign: center;
+
+  Button open_banner{
+    label: "Open Banner";
+  }
+
+  }
+
+}
--- a/src/Library/demos/Banner/main.blp
+++ b/src/Library/demos/Banner/main.blp
@@ -1,24 +1,46 @@
 using Gtk 4.0;
+using Adw 1;
 
-Box welcome {
-  orientation: vertical;
-  valign: center;
-  halign: center;
+Gtk.Window window {
+  default-width: 800;
+  default-height: 600;
+  title: "Banner";
+  titlebar: Gtk.HeaderBar {
+    [end]
+    MenuButton button_menu {
+      icon-name: "open-menu-symbolic";
+      primary: true;
+    }
+  };
 
-  Label {
-    label: "A widget for communicating persistent and important information.";
-    margin-bottom: 20;
+  Adw.ToastOverlay overlay {
+    Adw.StatusPage {
+      title: "Adw Banner";
+      description: "Gnome Banner Test";
+
+      Box {
+        orientation: vertical;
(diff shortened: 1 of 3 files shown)


src/Library/demos/Banner/main.js
```diff
@@ -0,0 +1,42 @@
+import Gtk from "gi://Gtk?version=4.0";
+const Adw = imports.gi.Adw;
+Gtk.init();
+
+const _bannerBox = workbench.builder.get_object("banner_box");
+const _openBanner = workbench.builder.get_object("open_banner");
+
+function adwBanner() {
+  // Create a new AdwBanner widget
+  const banner = new Adw.Banner({
```
sonnyp, line 10: Similar to https://github.com/sonnyp/Workbench/pull/248

please create the banner from the blueprint

andyholmes: Since these are generally intended to be place in a window, directly under the header bar, it would be a good idea to create a small window to demonstrate this, too.

The button in Workbench could pop up a window, and contain a few widget to control the info bar's properties.

sonnyp: Agree with Andy, the Banner should be in a Window.

No need for a custom button to open the window though - Workbench will directly offer the option if it detects a `Window` in the Blueprint. Similar to the "Window" library entry.

sonnyp: Actually, the libadwaita demo doesn't use an external banner for its demo – I think it's safe to follow them and makes the demo visible immediately.

image

works for you @andyholmes ?

andyholmes: Yes, that's good too. I think the important part is that it doesn't have margins on the top or left/right sides.

Replicating the libadwaita demo is good idea in general I think.

src/Library/demos/Banner/main.blp
```diff
@@ -0,0 +1,46 @@
...
+        halign: center;
+        valign: center;
+
+      Adw.Banner banner{
+        button-label: "Troubleshoot";
+        title: "An error occurred: Could not resolve host.";
+        revealed: true;
+        use-markup: true;
+        margin-bottom: 20;
+      }
```
andyholmes, line 32: This is better, but the banner should really be up at the top of the window, with no margins, right under the header bar.

Then maybe make it so clicking "Troubleshoot" hides the banner and shows the toast. Where the banner is now, add a button so the banner can be re-opened.

src/Library/demos/Banner/main.blp
```diff
@@ -0,0 +1,46 @@
...
+    MenuButton button_menu {
+      icon-name: "open-menu-symbolic";
+      primary: true;
+    }
+  };
+
+  Adw.ToastOverlay overlay {
+    Adw.StatusPage {
+      title: "Adw Banner";
+      description: "Gnome Banner Test";
```
andyholmes, line 19:
```suggestion
      description: "GNOME Banner Test";
```

andyholmes (changes requested):
Looking better!

Be sure to mark all user-visible strings in the `.js` and `.blp` file as translatable, by putting them in the gettext function, like `_("Words to translate")`.

sonnyp:
hey @Kodecheff, do you want to finish this? If so, make sure to add your name to the list of contributors in about.js

sonnyp:
https://github.com/sonnyp/Workbench/pull/313

PR #244, opened 2023-03-22
library: Add Font Dialog Entry

SoNiC-HeRE:
Adds Font Dialog Entry
#69


src/Library/demos/Font Dialog/main.js
```diff
@@ -0,0 +1,44 @@
...
+  modal: true,
+});
+font_dialog_button.set_dialog(dialog_standard);
+
+font_dialog_button.connect("notify::font-desc", () => {
+  let fontname = font_dialog_button.get_font_desc().to_string();
+  console.log(`Font Dialog Button: The font selected is ${fontname}`);
+});
+
+const dialog_custom = new Gtk.FontChooserDialog({
```
sonnyp, line 25: this is the deprecated API

sonnyp (changes requested):
The entry doesn't appear in the library

> (re.sonny.Workbench.Devel:4): re.sonny.Workbench.Devel-WARNING **: 19:53:44.496: The demo name "Font Chooser" does not match the folder name "Font Dialog" and will be ignored.

SoNiC-HeRE:
I'll fix it

src/Library/demos/Font Dialog/main.js
SoNiC-HeRE: okay, I'll update this with the new one

sonnyp:
@SoNiC-HeRE please make sure to update this to use `FontDialog` instead of the deprecated `FontChooserDialog`

SoNiC-HeRE:
I'm unsure why I'm getting an empty object or undefined value when i'm trying to get the value from Gtk.FontDialog. It seeems to work fine with Gtk.FontDialogButton but not with a custom button. I'm using this for reference: https://docs.gtk.org/gtk4/class.FontDialog.html

SoNiC-HeRE:
The demo seems to be working now , had to fix an error. It is ready for review now

SoNiC-HeRE:
The demo seems to be working now , had to fix an error. It is ready for review now

sonnyp (approved):
:+1: 

made some minor changes https://github.com/sonnyp/Workbench/pull/244/commits/23210020aff341dea2dab031390a294cd0153148

PR #245, opened 2023-03-22
library: Add Alert Dialog Entry

SoNiC-HeRE:
Adds Alert Dialog Entry
Issue #69

### DIFF
--- a/src/Library/demos/Alert Dialog/main.blp
+++ b/src/Library/demos/Alert Dialog/main.blp
@@ -1,0 +1,28 @@
+using Gtk 4.0;
+using Adw 1;
+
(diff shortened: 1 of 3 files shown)


sonnyp:
I didn't think about that before but I'm not sure if we want to advertise `Gtk.AlertDialog` since `Adw.MessageDialog` is a better choice and does the same thing. I'll leave this open for now and we can think about it.

sonnyp:
Ok I confirmed with Alice that there is no reason to use Gtk.AlertDialog with libadwaita. We alreadty have Adw.MessageDialog entry so I will close this.

PR #251, opened 2023-03-26
library: Add Overlay entry

AkshayWarrier:
Adds overlay entry
Issue #69


AkshayWarrier:
I'm unsure on how I should demonstrate interaction with the background with a ``Gtk.Picture``
1) What  should happen when someone clicks/hovers on the picture
2) How do I listen for that event on a ``Gtk.Picture``

Right now I have added a simple darkening effect to the osd on hover

src/Library/demos/Overlay/main.js
```diff
@@ -0,0 +1,10 @@
+import Gtk from "gi://Gtk";
+import Gio from "gi://Gio";
+import GLib from "gi://GLib";
+
+const file = Gio.File.new_for_path("image.png");
```
sonnyp, line 5: doesn't work — please use the example I shared with you a while ago

src/Library/demos/Overlay/main.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "Overlay",
+  "category": "user_interface",
+  "description": "Overlay widgets on top of a main widget",
```
sonnyp, line 4:
```suggestion
  "description": "Overlay widgets on top of each others",
```

src/Library/demos/Overlay/main.blp
```diff
@@ -0,0 +1,56 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "Overlay";
+  description: "Overlay widgets on top of a main widget";
+
+  Adw.Clamp{
```
sonnyp, line 8: doesn't `StatusPage` already clamp?

sonnyp (commented):
> What should happen when someone clicks/hovers on the picture

yeah it's not super straightforward - let's leave it out for now - we can always improve this demo later :+1:

sonnyp:
For later here is an interesting use case

[Screencast from 2023-03-26 14-47-31.webm](https://user-images.githubusercontent.com/19673/227776774-386c3dd6-41a4-4e79-a78a-bd08ed60aa05.webm)

src/Library/demos/Overlay/main.blp
AkshayWarrier: I think it doesn't clamp enough, otherwise it looks like this

image

Or do you suggest something else?

sonnyp: ha I see 

fine as is :+1:

src/Library/demos/Overlay/main.js
sonnyp: It depends on what the working directory is (pwd) so it's fragile

PR #253, opened 2023-03-27
Library: add Drag-and-Drop entry

halfmexican:
Currently just drafting this as i'm having difficultly porting my drag and drop to javascript from vala and school work is eating my schedule


sonnyp:
Related https://matrix.to/#/!pOGgkAueAsusiyFCTb:matrix.org/$16828312569XwarW:nekover.se?via=libera.chat&via=matrix.org&via=gnome.org

sonnyp:
Any chance you could pick this up again? It would be great to have this.

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,60 @@
...
+import GObject from "gi://GObject";
+
+const list = workbench.builder.get_object("list");
+
+// Calculate the length of our ListBox
+let last_row = list.get_last_child();
+let list_length = last_row.get_index() + 1;
+
+// Create a Dragsource and DropTarget for each row
+for (let i = 0; i < list_length; i++) {
```
andyholmes, line 13: First thing that will help here, is to cleanup the variable names and scope. Use `const` wherever possible and I'd suggest `dragSource` and `dropTarget` (`_dropTarget` in callback signatures if you like).

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,60 @@
...
+  let row = list.get_row_at_index(i);
+  row.selectable = false;
+  let _drag_x;
+  let _drag_y;
+
+  let allocation;
+  const drag = new Gtk.DragSource();
```
andyholmes, line 20:
```suggestion
  const drag = new Gtk.DragSource({
    actions: Gdk.DragAction.MOVE,
  });
```

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,60 @@
...
+  const drop = Gtk.DropTarget.new(row.constructor.$gtype, Gdk.DragAction.COPY);
```
andyholmes, line 21:
```suggestion
  const drop = Gtk.DropTarget.new(row.constructor.$gtype, Gdk.DragAction.MOVE);
```

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,60 @@
...
+
+  row.add_controller(drag);
+  row.add_controller(drop);
+
+  // Drag controller
+  drag.connect("prepare", (source, _x, _y) => {
+    _drag_x = _x;
+    _drag_y = _y;
+
+    allocation = row.get_allocation();
```
andyholmes, line 31: Definitely do this in `drag-being`, right before you construct the widget you will use it for. Make it `const` and local to the callback, as well.

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,60 @@
...
+    let value = new GObject.Value();
+    value.init(GObject.TYPE_OBJECT);
```
andyholmes, line 33:
```suggestion
    value.init(Gtk.ListBoxRow);
```

This should be as specific as possible, and we know the widget must be a `Gtk.ListBoxRow` because it's a child of a `Gtk.ListBox`.

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,60 @@
...
+    value.set_object(row);
+    return Gdk.ContentProvider.new_for_value(value);
+  });
+
+  drag.connect("drag-begin", (drag) => {
```
andyholmes, line 38: Rename the arguments here to something like `(_dragSource, drag) => { }`.

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,60 @@
...
+    drag_widget.drag_highlight_row(drag_row);
+
+    let icon = Gtk.DragIcon.get_for_drag(drag.get_drag());
+    icon.child = drag_widget;
+  });
+
+  // Drop controller
+  drop.connect("drop", (drop, value, _x, _y) => {});
+
+  drop.connect("enter", (_x, _y) => {});
```
andyholmes, line 59:
```suggestion
```

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,60 @@
...
+    drag_widget.add_css_class("boxed-list");
+
+    let drag_row = new Adw.ActionRow();
+    drag_row.title = row.title;
+    drag_row.add_css_class("boxed-list");
+
+    drag_widget.append(drag_row);
+    drag_widget.drag_highlight_row(drag_row);
+
+    let icon = Gtk.DragIcon.get_for_drag(drag.get_drag());
```
andyholmes, line 52: Use `drag` (a `Gtk.Drag` boxed type) from the callback arguments.

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,60 @@
...
+    allocation = row.get_allocation();
+    let value = new GObject.Value();
+    value.init(GObject.TYPE_OBJECT);
+    value.set_object(row);
+    return Gdk.ContentProvider.new_for_value(value);
+  });
+
+  drag.connect("drag-begin", (drag) => {
+    list.remove(row);
+    row.set_state_flags(Gtk.StateFlags.DROP_ACTIVE, true);
```
andyholmes, line 40: Is this necessary? I assumed GTK would do this automatically.

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,60 @@
...
+    let drag_widget = new Gtk.ListBox();
+    drag_widget.set_size_request(allocation.width, allocation.height);
+    drag_widget.add_css_class("boxed-list");
+
+    let drag_row = new Adw.ActionRow();
+    drag_row.title = row.title;
+    drag_row.add_css_class("boxed-list");
```
andyholmes, line 47:
```suggestion
```

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,60 @@
...
+
+  drag.connect("drag-begin", (drag) => {
+    list.remove(row);
+    row.set_state_flags(Gtk.StateFlags.DROP_ACTIVE, true);
+    let drag_widget = new Gtk.ListBox();
+    drag_widget.set_size_request(allocation.width, allocation.height);
+    drag_widget.add_css_class("boxed-list");
+
+    let drag_row = new Adw.ActionRow();
+    drag_row.title = row.title;
```
andyholmes, line 46:
```suggestion
    let drag_row = new Adw.ActionRow({ title: row.title });
```

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,60 @@
...
+
+    drag_widget.append(drag_row);
+    drag_widget.drag_highlight_row(drag_row);
+
+    let icon = Gtk.DragIcon.get_for_drag(drag.get_drag());
+    icon.child = drag_widget;
+  });
+
+  // Drop controller
+  drop.connect("drop", (drop, value, _x, _y) => {});
```
andyholmes, line 57: Personally, I would wait until this is called to remove the row, then place it at `Gtk.ListBox.get_row_at_y().get_index()` +/- 1.

Note that `value` is unpacked for you, so the type of this object will be `Adw.ActionRow`, not `GObject.Value`.

src/Library/demos/Drag and Drop/main.js
sonnyp: We use snake case for variables. I'd suggest  `drag_source`
the main reasoning is that it makes it easy to differentiate quickly from a GI class

src/Library/demos/Drag and Drop/main.js
sonnyp, line 13: widgets are iteratables in GJS

```suggestion
// Iterate of ListBox children
for (const row of listbox) {
```

src/Library/demos/Drag and Drop/main.js
sonnyp: I've often used `self` for the first argument but no strong opinion

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,64 @@
...
+
+const list = workbench.builder.get_object("list");
+
+// Iterate over ListBox children
+for (const row of list) {
+  row.selectable = false;
+  let _drag_x;
+  let _drag_y;
+
+  const drag_source = new Gtk.DragSource();
```
andyholmes, line 14:
```suggestion
  const drag_source = new Gtk.DragSource({
    actions: Gtk.DragAction.MOVE,
  });
```

You need to set the drag action here. It's likely that the `Gtk.DropTarget` is ignoring actions from the source that don't match the target.

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,64 @@
...
+
+    const icon = Gtk.DragIcon.get_for_drag(drag);
+    icon.child = drag_widget;
+
+    drag.set_hotspot(_drag_x, _drag_y);
+  });
+
+  // Drop controller
+  drop_target.connect("drop", (drop, value, _x, _y) => {
+    const value_row = value.get_object();
```
andyholmes, line 58: `value` is already unpacked as a `GObject` for you, so `get_object()` will fail.

src/Library/demos/Drag and Drop/main.js
halfmexican: ah okay, thank you!

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,67 @@
...
+  row.selectable = false;
+  let _drag_x;
+  let _drag_y;
+
+  const drag_source = new Gtk.DragSource({
+    actions: Gdk.DragAction.MOVE,
+  });
+
+  const drop_target = Gtk.DropTarget.new(
+    row.constructor.$gtype,
```
andyholmes, line 19: This returns the correct `GType`, as confirmed by logging, but for some reason the drop never triggers unless it's `Gtk.ListBoxRow`. I'm not sure why that is, but if it works for you too, lets go for that.

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,67 @@
...
+    const icon = Gtk.DragIcon.get_for_drag(drag);
+    icon.child = drag_widget;
+
+    drag.set_hotspot(_drag_x, _drag_y);
+  });
+
+  // Drop controller
+  drop_target.connect("drop", (drop, value, _x, _y) => {
+    const value_row = value;
+    const target_index = list.get_row_at_y(_y).get_index() - 1;
```
andyholmes, line 62: This fires for me with the above change, however `Gtk.ListBox.get_row_at_y()` is not returning the row underneath the cursor. I'll poke at this a bit more.

halfmexican: okay thank you!

andyholmes: :facepalm: Now I remember how this works: the `drop` signal is being emitted *for* the widget you're dropping on.

In this case, I believe the best thing to do is add the `Gtk.DropTarget` controller only to the `Gtk.ListBox`. Then the `(x, y)` coordinates for the `drop` signal will be the bounds of the list box.

src/Library/demos/Drag and Drop/main.js
halfmexican: We can change it to Gtk.ListBoxRow. That is how I read I should get types, but it doesn't work for me either.

sonnyp (commented):
Design review:

* The drag icon should be dimmed
* There should be visual feedback of where the row is going to be dropped (swapped with)
* There should be a hover effect on each row

You can look at GNOME Settings -> Search for all 3

I suggest you implement these things and then @andyholmes can do the technical review.

sonnyp:
@halfmexican please request review from @andyholmes if this is ready

image

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  // Update row visuals during DnD operation
+  drop_controller.connect("enter", (_x, _y) => {
+    row.set_state_flags(Gtk.StateFlags.DROP_ACTIVE, false);
+  });
+
+  drop_controller.connect("leave", (_x, _y) => {
+    row.set_state_flags(Gtk.StateFlags.NORMAL, true);
+  });
```
andyholmes, line 68: You should just use `Gtk.ListBox.drag_highlight_row()` and `Gtk.ListBox.drag_unhighlight_row()` here, which are intended for this common case. It's guaranteed to set the proper state and avoids easy mistakes like:

```js
row.set_state_flags(Gtk.StateFlags.NORMAL, true);
```

Which should be:

```js
row.unset_state_flags(Gtk.StateFlags.DROP_ACTIVE);
```

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// Drop Handling
+drop_target.connect("drop", (drop, value, _x, _y) => {
+  const value_row = value;
```
andyholmes, line 73:
```suggestion
  const value_row = value;
```

There's no need for this.

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,88 @@
...
+  const target_index = list.get_row_at_y(_y).get_index();
+  const target_row = list.get_row_at_index(target_index);
+
+  // If value or the target row is null, do not accept the drop
+  if (!value || !target_row) {
+    return false;
+  }
+
+  list.remove(value_row);
+  list.insert(value_row, target_index);
```
andyholmes, line 83:
```suggestion
  list.remove(value);
  list.insert(value, target_index);
```

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,88 @@
...
+  drop_controller.connect("leave", (_x, _y) => {
+    row.set_state_flags(Gtk.StateFlags.NORMAL, true);
+  });
+}
+
+// Drop Handling
+drop_target.connect("drop", (drop, value, _x, _y) => {
+  const value_row = value;
+  const target_index = list.get_row_at_y(_y).get_index();
+  const target_row = list.get_row_at_index(target_index);
```
andyholmes, line 75:
```suggestion
  const target_row = list.get_row_at_y(_y);
  const target_index = target_row.get_index();
```

No need to go in circles for the same row.

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,88 @@
...
+
+  const drag_source = new Gtk.DragSource({
+    actions: Gdk.DragAction.MOVE,
+  });
+
+  row.add_controller(drag_source);
+  row.add_controller(drop_controller);
+
+  // Drag handling
+  drag_source.connect("prepare", (source, _x, _y) => {
```
andyholmes, line 26: Update the callbacks, so that arguments that are used aren't prefixed with `_`, like `x` and `y` here. Arguments that aren't used like `source` should be `_source`.

src/Library/demos/Drag and Drop/main.js
```diff
@@ -0,0 +1,88 @@
...
+    drag_widget.drag_highlight_row(drag_row);
+
+    const icon = Gtk.DragIcon.get_for_drag(drag);
+    icon.child = drag_widget;
+
+    drag.set_hotspot(_drag_x, _drag_y);
+  });
+
+  // Update row visuals during DnD operation
+  drop_controller.connect("enter", (_x, _y) => {
```
andyholmes, line 62: For callbacks where you use none of the arguments, you can omit them entirely or prefix them with `_`, but they should always be correct. These two callbacks should include the emitting object like `(_controller, _x, _x) => {}`, or be empty like `() => {}`.

andyholmes (changes requested):
Looks really good, a nice clean example of implementing drag-and-drop. Just a few cleanups and it LGTM.

halfmexican:
I think we're ready to merge

andyholmes (approved):
Looks perfect to me, thanks for sticking with this one, it will be a very helpful demo.

PR #255, opened 2023-03-28
Library: Add DropDown Entry

Kodecheff:
Partially fix https://github.com/sonnyp/Workbench/issues/268


andyholmes (changes requested):
This is a great start for demo, nice and simple but demonstrates a common API.

It would be good to add a bit of JavaScript or Vala demonstrating how the developer will respond to an item being selected by the used, and also how they might change the selection programmatically.

I don't think we can ship e.g. the Twitter icon in Workbench, but something like having an icon change when the selection changes would be an good example.

sonnyp:
Hey @Kodecheff  , do you want to finish this? If so, make sure to add your name to the list of contributors in about.js

sonnyp:
https://github.com/sonnyp/Workbench/pull/323

PR #260, opened 2023-04-09
library: Add Header Bar entry

AkshayWarrier:
Adds Header Bar entry.
Issue #69

### DIFF
--- a/src/Library/demos/Header Bar/main.blp
+++ b/src/Library/demos/Header Bar/main.blp
@@ -4,7 +4,7 @@ using Adw 1;
 Gtk.Window {
   default-width: 800;
   default-height: 600;
-  title: "Window";
+  title: _("Header Bar");
   titlebar: Adw.HeaderBar {
     centering-policy: strict;
     show-start-title-buttons: true;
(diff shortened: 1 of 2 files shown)


AkshayWarrier:
No worries at all :)

andyholmes (approved):
LGTM! We can merge if @sonnyp agrees

sonnyp:
This demonstrates how to use the objects/properties but with Workbench we're also trying to encourage developers on how to use them properly. 

Whatever people copy paste from Workbench should be correct and in accordance with the HIG / GNOME practices.

I think there are a couple of issues with this entry:

1. The main menu should use the `open-menu-symbolic` icon, have `primary: true` and "Main Menu" tooltip
2. some properties don't do anything 
3. we don't want to customize decoration layout, developers should use the default

However if we fix all of these, the entry becomes over-simplified. What might be interesting is to show how header bars can be used in different ways depending on content and type of window.

@bertob wdyt?

sonnyp:
Ok so I talked with Tobias. 
Our suggestion is to replicate this one https://developer.gnome.org/hig/patterns/containers/header-bars.html#button-style (with my comments).

And since other library entries will feature hearbars such as the existing Window and and future ToolbarView entries we will make sure that searching for "header bar" shows all the relevant entries.

AkshayWarrier:
I could not find some of those icons from that example in the library so I chose to keep the previous ones. I made the other changes you requested.

sonnyp (approved):

PR #273, opened 2023-05-17
flatpak: Copy vls and deps from the vala sdk

lw64:


lw64:
Not sure if its related, but vls cannot find gtk4 and workbench namespaces anymore.

re.sonny.Workbench.Devel.json
```diff
@@ -148,6 +99,11 @@
           "type": "dir",
           "path": "."
         }
+      ],
+      "post-install": [
+        "cp -a /usr/lib/sdk/vala/bin/* /app/bin/",
```
sonnyp, line 104: > cp: cannot stat '/usr/lib/sdk/vala/bin/*': No such file or directory

can you use `install` ?

lw64: This is strange. it seems like the vala sdk is not mounted into the extension point??

sonnyp: This is a flatpak-vscode only - I fixed it in this commit https://github.com/sonnyp/Workbench/pull/273/commits/e05ff73b08cdd15e38b05d5e66f14a48c7d6016f

sonnyp:
> Not sure if its related, but vls cannot find gtk4 and workbench namespaces anymore.

I can reproduce, definitely related. Could you look into it?

sonnyp (approved):
I also cannot reproduce the issue anymore.

PR #274, opened 2023-05-17
Support screenshot of external preview

lw64:


lw64:
image

image

src/Previewer/previewer.vala
```diff
@@ -6,20 +6,55 @@ namespace Workbench {
...
+          size = Graphene.Size () {
+            width = width,
+            height = height
+          }
+        };
+        Gdk.Texture texture = renderer.render_texture (node, rect);
+        string path = Path.build_filename (Environment.get_user_data_dir (), "Workbench screenshot.png");
+        texture.save_to_png (path);
+
+        Xdp.Parent parent = Xdp.parent_new_gtk (window);
```
sonnyp, line 43: Can you pass the primitive back to the main process instead, and handle this with the same code path as the internal previewer?

It will be easier to maintain.

Feel free to leave the changes to the manifest and meson.build, we will need them anyway when porting libportal examples to Vala.

lw64: These 3 lines are easier to maintain than another dbus interconnection I think. As long as there should not be a screenshot button in the main worbench window or the dialog modal to its the best option I think.

sonnyp (approved):
Thanks.

I have connected the dots with JS and simplified the vala code a bit 

https://github.com/sonnyp/Workbench/pull/274/commits

lw64:
thanks!

PR #275, opened 2023-05-21
Improve preview screenshot

sonnyp:
Follow up of https://github.com/sonnyp/Workbench/pull/274

- [x] Don't overwrite previous screenshot
- [x] Use the same date format as GNOME Shell
- [x] Save screenshots in a dedicated directory
- [x] Use open_directory instead of open_file
- [x] Check why screenshot of "View Switcher" does not work (JS) (only on modal window)
- [ ] https://github.com/sonnyp/Workbench/issues/298
- [ ] Make sure internal and external preview the same thing

### DIFF
--- a/src/Previewer/Previewer.js
+++ b/src/Previewer/Previewer.js
@@ -3,11 +3,6 @@ import GObject from "gi://GObject";
 import GLib from "gi://GLib";
 import Gio from "gi://Gio";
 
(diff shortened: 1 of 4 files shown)


sonnyp:
https://github.com/sonnyp/Workbench/commit/992044c5176cce68d70ac43a9e2da8fdda960d55

PR #299, opened 2023-06-01
library: Add WebView entry

### LINKED ISSUE #276: WebView library entry
Let's add a WebView library entry.

Here is a quick example, if possible let's use UI to create the `WebView` object instead

```js
import WebKit from "gi://WebKit?version=6.0";

const web_view = new WebKit.WebView();

web_view.load_uri("https://gnome.org/");
workbench.preview(web_view);
```

Unfortunately the documentation at https://webkitgtk.org/reference/webkit2gtk/stable/ isn't up to date (4.1 instead of 6.0) but on Fedora you can install the correct documentation with `sudo dnf install webkitgtk6.0-doc` and read it with [DevHelp](https://flathub.org/apps/org.gnome.Devhelp)

Let's look at important properties/methods too and add a few controls (without replicating a full fledge web browser).

Such as url bar, next/previous buttons and so on.

AkshayWarrier:
Add WebView Entry
Closes #276 

Also I would like to know, where should I include the status page for this demo - title, description, links etc. Since the webview takes up the entire window. I was thinking initially when the demo loads, we show the status page in the window and when a link is entered, the webview is shown. Wdyt?

### DIFF
--- a/src/Library/demos/Web View/main.blp
+++ b/src/Library/demos/Web View/main.blp
@@ -1,0 +1,40 @@
+using Gtk 4.0;
+using Adw 1;
+using WebKit 6.0;
+
+Window window{
+  default-width:1000;
+  default-height:800;
+
+  [titlebar]
+  Adw.HeaderBar {
+    title-widget:
+    Entry url_bar{
+      input-purpose: url;
+      hexpand:true;
+      margin-start: 54;
+      margin-end: 54;
+    };
+
+    [start]
+    Box {
+      spacing:6;
+      Button button_back{
+        icon-name: "arrow1-left-symbolic";
+        tooltip-text: _("Back");
+      }
+      Button button_forward{
+        icon-name: "arrow1-right-symbolic";
+        tooltip-text: _("Forward");
+      }
+      Button button_reload{
+        icon-name:"refresh-symbolic";
+        tooltip-text: _("Reload");
+      }
+      Button button_stop {
+        icon-name:"stop-sign-symbolic";
+        tooltip-text: _("Stop");
+      }
+    }
+  }
+}
--- a/src/Library/demos/Web View/main.blp
+++ b/src/Library/demos/Web View/main.blp
@@ -58,6 +58,7 @@ Adw.Clamp {
       margin-bottom: 18;
       vexpand: true;
       hexpand: true;
+      styles ["view"]
     }
 
     LinkButton {
--- a/src/Library/demos/Web View/main.blp
+++ b/src/Library/demos/Web View/main.blp
@@ -1,40 +1,69 @@
 using Gtk 4.0;
 using Adw 1;
-using WebKit 6.0;
 
-Window window{
-  default-width:1000;
-  default-height:800;
+Adw.Clamp {
+  maximum-size: 700;
 
-  [titlebar]
-  Adw.HeaderBar {
-    title-widget:
-    Entry url_bar{
-      input-purpose: url;
-      hexpand:true;
-      margin-start: 54;
-      margin-end: 54;
-    };
+  Box {
+    orientation: vertical;
 
-    [start]
     Box {
-      spacing:6;
-      Button button_back{
+      orientation: vertical;
+      margin-top: 12;
+      margin-bottom: 12;
+      spacing: 6;
+
+      Label {
+        label: "Web View";
+
+        styles ["title-1"]
+      }
+
+      Label {
+        label: "Load and display webpages and HTML";
+      }
+    }
+
+    Box controls {
+      spacing: 6;
+
+      Button button_back {
         icon-name: "arrow1-left-symbolic";
         tooltip-text: _("Back");
       }
-      Button button_forward{
+
+      Button button_forward {
         icon-name: "arrow1-right-symbolic";
         tooltip-text: _("Forward");
       }
-      Button button_reload{
-        icon-name:"refresh-symbolic";
+
+      Entry url_bar {
+        input-purpose: url;
+        hexpand: true;
+      }
+
+      Button button_reload {
+        icon-name: "refresh-symbolic";
         tooltip-text: _("Reload");
       }
+
       Button button_stop {
-        icon-name:"stop-sign-symbolic";
+        icon-name: "stop-sign-symbolic";
         tooltip-text: _("Stop");
       }
+    }
+
+    Adw.Bin container {
+      margin-top: 18;
+      margin-bottom: 18;
+      vexpand: true;
+      hexpand: true;
+    }
+
+    LinkButton {
+      margin-bottom: 12;
+      label: "API Reference";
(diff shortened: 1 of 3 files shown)


sonnyp:
I would prefer the WebView to be immediately available (and load https://gnome.org) so there is no extra step.
It's possible to show links below regardless. 

> Since the webview takes up the entire window.

It doesn't have to be this way.

src/Library/demos/Web View/main.blp
```diff
@@ -0,0 +1,69 @@
...
+        tooltip-text: _("Reload");
+      }
+
+      Button button_stop {
+        icon-name: "stop-sign-symbolic";
+        tooltip-text: _("Stop");
+      }
+    }
+
+    Adw.Bin container {
```
sonnyp, line 56: `Bin` is made for subclassing. You can use a simple `Box` here

src/Library/demos/Web View/main.js
```diff
@@ -0,0 +1,91 @@
...
+const { Uri } = imports.gi.GLib;
+
+const container = workbench.builder.get_object("container");
+const button_back = workbench.builder.get_object("button_back");
+const button_forward = workbench.builder.get_object("button_forward");
+const button_reload = workbench.builder.get_object("button_reload");
+const button_stop = workbench.builder.get_object("button_stop");
+const url_bar = workbench.builder.get_object("url_bar");
+const web_view = new WebKit.WebView({
+  zoom_level: 0.8,
```
sonnyp, line 12: any particular reason?

src/Library/demos/Web View/main.js
```diff
@@ -0,0 +1,91 @@
...
+
+// URL bar displays the current loaded page
+web_view.bind_property(
+  "uri",
+  url_bar.buffer,
+  "text",
+  GObject.BindingFlags.DEFAULT,
+);
+
+// URL bar displays the current loaded page
```
sonnyp, line 24: Doesn't look relevant here

src/Library/demos/Web View/main.js
```diff
@@ -0,0 +1,91 @@
...
+web_view.load_uri("https://www.gnome.org/");
+
+url_bar.connect("activate", () => {
+  let url = url_bar.buffer.text;
+  const scheme = Uri.peek_scheme(url);
+  if (scheme == null) {
+    url = `https://${url}`;
```
sonnyp, line 31:
```suggestion
    url = `http://${url}`;
```

default for the web is still `http` - the website will switch to https if it is supported

src/Library/demos/Web View/main.js
```diff
@@ -0,0 +1,91 @@
...
+  web_view.reload();
+});
+
+button_stop.connect("clicked", () => {
+  web_view.stop_loading();
+});
+
+web_view.connect("load-changed", (view, load_event) => {
+  switch (load_event) {
+    case WebKit.LoadEvent["STARTED"]:
```
sonnyp, line 54:
```suggestion
    case WebKit.LoadEvent.STARTED:
```

src/Library/demos/Web View/main.js
```diff
@@ -0,0 +1,91 @@
...
+      console.log("Page loading started");
+      break;
+    case WebKit.LoadEvent["FINISHED"]:
```
sonnyp, line 57:
```suggestion
    case WebKit.LoadEvent.FINISHED;
```

src/Library/demos/Web View/main.js
```diff
@@ -0,0 +1,91 @@
...
+      console.log("Page loading has finished ");
+      break;
+  }
+});
+
+web_view.connect("load-failed", (view, load_event, fail_url, error) => {
+  // Dont display error page if it is caused by stop_loading()
+  if (error.code !== 302) {
```
sonnyp, line 65: any idea i there is an enum we can use instead?

src/Library/demos/Web View/main.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "Web View",
+  "category": "network",
+  "description": "Load and display webpages and HTML",
+  "panels": ["ui", "preview"],
```
sonnyp, line 5: `code` is more relevant here

```suggestion
  "panels": ["code", "preview"],
```

src/Library/demos/Web View/main.blp
sonnyp: also please add `styles ["view"]` to make it white

sonnyp (commented):
Looks great.

Just a couple of things

src/Library/demos/Web View/main.blp
AkshayWarrier: For some reason the ``WebView`` only works with container widgets that can only have a single child, I tried using ``Box`` but it doesn't work. It also works with ``Clamp`` but I didn't want to clamp the view so I chose ``Bin``. Is there any other container widget I can use that takes only a single child?

src/Library/demos/Web View/main.js
AkshayWarrier: Oh it was because you couldn't really see much of the webpage because the view is small, so I decreased the zoom level. It also gave me a reason to show the property.

src/Library/demos/Web View/main.js
AkshayWarrier: The comment? I'll remove it

src/Library/demos/Web View/main.js
AkshayWarrier: Yeah I just found ``WebKit.NetworkError.CANCELLED``, I can use that instead.

sonnyp: yes please, it's always good to avoid "magic numbers"

https://en.wikipedia.org/wiki/Magic_number_(programming)

src/Library/demos/Web View/main.js
sonnyp: Interesting find. I can confirm it's the same with `Box` and `append`.

src/Library/demos/Web View/main.blp
sonnyp: That's an interesting find. I have no idea why.

I can confirm, even with `Box` and `append`

According to the inspector the WebView isn't even in the tree.

src/Library/demos/Web View/main.js
```diff
@@ -0,0 +1,90 @@
+import WebKit from "gi://WebKit?version=6.0";
+const GObject = imports.gi.GObject;
+const { Uri } = imports.gi.GLib;
```
sonnyp, line 3: I'll fix it, but please remember to be consistent

`import x from y` should be used

sonnyp (approved):
Well done.

I figured the reason why the webview cannot be created from Blueprint. It's a bug in Workbench, I will fix it and adjust the demo.

PR #301, opened 2023-06-04
library: add Adw.Clamp entry

### LINKED ISSUE #287: Adw Clamp
https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.Clamp.html

You could for example use a clamp to clamp a placehodler text generated from https://flathub.org/apps/org.gnome.design.Lorem

SoNiC-HeRE:
Closes #287 

Implemented Adw.Clamp demo following the description that was discussed. Added some more functionality to enhance UI. Looking for an early review.


src/Library/demos/Clamp/main.blp
```diff
@@ -0,0 +1,51 @@
...
+      Button decrease {
+        icon-name: "minus";
+        halign: center;
+      }
+
+    }
+
+    LinkButton {
+      label: "API Reference";
+      uri: "https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.Clamp.html";
```
sonnyp, line 48: Links should be to adw 1.3 docs

sonnyp (changes requested):
It doesn't build.
The formatting is off.
The strings should use `_()`

Please test and review your work before submitting.

SoNiC-HeRE:
Made the following Changes:
1) Fixed the build error
2) Replaced Gtk.TextView with Gtk.Label
3) Refined Code

andyholmes:
If this is one ready for re-review, please remember to hit that button again :slightly_smiling_face: We don't want to make you wait, or start re-reviewing before you're ready.

sonnyp (approved):
I wanted to do something more complicated but I think this is a good entry anyway and it's easy to go from it to the documentation.

PR #314, opened 2023-06-10
library: add css example

### LINKED ISSUE #58: Add example explaining how to use CSS
It's confusing in GTK.

The `#id` selector actually matches the Widget `name`
object `id` is for `get_object`

It can be difficult to figure out how to set css classes via code or builder
maybe list stuff that's not implemented

https://docs.gtk.org/gtk3/css-overview.html
https://docs.gtk.org/gtk4/css-overview.html
https://docs.gtk.org/gtk4/css-properties.html

halfmexican:
closes #58


halfmexican:
I feel like more needs to be done, so I'll leave this up to review to decide what that is.

sonnyp (approved):
made some small improvements https://github.com/sonnyp/Workbench/pull/314/commits/0ee82d21e344334a5ec9bb6c5e6abc8ecd0cb61c

Comment for later:

This would be a great candidate for some kind of tutorial format where we can go through every CSS feature.

Ideally, we would also demonstrate CSS animations.

PR #315, opened 2023-06-11
library: Adds TextView entry 

### LINKED ISSUE #277: TextView libray entry
# Description

> GTK has an extremely powerful framework for multiline text editing. The primary objects involved in the process are GtkTextBuffer, which represents the text being edited, and GtkTextView, a widget which can display a GtkTextBuffer. Each buffer can be displayed by any number of views.

We should have a demo demonstrating usage of TextView, in a ScrolledView with a lot of text.

With a few buttons

* Select all
* Copy
* Paste (at the cursor)

You can find some code examples here https://github.com/sonnyp/Commit/blob/a2dfeddd38ed96bbd0f4ce29fb6a1bc5b215ee44/src/editor.js

Workbench itself also makes use of these (SourceView extends TextView)

# Bookmarks

* https://docs.gtk.org/gtk4/section-text-widget.html
* https://docs.gtk.org/gtk4/class.TextView.html
* https://docs.gtk.org/gtk4/class.TextBuffer.html

SoNiC-HeRE:
Preview:

[Screencast from 2023-06-11 16-35-36.webm](https://github.com/sonnyp/Workbench/assets/96797205/4298f125-dd3f-40c2-9e63-f14fff1d3005)

Closes #277 

Apart from cut,copy,select all,clear all and paste functions;
Also added logic for the key bindings:
- ```ctrl + C```
- ```ctrl + V```
- ```ctrl + X```

Also supports direct interaction with ```mouse selection``` and ```backspace key```

### DIFF
--- a/src/Library/demos/TextView/main.blp
+++ b/src/Library/demos/TextView/main.blp
@@ -12,28 +12,28 @@ Adw.StatusPage {
     Frame {
       margin-bottom: 12;
 
-        ScrolledWindow {
-          height-request: 180;
-          width-request: 600;
+      ScrolledWindow {
+        height-request: 180;
+        width-request: 600;
 
-            TextView textview {
-              bottom-margin: 12;
-              left-margin: 12;
-              right-margin: 12;
-              top-margin: 12;
-              editable: true;
-              cursor-visible: true;
-              wrap-mode: char;
-             }
-           }
-         }
+        TextView textview {
+          bottom-margin: 12;
+          left-margin: 12;
+          right-margin: 12;
+          top-margin: 12;
+          editable: true;
+          cursor-visible: true;
+          wrap-mode: char;
+        }
+      }
+    }
 
     Box {
-    orientation: horizontal;
-    margin-top: 12;
-    margin-bottom: 12;
-    halign: center;
-    spacing: 18;
+      orientation: horizontal;
+      margin-top: 12;
+      margin-bottom: 12;
+      halign: center;
+      spacing: 18;
 
         Button copy {
           label: _("Copy");
@@ -49,10 +49,10 @@ Adw.StatusPage {
     }
 
     Box {
-    orientation: horizontal;
-    halign: center;
-    spacing: 18;
-    margin-bottom: 18;
+      orientation: horizontal;
+      halign: center;
+      spacing: 18;
+      margin-bottom: 18;
 
         Button select {
           halign: center;
--- a/src/Library/demos/TextView/main.blp
+++ b/src/Library/demos/TextView/main.blp
@@ -9,21 +9,24 @@ Adw.StatusPage {
     orientation: vertical;
     halign: center;
 
-    ScrolledWindow {
-    height-request: 180;
-    width-request: 600;
+    Frame {
+      margin-bottom: 12;
 
-        TextView textview {
-          bottom-margin: 12;
-          left-margin: 12;
-          right-margin: 12;
-          top-margin: 12;
-          margin-bottom: 12;
-          editable: true;
-          cursor-visible: true;
-          wrap-mode: char;
-          }
-        }
+        ScrolledWindow {
+          height-request: 180;
+          width-request: 600;
+
+            TextView textview {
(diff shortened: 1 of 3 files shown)


src/Library/demos/TextView/main.js
```diff
@@ -0,0 +1,63 @@
...
+    buffer.cut_clipboard(clipboard, true);
+    console.log("Text cut to clipboard");
+  } else {
+    console.log("No text selected to cut");
+  }
+}
+
+copy.connect("clicked", handleCopy);
+paste.connect("clicked", handlePaste);
+cut.connect("clicked", handleCut);
```
andyholmes, line 47: This is great with the buttons being tied to the buffer. I was thinking the better way is to use the (undocumented) GActions, like `clipboard.copy`, but then I think those would be less obvious being in the UI file.

What do you think? If you think it's fine like this, let's go ahead and we can revisit later if necessary.

src/Library/demos/TextView/main.blp
```diff
@@ -0,0 +1,74 @@
...
+          left-margin: 12;
+          right-margin: 12;
+          top-margin: 12;
+          margin-bottom: 12;
+          editable: true;
+          cursor-visible: true;
+          wrap-mode: char;
+          }
+        }
+
```
andyholmes, line 27: Let's put this `ScrolledWindow` in a `Gtk.Frame`, then move `TextView {margin-bottom: 12; }` to `Frame { margin-bottom: 12; }`.

This boxes up the TextView nicely, and it makes it visible that the text continues below the viewport.

andyholmes (changes requested):
Another great one, just some cleanup for formatting and frame around the scrolled window.

src/Library/demos/TextView/main.js
SoNiC-HeRE: Maybe we could comment the extra methods?

src/Library/demos/TextView/main.blp
```diff
@@ -0,0 +1,76 @@
...
+              bottom-margin: 12;
+              left-margin: 12;
+              right-margin: 12;
+              top-margin: 12;
+              editable: true;
+              cursor-visible: true;
+              wrap-mode: char;
+             }
+           }
+         }
```
andyholmes, line 29:
```suggestion
      ScrolledWindow {
        height-request: 180;
        width-request: 600;

        TextView textview {
          bottom-margin: 12;
          left-margin: 12;
          right-margin: 12;
          top-margin: 12;
          editable: true;
          cursor-visible: true;
          wrap-mode: char;
        }
      }
    }
```

src/Library/demos/TextView/main.blp
```diff
@@ -0,0 +1,76 @@
...
+
+        Button select {
+          halign: center;
+          label: _("Select all");
+        }
+
+        Button clear {
+          halign: center;
+          label: _("Clear all");
+        }
```
andyholmes, line 65:
```suggestion
      Button select {
        halign: center;
        label: _("Select all");
      }

      Button clear {
        halign: center;
        label: _("Clear all");
      }
```

src/Library/demos/TextView/main.js
andyholmes: Probably it would be best done with gi-docgen annotations (used to generate docs.gtk.org etc), because they can add extra things that aren't a part of GI. That's a longer fix, but I'm not sure GActions are guaranteed to be "API-stable", at least not until they're part of the documentation.

Maybe we can bring this up in #321, or a related demo.

andyholmes (changes requested):
Looks good once you apply the formatting suggestions.

We really need an auto-formatter, because formatting is important but not something developers should be doing by hand IMO.

SoNiC-HeRE:
> Looks good once you apply the formatting suggestions.
> 
> We really need an auto-formatter, because formatting is important but not something developers should be doing by hand IMO.

That'd be really helpful specially when we're dealing with large UI Files

andyholmes (approved):

PR #317, opened 2023-06-11
library: Add Frame Entry

### LINKED ISSUE #316: Frame Entry
https://docs.gtk.org/gtk4/class.Frame.html

Despite having less properties , can be a good demo

SoNiC-HeRE:
Currently implemented a simple UI to demonstrate usage of Gtk.Frame but can be implemented in a more interactive way that shows more complex use of Gtk.Frame such as:
- Collapsible Frame Sections:
Divide the frame's content into collapsible sections using Gtk.Expander widgets. Each section can be expanded or collapsed individually, allowing users to focus on specific parts of the frame's content.
- Frame with Toolbar:
Embed a Gtk.Toolbar within the frame to provide additional functionality or actions related to the frame's contents. The toolbar can contain buttons, icons, or dropdown menus to perform specific tasks.
- Or just for enhancing UI

Closes #316 

## Open to suggestions and looking for an early review


andyholmes (commented):
I like the idea of toolbars and revealers. It might be interesting to contrast this demo with the TextView demo where you put it in a scrollable frame.

For example, in the TextView demo everything will fit on "one page" and you can scroll in the area. But other situations you want the *page* to scroll, which is usually what you want for fixed content like labels and images.

One thing I'd try to avoid is duplicating the use-case of `Adw.ExpanderRow` *too* closely, because it might add some confusion about which one to use.

Side-note: `Gtk.Toolbar` doesn't exist in GTK4, because it was basically just a `Gtk.Box` with some special styling. So we just use `Gtk.Box` (or `Gtk.CenterBox`) with the `toolbar` CSS class.

SoNiC-HeRE:
> I like the idea of toolbars and revealers. It might be interesting to contrast this demo with the TextView demo where you put it in a scrollable frame.
> 
> For example, in the TextView demo everything will fit on "one page" and you can scroll in the area. But other situations you want the _page_ to scroll, which is usually what you want for fixed content like labels and images.
> 
> One thing I'd try to avoid is duplicating the use-case of `Adw.ExpanderRow` _too_ closely, because it might add some confusion about which one to use.
> 
> Side-note: `Gtk.Toolbar` doesn't exist in GTK4, because it was basically just a `Gtk.Box` with some special styling. So we just use `Gtk.Box` (or `Gtk.CenterBox`) with the `toolbar` CSS class.

How about demonstrating different usecases in a scrollable view?
1) Layout only
2) Toolbar
3) TextView

wdyt?

andyholmes:
Yeah, that sounds good. Let's give that a try and see how it looks.

SoNiC-HeRE:
Made the following changes:
- Divided the examples in 3 sections with different usecases
- Implemented a clean UI
- Added necessary JS Code

src/Library/demos/Frame/main.blp
```diff
@@ -0,0 +1,184 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "Frame";
```
sonnyp, line 5:
```suggestion
  title: _("Frame");
```

sonnyp (changes requested):
Looks great.

Only issue is we need to use an image in public domain.

What about this one https://www.publicdomainpictures.net/en/view-image.php?image=508689&picture=vintage-illustration-cat-kitten ?

SoNiC-HeRE:
> Looks great.
> 
> Only issue is we need to use an image in public domain.
> 
> What about this one https://www.publicdomainpictures.net/en/view-image.php?image=508689&picture=vintage-illustration-cat-kitten ?

works

src/Library/demos/Frame/main.blp
```diff
@@ -14,20 +14,144 @@ Adw.StatusPage {
...
+      halign: center;
+      spacing: 18;
+      margin-bottom: 18;
+      homogeneous: false;
+
+      ScrolledWindow {
+        height-request: 240;
+        width-request: 240;
+
+      TextView textview_with_frame {
```
andyholmes, line 125: It looks like these two ID's got mixed up, since this is the TextView without a frame.

I think you're right about the scrolled window here though, since it has more to do with `Gtk.ScrolledWindow` than `Gtk.TextView`. Just bump in the indentation here so the formatting is correct.

src/Library/demos/Frame/main.blp
```diff
@@ -14,20 +14,144 @@ Adw.StatusPage {
...
+
+    Label {
+      halign: start;
+      label: _("With Frame");
+      margin-bottom: 12;
+      styles ["title-4"]
+    }
+
+
+    Frame {
```
andyholmes, line 90: This one would be good to demonstrate the `label` property or `label-widget` property. The simplest thing would be to just set `Gtk.Frame:label` to something like `_("Label that handles Accessible setup")`.

It's not commonly used, but has the benefit of setting up accessible roles and relationships automatically.

src/Library/demos/Frame/main.blp
```diff
@@ -14,20 +14,144 @@ Adw.StatusPage {
...
+          width-request: 240;
 
-        Label {
-          wrap: true;
+        TextView textview_without_frame {
+          bottom-margin: 12;
+          left-margin: 12;
+          right-margin: 12;
+          top-margin: 12;
+          margin-bottom: 12;
```
andyholmes, line 151:
```suggestion
```

This can go, since you don't want a gap between the TextView and the Frame. You have the `Gtk.TextView:bottom-margin` set, and that's the correct one to use here.

andyholmes (changes requested):
Looks good, just a few tweaks and it's good to go!

sonnyp:
@SoNiC-HeRE please update the image https://github.com/sonnyp/Workbench/pull/317#pullrequestreview-1484894518

andyholmes (approved):
Nice one!

sonnyp (changes requested):
One last thing

the label needs to be center with the widget on top in the first and last examples

image

SoNiC-HeRE:
Fixed Spacing ,removed arbitrary margins and use box properties

src/Library/demos/Frame/main.blp
```diff
@@ -0,0 +1,188 @@
...
+          height-request: 180;
+        }
+      }
+    }
+
+    Box {
+      homogeneous: true;
+      margin-bottom: 24;
+      halign: center;
+      spacing: 108;
```
sonnyp, line 42: This is also considered an arbitrary number

please use box layout for centering
ask if you need help

sonnyp (changes requested):
Labels are still not centered vertically with the widget they describe

image

sonnyp:
Please be more thorough

SoNiC-HeRE:
> Please be more thorough

Yes, i think i misunderstood something ; fixed the error
Updated with Box layout

sonnyp (approved):

PR #324, opened 2023-06-12
library: Update Progress Bar demo

### LINKED ISSUE #322: Port usage of setInterval to AdwAnimation in Progress Bar demo
Let's update the Progress Bar demo to have

1. Progress bar that use AdwAnimation for smooth filling
2. Progress bar with a pulse animation

consider Keeping a progreess bar with setInterval

AkshayWarrier:
Closes #322
I haven't made any changes to the UI, and I didn't add another progress bar that makes use of setInterval because to pulse the progress bar I'm making use of a setInterval anyway.

### DIFF
--- a/src/Library/demos/Progress Bar/main.js
+++ b/src/Library/demos/Progress Bar/main.js
@@ -37,6 +37,7 @@ function pulseProgress() {
     if (counter >= 1.0) {
       clearInterval(interval);
       counter = 0;
+      second_bar.fraction = 0;
       return;
     }
 
--- a/src/Library/demos/Progress Bar/main.js
+++ b/src/Library/demos/Progress Bar/main.js
@@ -1,48 +1,61 @@
+import Adw from "gi://Adw";
+
 const first_bar = workbench.builder.get_object("first");
 const second_bar = workbench.builder.get_object("second");
 const play = workbench.builder.get_object("play");
 const progress_tracker = workbench.builder.get_object("progress_tracker");
 
-function handleProgress() {
-  //Counters for respective progress bars and for progress tracker label
-  let counter = 0.3;
-  let counter_two = 0.25;
-  let timeLeft = 12;
+const target = Adw.PropertyAnimationTarget.new(first_bar, "fraction");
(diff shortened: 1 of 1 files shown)


src/Library/demos/Progress Bar/main.js
```diff
@@ -1,48 +1,61 @@
...
+      progress_tracker.label = "";
+      clearInterval(interval);
+      console.log("Operation complete!");
+      return;
+    }
+
+    progress_tracker.label = `${time} seconds remaining…`;
+    time -= 1;
+  }, 1000);
+}
```
andyholmes, line 61: I think this should also reset the pulsed progress bar, like the animation callback.

src/Library/demos/Progress Bar/main.js
```diff
@@ -1,48 +1,61 @@
...
+  let time = 10;
+  const interval = setInterval(() => {
+    if (time === 0) {
+      progress_tracker.label = "";
+      clearInterval(interval);
+      console.log("Operation complete!");
+      return;
+    }
+
+    progress_tracker.label = `${time} seconds remaining…`;
```
andyholmes, line 58: I think this is fine for now, but later we might want to revisit the translatable string:

```js
const Format = imports.format;
const Gettext = imports.gettext;

Format.vprintf(Gettext.ngettext('%d second remaining',
    '%d seconds remaining', time), [time]);
```

andyholmes (changes requested):
LGTM, just needs a matching reset for the pulsed bar when the operation completes. Good to see that real ellipses, too :+1:

src/Library/demos/Progress Bar/main.js
AkshayWarrier: Yeah we'll have to do this for all the demos I guess.

AkshayWarrier:
Also, the animation progress bar and the pulse progress bar are not *completely* synchronized because of the way setInterval works.

andyholmes (approved):
LGTM

PR #331, opened 2023-06-13
library: Add Revealer entry

AkshayWarrier:
Adds Revealer demo entry


src/Library/demos/Revealer/main.blp
```diff
@@ -0,0 +1,60 @@
...
+    Box {
+      spacing: 12;
+      orientation: vertical;
+
+      Adw.Clamp {
+        maximum-size: 400;
+
+        Revealer revealer {
+          transition-duration: 300;
+          transition-type: slide_up;
```
sonnyp, line 28:
```suggestion
          // https://docs.gtk.org/gtk4/enum.RevealerTransitionType.html
          transition-type: slide_up;
```

src/Library/demos/Revealer/main.blp
```diff
@@ -0,0 +1,60 @@
...
+
+          ListBox {
```
sonnyp, line 30: When clicking on the button, it looks buggy because the cursor is above the first item which gets a special style for hover

can you either disable that or use an other child type? 

Screenshot from 2023-06-17 21-27-36

src/Library/demos/Revealer/main.js
```diff
@@ -0,0 +1,14 @@
+const button = workbench.builder.get_object("button");
+const revealer = workbench.builder.get_object("revealer");
+
+button.connect("clicked", () => {
```
sonnyp, line 4: it's a toggle button, this isn't the right thing to listen to

check the toggle button library entry :)

src/Library/demos/Revealer/main.js
```diff
@@ -0,0 +1,14 @@
...
+  revealer.reveal_child = !revealer.reveal_child;
```
sonnyp, line 5: you can use the state of the toggle button instead

src/Library/demos/Revealer/main.js
```diff
@@ -0,0 +1,14 @@
...
+  switch (button.label) {
+    case "Show":
+      button.label = "Hide";
+      return;
+    case "Hide":
+      button.label = "Show";
+      return;
```
sonnyp, line 12: never use user facing strings for conditions, it's fragile

you can use the state of the toggle button

sonnyp (changes requested):
Good start.

wdyt of adding an example? A cross-fade between 2 overlays can look very cool. Feel free to play around and find something nice.

sonnyp (approved):
neat

PR #333, opened 2023-06-14
library: Add Wallpaper entry

### LINKED ISSUE #332: Add Wallpaper Demo
https://libportal.org/method.Portal.set_wallpaper.html

SoNiC-HeRE:
[Screencast from 2023-06-14 10-17-24.webm](https://github.com/sonnyp/Workbench/assets/96797205/6a531a1f-b755-4182-90d8-e1505137891b)

Closes #332


SoNiC-HeRE:
This seems to work fine however we also get an additional warning as unknown XdptargetValue

sonnyp (changes requested):
You're not using the `promisified` version of `set_wallpaper`.

Please refer to the `Screenshot` Library entry.

src/Library/demos/Wallpaper/main.blp
```diff
@@ -0,0 +1,25 @@
...
+
+  Box {
+    orientation: vertical;
+    halign: center;
+
+    Button button {
+      label: _("Select Wallpaper");
+      margin-bottom: 42;
+      halign: center;
+      styles ["suggested-action"]
```
sonnyp, line 17:
```suggestion
      styles ["suggested-action", "pill"]
```

src/Library/demos/Wallpaper/main.js
```diff
@@ -0,0 +1,41 @@
...
+const portal = new Xdp.Portal();
+const parent = XdpGtk.parent_new_gtk(workbench.window);
+const button = workbench.builder.get_object("button");
+
+Gio._promisify(Gtk.FileDialog.prototype, "open", "open_finish");
+Gio._promisify(Xdp.Portal.prototype, "set_wallpaper", "set_wallpaper_finish");
+
+const filter = new Gtk.FileFilter({
+  name: "File",
+});
```
sonnyp, line 16: did you check what this does?

src/Library/demos/Wallpaper/main.js
```diff
@@ -0,0 +1,41 @@
...
+
+const dialog = new Gtk.FileDialog({
```
sonnyp, line 18: we only want to let the user select an image :)
an other option is to ship an image with Workbench and use that one 

We already have one, check the Overlay entry. I think it might make the entry clearer that any image can be set, not just something selected by the user.

sonnyp (changes requested):

SoNiC-HeRE:
Made the following changes:
- Refined Code
- Removed async handler
- Remove Gtk.FileDialog
- Improved UI

SoNiC-HeRE:
Made the following changes:
- Refined Code
- Removed async handler
- Remove Gtk.FileDialog
- Improved UI

sonnyp (approved):
Looks great.

As discussed, we need a different wallpaper in public domain, but I'll take care of updating it beore merging.

sonnyp:
I updated the image and fixed a warning:

PR #337, opened 2023-06-14
library: Add Check Button demo

### LINKED ISSUE #336: Add CheckButton demo

AkshayWarrier:
Closes #336
Right now the inconsistent checkbox is not functioning properly, after it has completed one cycle of inconsistent -> on -> off -> inconsistent, the next time the check button turns on "Mark as Done: On" is logged twice because of ``checkbutton3.active = true`` under case 0, the button triggers ``::toggled`` again.

A review would be appreciated.

### DIFF
--- a/src/Library/demos/Check Button/main.blp
+++ b/src/Library/demos/Check Button/main.blp
@@ -1,0 +1,90 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: _("Check Button");
+  description: _("Allow users to control a binary operation or make a selection from a set of options");
+
+  Box{
+    orientation: vertical;
+    spacing:12;
+
+    Box {
+      halign: center;
+      spacing: 60;
+      Box {
+        orientation: vertical;
+        spacing: 12;
+        Label {
+          label: _("Checkboxes");
+          styles ["title-4"]
+        }
+        Box {
+          orientation: vertical;
+          halign: center;
+          spacing:12;
+
+          CheckButton check_button1{
+            label: _("Show Notifications");
+            active: true;
+          }
+
+          CheckButton check_button2{
+            label: _("Auto-save changes");
+          }
+
+          CheckButton check_button3{
+            label: _("Mark as Done");
+            inconsistent: true;
+          }
+        }
+
+        LinkButton {
+          margin-top:12;
+          label: "Human Interface Guidelines";
+          uri: "https://developer.gnome.org/hig/patterns/controls/checkboxes.html";
+        }
+      }
+
+      Box {
+        orientation: vertical;
+        spacing: 12;
+        Label {
+          label: _("Radio Buttons");
+          styles ["title-4"]
+        }
+        Box {
+          orientation: vertical;
+          halign: center;
+          spacing:12;
+
+          CheckButton radio_button1{
+            label: _("Force System Style");
+          }
+
+          CheckButton radio_button2{
+            label: _("Force Light Mode");
+            group: radio_button1;
+          }
+
+          CheckButton radio_button3{
+            label: _("Force Dark Mode");
+            group: radio_button1;
+          }
+        }
+
+        LinkButton {
+          margin-top:12;
+          label: "Human Interface Guidelines";
+          uri: "https://developer.gnome.org/hig/patterns/controls/radio-buttons.html";
+        }
+      }
+
+    }
+
+    LinkButton {
+    label: "API Reference";
+    uri: "https://docs.gtk.org/gtk4/class.CheckButton.html";
+    }
+  }
+}
--- a/src/Library/demos/Check Button/main.blp
+++ b/src/Library/demos/Check Button/main.blp
@@ -7,14 +7,16 @@ Adw.StatusPage {
 
   Box{
     orientation: vertical;
-    spacing:12;
+    spacing: 12;
 
     Box {
-      halign: center;
-      spacing: 60;
+      homogeneous: true;
+      margin-start: 102;
+      margin-end: 102;
       Box {
         orientation: vertical;
         spacing: 12;
+
         Label {
           label: _("Checkboxes");
           styles ["title-4"]
@@ -22,21 +24,18 @@ Adw.StatusPage {
         Box {
           orientation: vertical;
           halign: center;
-          spacing:12;
-          CheckButton checkbox1{
-            label: _("Additions");
-          }
+          spacing: 12;
 
-          CheckButton checkbox2{
-            label: "Tomato";
+          CheckButton checkbox1{
+            active: true;
+            label: _("Enable Notifications");
           }
-          CheckButton checkbox3{
-            label: "Pickles";
+          CheckButton checkbox2{
+            label: _("Auto-save changes");
           }
         }
 
         LinkButton {
-          margin-top:12;
           label: "Human Interface Guidelines";
           uri: "https://developer.gnome.org/hig/patterns/controls/checkboxes.html";
         }
@@ -45,6 +44,7 @@ Adw.StatusPage {
       Box {
         orientation: vertical;
         spacing: 12;
+
         Label {
           label: _("Radio Buttons");
           styles ["title-4"]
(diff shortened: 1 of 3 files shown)


andyholmes:
So I think what you'll have to do is store a bit of state in an object/dictionary. Usually you'd have GSettings for this stuff somewhere, or state in your application.

These aren't used very often anymore, but they usually happen in a situation like this:

* [ ] Inconsistent (not possible in markdown, I think)
    * [X] Active
    * [ ] Inactive

It's not really important how you store the state for the demo, but something like this could work:

```js
const state = {
    'checkbox': 0.5,
    'checkbox_child1': 1.0,
    'checkbox_child2': 0.0,
};
```

> Three times: return the box to its mixed state, restoring each selected object’s original value for that setting.

This is the only tricky part about this, because you have to return the checkboxes to their original state. So my suggestion would be use the *actual* checkbox `active` states in your `switch` statement, only using `checkbox_child1` and `checkbox_child2` when restoring the original values.

In other words, if the parent is toggled you should only update `state.checkbox` and leave `state.checkbox_child1` and `state.checkbox_child1`, while updating the child widgets. However, if the children are toggled, you should then update everything.

Does that makes sense?

sonnyp:
@andyholmes could you help @AkshayWarrier with this one?

andyholmes:
Here's what I came up with, but check it out because I may have missed a corner case:

```js
import GObject from "gi://GObject";

const checkbox_all = workbench.builder.get_object("checkbox_all");
const checkbox2 = workbench.builder.get_object("checkbox2");
const checkbox3 = workbench.builder.get_object("checkbox3");

// Define some mock settings and perform setup
const settings = {
  checkbox_all: false,
  checkbox2: false,
  checkbox3: true,
};

checkbox_all.active = settings.checkbox_all;
checkbox_all.inconsistent = settings.checkbox2 !== settings.checkbox3;
checkbox2.active = settings.checkbox2;
checkbox3.active = settings.checkbox3;

/* GNOME HIG for inconsistent checkbox states
 *
 * Behaviour pattern starts from an inconsistent state
 */
function toggleAllCallback(checkbutton) {
  console.log(`toggleAllCallback(${checkbutton.label})`);

  GObject.signal_handlers_block_by_func(checkbox2, toggleOneCallback);
  GObject.signal_handlers_block_by_func(checkbox3, toggleOneCallback);

  // Once: check the box and apply the setting to all the
  // selected objects.
  if (!settings.checkbox_all && checkbox_all.inconsistent) {
    checkbox_all.active = true;
    checkbox_all.inconsistent = false;

    checkbox2.active = true;
    checkbox3.active = true;

    // Twice: uncheck the box and unset the setting to all
    // the selected objects.
  } else if (settings.checkbox_all && !checkbox_all.inconsistent) {
    checkbox_all.active = false;
    checkbox_all.inconsistent = false;

    checkbox2.active = false;
    checkbox3.active = false;

    // Three times: return the box to its mixed state, restoring
    // each selected object’s original value for that setting.
  } else {
    checkbox_all.active = settings.checkbox2 && settings.checkbox2;
    checkbox_all.inconsistent = settings.checkbox2 !== settings.checkbox3;

    checkbox2.active = settings.checkbox2;
    checkbox3.active = settings.checkbox3;
  }

  // Update the mock setting
  settings.checkbox_all = checkbox_all.active;

  GObject.signal_handlers_unblock_by_func(checkbox2, toggleOneCallback);
  GObject.signal_handlers_unblock_by_func(checkbox3, toggleOneCallback);
}

function toggleOneCallback(checkbutton) {
  console.log(`toggleOneCallback(${checkbutton.label})`);

  GObject.signal_handlers_block_by_func(checkbox_all, toggleAllCallback);

  if (checkbox2.active && checkbox3.active) {
    settings.checkbox_all = true;
    checkbox_all.active = true;
    checkbox_all.inconsistent = false;
  } else if (checkbox2.active || checkbox3.active) {
    settings.checkbox_all = false;
    checkbox_all.active = false;
    checkbox_all.inconsistent = true;
  } else {
    settings.checkbox_all = false;
    checkbox_all.active = false;
    checkbox_all.inconsistent = false;
  }

  // Update the mock settings
  settings.checkbox2 = checkbox2.active;
  settings.checkbox3 = checkbox3.active;

  GObject.signal_handlers_unblock_by_func(checkbox_all, toggleAllCallback);
}

checkbox_all.connect("toggled", toggleAllCallback);
checkbox2.connect("toggled", toggleOneCallback);
checkbox3.connect("toggled", toggleOneCallback);
```

src/Library/demos/Check Button/main.js
```diff
@@ -0,0 +1,91 @@
...
+      checkbox1.active = state.checkbox3;
+    } else {
+      state.checkbox3 = checkbox3.active;
+    }
+  } else if (checkbox2.active !== checkbox3.active) {
+    checkbox1.inconsistent = true;
+    state.checkbox1 = 0.5;
+    state.checkbox3 = checkbox3.active;
+  }
+});*/
```
andyholmes, line 75: Okay, so let's just rid of all this stuff for now, and replace it with signals or something like with radio buttons. Maybe a logged message that lists all selected boxes.

We can resume the "inconsistent" use case, after we hear back from the design team.

src/Library/demos/Check Button/main.blp
```diff
@@ -0,0 +1,86 @@
...
+        Box {
+          orientation: vertical;
+          halign: center;
+          spacing:12;
+          CheckButton checkbox1{
+            label: _("Additions");
+          }
+
+          CheckButton checkbox2{
+            label: "Tomato";
```
andyholmes, line 31: Make sure all these get marked for translation.

src/Library/demos/Check Button/main.blp
```diff
@@ -0,0 +1,86 @@
...
+        orientation: vertical;
+        spacing: 12;
+        Label {
+          label: _("Checkboxes");
+          styles ["title-4"]
+        }
+        Box {
+          orientation: vertical;
+          halign: center;
+          spacing:12;
```
andyholmes, line 25: Make sure all the indentation is good.

src/Library/demos/Check Button/main.js
```diff
@@ -0,0 +1,91 @@
...
+  if (radiobutton1.active) console.log("Force System Style");
+});
+
+radiobutton2.connect("toggled", () => {
+  if (radiobutton2.active) console.log("Force Light Mode");
+});
+
+radiobutton3.connect("toggled", () => {
+  if (radiobutton3.active) console.log("Force Dark Mode");
+});
```
andyholmes, line 91: This is good, let's add one `console.log()` message below the signal connector's, so it's printed when the demo is started.

andyholmes (changes requested):
Looks almost good to go, now that we're waiting on the "inconsistent" state. Let's get this one cleaned up for the standard use and merge what we have.

src/Library/demos/Check Button/main.js
AkshayWarrier: Should I just remove the inconsistent button for now? Just have two checkboxes, one set to active and the other one inactive, or should I still leave it in?

src/Library/demos/Check Button/main.blp
AkshayWarrier: Sure. And also if we are not doing the inconsistent button, I'll choose more fitting labels?

andyholmes: That sounds good, maybe something that contrasts with the radio buttons being "one at a time".

src/Library/demos/Check Button/main.js
AkshayWarrier: What should I be logging, the states of all the buttons? I'll have to add a bunch more if statements below then

AkshayWarrier:
> Maybe a logged message that lists all selected boxes.

I decided to individually log the states of the checkboxes to show their independence from each other, also it makes the demo more copiable :)

src/Library/demos/Check Button/main.blp
```diff
@@ -7,36 +7,35 @@ Adw.StatusPage {
...
           orientation: vertical;
           halign: center;
-          spacing:12;
+          spacing: 12;
+
           CheckButton checkbox1{
-            label: _("Additions");
+            active: true;
+            label: _("Enable Notifications");
           }
```
andyholmes, line 32:
```suggestion
          }

```

src/Library/demos/Check Button/main.blp
```diff
@@ -45,37 +44,32 @@ Adw.StatusPage {
...
-          spacing:12;
+          spacing: 12;
 
           CheckButton radio_button1{
-            label: _("Force System Style");
+            active: true;
+            label: _("Force Light Mode");
           }
 
           CheckButton radio_button2{
```
andyholmes, line 62:
```suggestion
          CheckButton radio_button2 {
```

andyholmes (changes requested):
Looks pretty much good to go.

A few formatting cleanups, also the last `LinkButton` I couldn't add a suggestion for, but the properties need to be indented.

After that looks good.

andyholmes (approved):
Well done, very clean and understandable. :+1:

PR #339, opened 2023-06-15
library: Do not use async handlers

sonnyp:
Signal handlers have semantic around return values.
Let's not encourage developers to use async signal handlers, as it can lead to hard to debug issues.

Also, some other improvements

### DIFF
--- a/src/Library/demos/Box/main.js
+++ b/src/Library/demos/Box/main.js
@@ -6,34 +6,9 @@ const button_prepend = workbench.builder.get_object("button_prepend");
 const button_remove = workbench.builder.get_object("button_remove");
 let count = 0;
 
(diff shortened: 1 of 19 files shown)


src/Library/demos/Box/main.js
```diff
@@ -103,3 +78,34 @@ valign_toggle_center.connect("toggled", () => {
...
+function remove() {
+  if (count) {
+    interactive_box.remove(interactive_box.get_last_child());
+    count--;
+  } else {
+    console.log("The box has no child widget to remove");
+  }
+}
+
+append();
```
sonnyp, line 111: so that we don't start with an empty state

andyholmes (approved):
Nice one, lots of good cleanups in here.

PR #340, opened 2023-06-15
library: add Gtk.Video Demo

### LINKED ISSUE #319: Add GtkVideo and GtkMediaControls demo
Add an example showing how simple videos or music playback can be handled in GTK, with controls.

https://docs.gtk.org/gtk4/class.Video.html
https://docs.gtk.org/gtk4/class.MediaControls.html

halfmexican:
closes #319


halfmexican:
This PR currently just autoplays a file selected by the user.

Not entirely sure how I would play Big Buck Bunny. I think I would have to use gstreamer

src/Library/demos/Video/main.blp
```diff
@@ -0,0 +1,30 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage page {
+  title: "Video";
+  description: _("A Widget that displays video with media controls");
```
sonnyp, line 6:
```suggestion
  description: _("Display video with media controls");
```

src/Library/demos/Video/main.blp
```diff
@@ -0,0 +1,30 @@
...
+
+  Box {
+    orientation: vertical;
+    halign: center;
+
+    Video video {
+      width-request: 640;
+      height-request: 360;
```
sonnyp, line 14: we should set this too the video we exact resolution of the video we will load

src/Library/demos/Video/main.blp
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    Button file_button {
+      halign: center;
+      label: _("Load Video File");
+      margin-top: 12;
+    }
```
sonnyp, line 21: let's ship a video and load it directly instead

you can use this one for now, I have asked Jakub if we can use it

https://github.com/sonnyp/Workbench/assets/19673/2f35ff89-64b7-4939-9ceb-ee93b29159ad

src/Library/demos/Video/main.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "Video",
+  "category": "user_interface",
+  "description": "A Widget that displays video with media controls",
```
sonnyp, line 4:
```suggestion
  "description": "Display video with media controls",
```

sonnyp (changes requested):
We should demonstrate at least one of the most common method in code.

A good way to do this is to enable play/pause when clicking on the video itself (not the media)

You can do this by retrieving the media stream with `video.media_stream` and using the MediaStream API 

https://docs.gtk.org/gtk4/class.MediaStream.html

andyholmes:
> This PR currently just autoplays a file selected by the user.
> 
> Not entirely sure how I would play Big Buck Bunny. I think I would have to use gstreamer

You should be able to pass any `GFile`, so `Gio.File.new_for_uri("https://download.blender.org/peach/trailer/trailer_400p.ogg")` should work.

sonnyp:
@andyholmes true but let's make sure demos work offline if possible 

We can ship a small video, it's not a problem https://github.com/sonnyp/Workbench/pull/340#discussion_r1233123180

src/Library/demos/Video/main.blp
sonnyp: Check the overlay demo to see how to load a file

sonnyp: Jakub agreed to us using this in the library under CC0 \o/

Can you add them to `artists` credits in `about.js` ? `Jakub Steiner https://jimmac.eu`

sonnyp: EDIT: nevermind I added Jakub in https://github.com/sonnyp/Workbench/pull/333/files

src/Library/demos/Video/main.blp
halfmexican: On my 1080p Monitor the exact resolution renders the demo basically unusable, I currently have it set to half the resolution.

sonnyp (approved):

PR #345, opened 2023-06-17
library: add Stack entry

halfmexican:


halfmexican:
not ready

halfmexican:
now we're good

src/Library/demos/Stacks/main.js
```diff
@@ -0,0 +1,54 @@
...
+    navigation_widget = new Gtk.StackSwitcher({ stack: stack });
+    root_box.prepend(navigation_widget);
+    root_box.orientation = Gtk.Orientation.VERTICAL;
+  } else {
+    root_box.remove(navigation_widget);
+    navigation_widget = new Gtk.StackSidebar({ stack: stack });
+    root_box.prepend(navigation_widget);
+    root_box.orientation = Gtk.Orientation.HORIZONTAL;
+  }
+});
```
andyholmes, line 36: I like this idea; switching the orientation is a nice little demo for the situation, too.

They layout looks a little off with the sidebar though, so I think there should be a separator or some kind of section-ing of the switcher/sidebar from the stack

src/Library/demos/Stacks/main.js
```diff
@@ -0,0 +1,54 @@
...
+
+transition_row.connect("notify::selected-item", () => {
+  stack.transition_type = transition_row.get_selected();
+});
+
+stack.interpolate = interpolate_switch.active;
+
+interpolate_switch.connect("notify::active", () => {
+  stack.interpolate = interpolate_switch.active;
+});
```
andyholmes, line 48: This one could probably just be bound as a property.

src/Library/demos/Stacks/main.js
```diff
@@ -0,0 +1,54 @@
...
+
+stack.transition_duration = transition_spin_button.value;
+
+transition_spin_button.connect("value-changed", () => {
+  stack.transition_duration = transition_spin_button.value;
+});
```
andyholmes, line 54: I'd like to say this could be bound also, but I guess one is a double and the other an integer. Probably okay.

src/Library/demos/Stacks/main.blp
```diff
@@ -0,0 +1,108 @@
...
+        }
+      };
+    }
+
+    StackPage page2 {
+      title: _("Page 2");
+      child: Adw.StatusPage {
+        Adw.PreferencesGroup {
+          styles ["boxed-list"]
+          halign: center;
```
andyholmes, line 50:
```suggestion
          halign: center;
```

I think this should be left to default (`fill`), or given a minimum width, because the rows get a little cramped depending on the dropdown selection.

src/Library/demos/Stacks/main.blp
```diff
@@ -0,0 +1,108 @@
...
+          }
+
+          Adw.ComboRow transition_row {
+            title: _("Transition Type");
+            model: StringList {
+              strings [
+                "None", "Cross-fade", "Slide Right", "Slide Left", "Slide Up",
+                "Slide Down", "Slide Left-Right", "Slide Up-Down", "Over Up", "Over Down",
+                "Over Left", "Over Right", "Under Up", "Under Down", "Under Left", "Under Right",
+                "Over Up-Down", "Over Down-Up" , "Over Left-Right", "Over Right-Left", "Rotate Left", "Rotate Right", "Rotate Left-Right"
```
andyholmes, line 66: Let's organize these into rough groupings, one grouping per-line (e.g. `Slide-*`, `Rotate-*`). I think that will make the source a bit easier to glance through.

andyholmes (changes requested):
Good start, sorry for the wait!

src/Library/demos/Stacks/main.blp
```diff
@@ -98,7 +98,7 @@ Box root_box {
 
     StackPage {
       title: _("Page 3");
-      child:
+      child:Over Down-Up
```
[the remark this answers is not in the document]
halfmexican: oops

andyholmes (approved):
LGTM

PR #348, opened 2023-06-18
library: Adds Camera Entry

### LINKED ISSUE #328: Adds Camera Demo
https://libportal.org/method.Portal.access_camera.html

SoNiC-HeRE:
Closes #328

### DIFF
--- a/src/Library/demos/Camera/main.js
+++ b/src/Library/demos/Camera/main.js
@@ -52,30 +52,44 @@ async function on_clicked() {
   // Create elements
   const source = Gst.ElementFactory.make("pipewiresrc", "source");
   const queue = Gst.ElementFactory.make("queue", "queue"); // add a queue element
-  const video_convert = Gst.ElementFactory.make("videoconvert", "video_convert");
-
-  // Set properties
-  source.set_property("fd", pwRemote); // pwRemote is the pipewiresrc obtained from libportal
-  //ource.set_property("path", "/dev/video0");
-  // Create the sink
   const paintable_sink = Gst.ElementFactory.make(
     "gtk4paintablesink",
     "paintable_sink",
   );
 
-  // Add elements to the pipeline
-  pipeline.add(source);
-  pipeline.add(queue);
-  pipeline.add(video_convert);
-  pipeline.add(paintable_sink);
-
-  // Link the elements
-  source.link(queue);
-  queue.link(video_convert); // link queue to the video_convert
-  video_convert.link(paintable_sink);
+  let video_convert;
+  let glsinkbin;
 
   const paintable = new GObject.Value();
-  paintable_sink.get_property("paintable", paintable);
+  if (true) {
+    const gltestsrc = Gst.ElementFactory.make("gltestsrc", "gltestsrc");
+    glsinkbin = Gst.ElementFactory.make("glsinkbin", "glsinkbin");
+    glsinkbin.set_property("sink", paintable_sink);
+
+    pipeline.add(gltestsrc);
+    pipeline.add(queue);
+    pipeline.add(glsinkbin);
+    gltestsrc.link(queue);
+    queue.link(glsinkbin);
+
+    paintable_sink.get_property("paintable", paintable);
+  } else {
+    const videotestsrc = Gst.ElementFactory.make(
+      "videotestsrc",
+      "videotestsrc",
+    );
+    video_convert = Gst.ElementFactory.make("videoconvert", "video_convert");
+    pipeline.add(videotestsrc);
+    pipeline.add(queue);
+    pipeline.add(video_convert);
+    pipeline.add(paintable_sink);
+    videotestsrc.link(queue);
+    queue.link(video_convert);
+    video_convert.link(paintable_sink);
+
+    paintable_sink.get_property("paintable", paintable);
+  }
+
   video.paintable = paintable.get_object();
 
   // Start the pipeline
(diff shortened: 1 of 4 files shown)


src/Library/demos/Camera/main.js
```diff
@@ -0,0 +1,76 @@
...
+    const videoSink = Gst.ElementFactory.make("gtksink", "videoSink");
+
+    // Set properties
+    source.set_property("path", pwRemote); // pwRemote is the pipewiresrc obtained from libportal
+
+    // Create the bin
+    const bin = new Gst.Bin();
+    bin.add(source);
+    bin.add(videoConvert);
+    pipeline.add(bin);
```
philn, line 38: So you create a bin but don't link it to the sink and you're not setting up ghost pads, not gonna work :)
A bin is not needed, you can add source and videoconvert to the pipeline directly, it's a bin already.

src/Library/demos/Camera/main.js
```diff
@@ -0,0 +1,76 @@
...
+    // Create the pipeline
+    const pipeline = new Gst.Pipeline();
+
+    // Create elements
+    const source = Gst.ElementFactory.make("pipewiresrc", "source");
+    const videoConvert = Gst.ElementFactory.make(
+      "videoconvert",
+      "videoConvert",
+    );
+    const videoSink = Gst.ElementFactory.make("gtksink", "videoSink");
```
philn, line 29: gtksink is gtk3, i don't think mixing gtk3 and gtk4 in the same process is recommended.

src/Library/demos/Camera/main.js
```diff
@@ -0,0 +1,76 @@
...
+    const bus = pipeline.get_bus();
+    bus.add_signal_watch();
+    bus.connect("message", (bus, message) => {
+      if (message.type === Gst.MessageType.EOS) {
+        // End of stream, handle accordingly
+      }
+    });
+
+    // Set the video stream on the GtkVideo widget
+    video.set_media_stream(videoSink);
```
philn, line 55: i don't think this would work, you need a `GtkMediaStream` and videoSink is a `GstElement`. I'm not familiar with the gtkvideo API, not sure how this would work.

src/Library/demos/Camera/main.js
```diff
@@ -0,0 +1,79 @@
...
+    "gtk4paintablesink",
+    "paintable_sink",
+  );
+
+  video_convert.link(paintable_sink);
+
+  // Add elements to the pipeline
+  pipeline.add(source);
+  pipeline.add(video_convert);
+  pipeline.add(paintable_sink);
```
philn, line 70: The order is wrong here, you need to add elements to the pipeline first, then link. Also it seems the source is not linked to anything in this version?

philn: Ah also, it's generally good practice to add a `queue` before sinks.

src/Library/demos/Camera/main.js
```diff
@@ -0,0 +1,89 @@
...
+  // Create elements
+  const source = Gst.ElementFactory.make("pipewiresrc", "source");
+  const queue = Gst.ElementFactory.make("queue", "queue"); // add a queue element
+  const video_convert = Gst.ElementFactory.make(
+    "videoconvert",
+    "video_convert",
+  );
+
+  // Set properties
+  source.set_property("path", pwRemote); // pwRemote is the pipewiresrc obtained from libportal
```
philn, line 58: Are you sure about this? Shouldn't be the `fd` property?

halfmexican: OH! "fd" is file descriptor that makes perfect sense. thank you so much!!!

halfmexican: I'm so sorry and I apologize for my ignorance, but I've never worked with Gstreamer or pipewire before, I'm not sure what to even do about these errors
image

halfmexican: It appears using a videotestsrc instead of a pipwiresrc seems to work

philn: try to set the `path` and `fd` properties, as in https://gitlab.gnome.org/-/snippets/19#LC68

philn: Also, pro-tip, for improved accessibility, please don't share screenshots of text. It doesn't cost much to copy/paste text in a code block and everyone will see it.

halfmexican: once again thank you so much for your help, the code you linked to set the path to "node_id" which the method takes as a parameter. https://libportal.org/method.Portal.open_pipewire_remote_for_camera.html says I should use the file descriptor "to create a a pw_core object, by using pw_context_fd()" and then I'll be able to access the camera nodes and I assume I could get the ID from their but I'm not sure how I would go about creating a pw_core object in GJS

halfmexican: oh do i need to use wireplumber?

philn: wireplumber is a session manager, i don't think it's much relevant here, but i may be wrong.

not sure about libportal API either, sorry....

what if for testing purpose you replace the gtksink with glimagesink  and replace videoconvert with `identity`?

halfmexican: okay will do!

halfmexican: This opens a completely black window called, "OpenGL renderer"

PR #354, opened 2023-06-19
library: add Audio demo

### LINKED ISSUE #320: Add audio with GtkMediaControls demo
Similar to https://github.com/sonnyp/Workbench/issues/319 but just for audio

halfmexican:
closes #320


src/Library/demos/Audio/main.blp
```diff
@@ -0,0 +1,54 @@
...
+  Box {
+    orientation: vertical;
+    halign: fill;
+
+		MediaControls controls {}
+
+    Label {
+      margin-top: 12;
+      label: _("Audio Playback Buttons");
+    }
```
sonnyp, line 17: I think this is superfluous

There is no ambiguity

sonnyp (changes requested):
I don't think we need that many sounds

maybe just 2 buttons

"Play sound" and "Play music"? (The icons are cool though) 

I think the dog is the best sounding one. 
Where did you get the files from?

Also the audio control should be clamped 

I get a bunch of errors when playing the sounds:

(re.sonny.Workbench.Devel:4): Gtk-CRITICAL **: 17:55:51.145: gtk_media_stream_stream_ended: assertion '!gtk_media_stream_get_ended (self)' failed

Could you look into it?

PR #355, opened 2023-06-21
library: Add location entry

### LINKED ISSUE #350: libportal location
# Description

The location portal allows to get the user location and even monitor their movement.

https://libportal.org/method.Portal.location_monitor_start.html

We should have a demo that display the information from the `location-updated` signal.

I have created a follow up for a map demo which is out of scope for this:

https://github.com/sonnyp/Workbench/issues/349

SoNiC-HeRE:
Closes #350 

Implemented the following changes:
- Made the UI interactive by using Revealer
- Used Gtk.SpinButton and Gtk.Dropdown for changing values 
- Covered all 8 properties 

``` Looking for an early review ```

### DIFF
--- a/src/Library/demos/Location/main.blp
+++ b/src/Library/demos/Location/main.blp
@@ -29,11 +29,11 @@ Adw.StatusPage {
             value: 0;
             climb-rate: 1;
             adjustment: Adjustment {
-                          lower: 0;
-                          upper: 100;
-                          step-increment: 1;
-                          value: 0;
-                        };
+              lower: 0;
+              upper: 100;
+              step-increment: 1;
+              value: 0;
+            };
           }
         }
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/Library/demos/Location/main.blp [hunk 1]
```diff
@@ -0,0 +1,181 @@
...
+            valign: center;
+            wrap: true;
+            value: 0;
+            climb-rate: 1;
+            adjustment: Adjustment {
+                          lower: 0;
+                          upper: 100;
+                          step-increment: 1;
+                          value: 0;
+                        };
```
sonnyp, line 36: only 2 spaces indent please, example:

```suggestion
            adjustment: Adjustment {
              lower: 0;
              ..
            };
```

src/Library/demos/Location/main.blp
```diff
@@ -0,0 +1,181 @@
...
+          }
+        }
+
+        Adw.ActionRow {
```
sonnyp, line 58: please use a Adw ComboRow instead of an `ActionRow` + `DropDown`

sonnyp (commented):
Looks good!

image

Could you make the stop button `sensitive: false` when the session is started and vice versa? Like ASHPD.

Something I noticed is that pressing "Run" after starting a session breaks the demo, could you look into that?

SoNiC-HeRE:
> Looks good!
> 
> image
> 
> Could you make the stop button `sensitive: false` when the session is started and vice versa? Like ASHPD.
> 
> Something I noticed is that pressing "Run" after starting a session breaks the demo, could you look into that?

I've applied all the mentioned changes and ``` Something I noticed is that pressing "Run" after starting a session breaks the demo, could you look into that? ``` doesn't happen with me; maybe can you try reproducing again?

sonnyp:
It happens when the session isn't closed before pressing "Run" again.

I'm not sure what we can do about it at the moment. Possibly we have to wait until we support opening JS demo in a separate process.

I'll investigate.

PR #357, opened 2023-06-21
library: Add libshumate entry

### LINKED ISSUE #349: Library entry for map / libshumate
https://gnome.pages.gitlab.gnome.org/libshumate/

GNOME Maps uses libshumate and is written in GJS so it should be straightforward.

Unfortunately it doesn't appear to be part of the Sdk. Maybe we can push for it?

AkshayWarrier:
Closes #349 
I have added some basic controls to the demo. But I think adding ``PathLayer`` will probably make the demo more complex considering how much code is already there. We could make follow-up to this if we find that ``PathLayer`` is an important class to demonstrate.


re.sonny.Workbench.Devel.json
```diff
@@ -71,6 +71,21 @@
...
+        "config-opts": [
+            "-Dgtk_doc=false",
+            "-Dlibsoup3=true"
+        ],
+        "sources" : [
+            {
+                "type" : "git",
+                "url" : "https://gitlab.gnome.org/GNOME/libshumate.git",
+                "branch" : "main"
+            }
```
sonnyp, line 86: We should use the latest stable version, please use the archive from download.gnome.org like we do for other GNOME modules.

sonnyp: Also, it should be added to `re.sonny.Workbench.json` as well

src/Library/demos/Map/main.blp
```diff
@@ -0,0 +1,63 @@
+using Gtk 4.0;
+using Adw 1;
+using Shumate 1.0;
+
+Adw.StatusPage {
+  title: _("Map");
+  description: _("Display an interactive map widget");
```
sonnyp, line 7:
```suggestion
  description: _("Display an interactive map");
```

src/Library/demos/Map/main.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "Map",
+  "category": "user_interface",
+  "description": "Display an interactive map widget",
```
sonnyp, line 4:
```suggestion
  "description": "Display an interactive map",
```

src/Library/demos/Map/main.js
```diff
@@ -0,0 +1,70 @@
...
+    else if (longitude > Shumate.MAX_LONGITUDE || longitude < Shumate.MIN_LONGITUDE) {
+      console.log(
+        `Longitudes must be between ${Shumate.MIN_LONGITUDE} and ${Shumate.MAX_LONGITUDE}`,
+      );
+    } else {
+      map_widget.map.go_to(latitude, longitude);
+    }
+  } else {
+    console.log("Please enter valid coordinates");
+  }
```
sonnyp, line 69: Too complex, use early returns to reduce complexity and indentation.



```
if (!(isNaN(latitude) || isNaN(longitude))) {
  console.log("Please enter valid coordinates");
  return;
}


...
```

sonnyp (changes requested):
First review/iteration

image


This is confusing. 

Let's separate the 2 things

1. Longitude / latitude "Go" (probably it should zoom too)
2. Place a marker

src/Library/demos/Map/main.js
AkshayWarrier: Okay I'll try simplifying this, input validation gets complex very quickly 😅

sonnyp (approved):
great job :+1:

PR #358, opened 2023-06-22
Add Documentation Viewer

### LINKED ISSUE #282: Expose SDK documentation
# Description

Workbench is made to work offline. 

Documentation is available in `org.gnome.Sdk.Doc` extension. We can install it by `cp` in a similar way we do with vala extension in https://github.com/sonnyp/Workbench/pull/273

We can make it available in Menu -> Platform Tools and a webkit webview

It should have search but its ok to make a follow up.

# Bookmarks

* https://floss.social/@sonny/110428662126555019

AkshayWarrier:
Closes #282


re.sonny.Workbench.Devel.json
```diff
@@ -3,7 +3,7 @@
   "runtime": "org.gnome.Sdk",
   "runtime-version": "44",
   "sdk": "org.gnome.Sdk",
-  "sdk-extensions": ["org.freedesktop.Sdk.Extension.vala"],
+  "sdk-extensions": ["org.freedesktop.Sdk.Extension.vala","org.gnome.Sdk.Docs"],
```
sonnyp, line 6: we will need this in `re.sonny.Workbench.json` as well

also, this doesn't get installed automatically so we need to copy the files into the Flatpak bundle

you can do so like this:

https://github.com/sonnyp/Workbench/pull/273/files#diff-cfaac9a05c81f5033c5692a6f05fa4c50383a5f371d10b61bb9caf74b66ec0e5R144-R147

And you can test it works correctly by using "Build" then "Export" and finally installing the `.flatpak`

src/DocumentationViewer.blp
```diff
@@ -0,0 +1,52 @@
+using Gtk 4.0;
+using Adw 1;
+
+Window documentation_viewer{
+  title: "Documentation";
+  default-width: 1000;
+  default-height: 600;
+  hide-on-close: true;
+  titlebar:
+  Adw.HeaderBar {
```
sonnyp, line 10:
```suggestion
  titlebar: Adw.HeaderBar {
```

sonnyp (commented):
Great start!

My suggestion for the time being is to build/iterate on the documention viewer as a demo. 
This way you can develop on it without restarting Workbench all the time :) 

I think the next step would be to list the available docs and let the user navigate to the `index.html` of their choice.

You can use https://gjs-docs.gnome.org/gjs/overrides.md#gio-fileenumerator-symbol-asynciterator

@andyholmes might also have input on this

re.sonny.Workbench.Devel.json
AkshayWarrier: What should I be copying for the docs sdk? I can't find anything else in ``/usr/lib/sdk/`` other than ``vala``.

sonnyp: you should do something like `cp -a /usr/share/docs/. /app/usr/share/docs/`

and then use the `/app/usr/share/docs` path

andyholmes:
We do have an example of recursively search through directories, although it's an example of deleting files so it will need to be adjusted. You can find that [here](https://gjs.guide/guides/gio/file-operations.html#recursively-deleting-a-directory).

The example doesn't use `Gio._promisify()`-style `async` functions, so it should be modified to do that. However, I would recommend not using `await` and instead storing the resulting `Promise`s as variables and using `await Promise.all(branches);` like in the example. That will allow the threaded functions to run in parallel, while resolving everything in proper order.

AkshayWarrier:
I need help with displaying the ListView in the sidebar, to show all the docs. I have created some helper functions to get all the docs but for some reason, the SignalListItemFactory's ``setup`` callback is not being called at all and none of the items are being displayed. I'm sure that the ListModel contains items in it. (I forgot to mention this here before, in case someone is going to take a look at the code)

sonnyp:
@AkshayWarrier is this ready for review?

AkshayWarrier:
Oh not yet I have to push my new changes first, I'll mark it ready for review after that.

AkshayWarrier:
For the go back and go forward buttons, I will have to add some extra logic to take care of the sidebar (enable/disable it appropriately). I haven't done that yet as of now, but maybe we should focus on a better solution for the sidebar first.

src/DocumentationViewer.blp
```diff
@@ -0,0 +1,69 @@
...
+        tooltip-text: "Toggle Sidepanel";
+      }
+      ToggleButton search_button {
+        icon-name: "loupe-large-symbolic";
+        tooltip-text: "Search";
+      }
+  };
+
+
+  Paned documentation_paned{
```
sonnyp, line 34:
```suggestion
  Paned documentation_paned {
```

src/DocumentationViewer.blp
```diff
@@ -0,0 +1,69 @@
...
+
+Window documentation_viewer{
+  title: "Documentation";
+  default-width: 1200;
+  default-height: 800;
+  hide-on-close: true;
+  titlebar:
+  Adw.HeaderBar {
+      Button back{
+        icon-name: "arrow1-left-symbolic";
```
sonnyp, line 13:
```suggestion
        icon-name: "go-previous-symbolic";
```

it's a good idea to check what other GNOME apps do, I got this from GNOME Web

src/DocumentationViewer.blp
```diff
@@ -0,0 +1,69 @@
...
+        tooltip-text: "Back";
+        styles ["flat"]
+      }
+      Button forward{
+        icon-name: "arrow1-right-symbolic";
```
sonnyp, line 18:
```suggestion
        icon-name: "go-next-symbolic";
```

src/DocumentationViewer.blp
```diff
@@ -0,0 +1,69 @@
...
+      }
+      ToggleButton sidebar_button{
+        active: false;
+        icon-name: "dock-left-symbolic";
+        tooltip-text: "Toggle Sidepanel";
+      }
+      ToggleButton search_button {
+        icon-name: "loupe-large-symbolic";
+        tooltip-text: "Search";
+      }
```
sonnyp, line 30: I don't think we need both and opening the sidebar should open the search. But let's iterate on the UI in a follow up.

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+  Gio.FileEnumerator.prototype,
+  "next_files_async",
+  "next_files_finish",
+);
+
+Gio._promisify(
+  Gio.File.prototype,
+  "load_contents_async",
+  "load_contents_finish",
+);
```
sonnyp, line 32: all `Gio._promisify` go to `init.js`

Please verify that they're not there already

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+  const container = builder.get_object("webview_container");
+  const webview = new WebKit.WebView();
+  const listbox = builder.get_object("listbox");
+
+  const sidebar_button = builder.get_object("sidebar_button");
+  const search_button = builder.get_object("search_button");
+  const search_entry = builder.get_object("search_entry");
+  const search_bar = builder.get_object("search_bar");
+  const back = builder.get_object("back");
+  const forward = builder.get_object("forward");
```
sonnyp, line 50:
```suggestion
  const back = builder.get_object("button_back");
  const forward = builder.get_object("button_forward");
```

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+  }) {
+  const builder = Gtk.Builder.new_from_resource(resource);
+  const window = builder.get_object("documentation_viewer");
+
+  const container = builder.get_object("webview_container");
+  const webview = new WebKit.WebView();
+  const listbox = builder.get_object("listbox");
+
+  const sidebar_button = builder.get_object("sidebar_button");
+  const search_button = builder.get_object("search_button");
```
sonnyp, line 46: type goes first

```suggestion
  const button_sidebar = builder.get_object("button_sidebar");
  const button_search = builder.get_object("button_search");
```

src/DocumentationViewer.js
sonnyp: Back button should be disabled when webview.can_go_back is false

Forward button should be disabled when webview.can_go_forward is false

Please use property bindings for that

src/window.js
```diff
@@ -78,6 +79,8 @@ export default function Window({ application }) {
 
   const placeholders = readDemo("Welcome");
 
+  const documentation_viewer = DocumentationViewer({window,application})
```
sonnyp, line 82:
```suggestion
  DocumentationViewer({window, application})
```
`documentation_viewer`  isn't used

consider moving this line next to `Libray()`

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+      } catch (e) {
+        // Ignore the error if the dir does not contain index.json
+        if (!e.matches(Gio.IOErrorEnum, Gio.IOErrorEnum.NOT_FOUND)) logError(e);
+      }
+    }
+
+    return docs;
+  } catch (e) {
+    logError(e);
+  }
```
sonnyp, line 166: since you have a error handler up the chain in https://github.com/sonnyp/Workbench/pull/358/files#diff-6533bf97682b99bf434791a454793569a69668f437288ddcedac958454abf4a7R122, you can remove the try/catch in this function

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+}
+
+async function readDocIndex(base_path, dir) {
+  const file = Gio.File.new_for_path(`${base_path}/${dir}/index.json`);
+  const [json, _] = await file.load_contents_async(null);
```
sonnyp, line 171:
```suggestion
  const [json] = await file.load_contents_async(null);
```

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+  return JSON.parse(new TextDecoder().decode(json));
```
sonnyp, line 172: we have a `decode` fn in `util.js`

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+      );
+      if (infos.length === 0) break;
+      for (const info of infos) {
+        files.push(info.get_name());
+      }
+    }
+    return files;
+  } catch (e) {
+    logError(e);
+  }
```
sonnyp, line 200: since you have a error handler up the chain in https://github.com/sonnyp/Workbench/pull/358/files#diff-6533bf97682b99bf434791a454793569a69668f437288ddcedac958454abf4a7R122, you can remove the try/catch in this function

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+);
+
+Gio._promisify(
+  Gio.File.prototype,
+  "load_contents_async",
+  "load_contents_finish",
+);
+
+export default function DocumentationViewer({
+  window:application_window,
```
sonnyp, line 35:
```suggestion
  window: application_window,
```

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+  const sidebar_button = builder.get_object("sidebar_button");
+  const search_button = builder.get_object("search_button");
+  const search_entry = builder.get_object("search_entry");
+  const search_bar = builder.get_object("search_bar");
+  const back = builder.get_object("back");
+  const forward = builder.get_object("forward");
+
+  const base_path = "/app/share/doc";
+
+  webview.load_uri(`file://${base_path}/gtk4/index.html`);
```
sonnyp, line 54: in general, never use string concat for building paths, or other well known formats
there are many pitfalls

Please use https://gjs-docs.gnome.org/glib20~2.0/glib.uri#function-build

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+  getDocs(base_path)
+    .then((docs) => {
+      for (const doc in docs) {
+        const row = new Adw.ActionRow({
+          title: doc,
+        });
+        row.add_suffix(new Gtk.Image({ icon_name: "go-next-symbolic" }));
+        listbox.append(row);
+      }
+      listbox.connect("row-selected", (_, row) => {
```
sonnyp, line 117: This is the convention in Workbench

```suggestion
      listbox.connect("row-selected", (self, row) => {
```

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+async function getDocs(base_path) {
+  try {
+    const docs = {};
+    const dirs = await list(base_path);
+
+    for (const dir of dirs) {
+      try {
+        const index = await readDocIndex(base_path, dir);
+        const namespace = `${index["meta"]["ns"]}-${index["meta"]["version"]}`;
+        const path = `file://${base_path}/${dir}/index.html`;
```
sonnyp, line 155: this isn't a path, it's a uri, please rename to uri and use `GLib.Uri.build`

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+        docs[namespace] = path;
+      } catch (e) {
+        // Ignore the error if the dir does not contain index.json
+        if (!e.matches(Gio.IOErrorEnum, Gio.IOErrorEnum.NOT_FOUND)) logError(e);
```
sonnyp, line 159: :+1:

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+    .then((docs) => {
+      for (const doc in docs) {
+        const row = new Adw.ActionRow({
+          title: doc,
+        });
+        row.add_suffix(new Gtk.Image({ icon_name: "go-next-symbolic" }));
+        listbox.append(row);
+      }
+      listbox.connect("row-selected", (_, row) => {
+        webview.load_uri(docs[row.title]);
```
sonnyp, line 118: let's build `docs` in a way that we can do this directly

```suggestion
        webview.load_uri(doc.uri);
```

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+  action_documentation.connect("activate", () => {
+    window.present();
+  });
+  application.add_action(action_documentation);
+  application.set_accels_for_action("app.documentation", ["<Control><Shift>D"]);
+}
+
+async function getDocs(base_path) {
+  try {
+    const docs = {};
```
sonnyp, line 148: please turn docs into an array of objects instead

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+    const dirs = await list(base_path);
+
+    for (const dir of dirs) {
+      try {
+        const index = await readDocIndex(base_path, dir);
+        const namespace = `${index["meta"]["ns"]}-${index["meta"]["version"]}`;
+        const path = `file://${base_path}/${dir}/index.html`;
+        docs[namespace] = path;
```
sonnyp, line 156: let's do something like this instead

```
docs.push({
  title: namespace,
  uri: ...
});
```

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+    const dir = Gio.File.new_for_path(dir_path);
+    const files = [];
+    const enumerator = await dir.enumerate_children_async(
+      "standard::name",
+      Gio.FileQueryInfoFlags.NOFOLLOW_SYMLINKS,
+      GLib.PRIORITY_DEFAULT,
+      null,
+    );
+    while (true) {
+      const infos = await enumerator.next_files_async(
```
sonnyp, line 187: we have a better syntax now using async iterator

https://gjs-docs.gnome.org/gjs/overrides.md#gio-fileenumerator-symbol-asynciterator

src/DocumentationViewer.js
```diff
@@ -0,0 +1,219 @@
...
+
+  const action_documentation = new Gio.SimpleAction({
+    name: "documentation",
+    parameter_type: null,
+  });
+  action_documentation.connect("activate", () => {
+    window.present();
+  });
+  application.add_action(action_documentation);
+  application.set_accels_for_action("app.documentation", ["<Control><Shift>D"]);
```
sonnyp, line 143: please remove the shortcut,

Ctrl+Shift+D is a well known short for opening gtk inspector

Hopefully we will soon better integrate the documentation viewer so a shortcut might not even be necessary

sonnyp (changes requested):
Good work, mostly smaller things to fix and ready to go!

We can then create follow up tickets to improve it.

src/DocumentationViewer.js
AkshayWarrier: I've made the change but I haven't used property bindings because ``can_go_back`` / ``can_go_forward`` are functions
https://github.com/sonnyp/Workbench/pull/358/commits/23b9f60498d0107d0c48a1aba3c14281302928c0#diff-6533bf97682b99bf434791a454793569a69668f437288ddcedac958454abf4a7R36-R37

src/DocumentationViewer.js
AkshayWarrier: Could you give an example of how I should build the uri, when I'm using ``GLib.Uri.build``, the actual path will still have to made using string concats right? Something like this
```js
const base_path = "/app/share/doc";
const path = `${base_path}/gtk4/index.html`;
const uri = GLib.Uri.build(
  GLib.UriFlags.NONE,
  "file",
  null,
  null,
  -1,
  path,
  null,
  null,
);
```
(Also the code above returns `file:/app/share/doc/gtk4/index.html` as supposed to `file:///app/share/doc/gtk4/index.html`, I'm not sure if I'm using it correctly)

sonnyp: hmm my bad, this is the way to go:

`new Gio.File('/app/share/doc').resolve_relative_path('gtk4/index.html').get_uri()`

src/DocumentationViewer.js
AkshayWarrier: I did not turn docs into an array of objects because I didn't want to iterate over it every time the user selects a doc, and check which doc was selected in the listbox, instead I could directly get the uri of the selected doc. Would the array make it inefficient or will it be fine?

sonnyp: We don't need to iterate over the docs everytime see https://github.com/sonnyp/Workbench/pull/358#discussion_r1245883792

src/DocumentationViewer.blp
```diff
@@ -1,41 +1,47 @@
...
   Paned documentation_paned {
     position: 400;
 
     [end]
-    Adw.Bin webview_container{}
+    WebKit.WebView webview {
+      settings: WebKit.Settings {
+        enable-back-forward-navigation-gestures: true;
+        enable-developer-extras: true;
+        enable-smooth-scrolling: true;
```
sonnyp, line 42: a few nice settings :)

FYI you can open the web inpector with right click "Inspect Element"

src/DocumentationViewer.js
```diff
@@ -1,39 +1,34 @@
...
-
+  const search_bar = builder.get_object("search_bar");
   const button_sidebar = builder.get_object("button_sidebar");
   const button_search = builder.get_object("button_search");
   const search_entry = builder.get_object("search_entry");
-  const search_bar = builder.get_object("search_bar");
-  const back = builder.get_object("button_back");
-  const forward = builder.get_object("button_forward");
+  const button_back = builder.get_object("button_back");
+  const button_forward = builder.get_object("button_forward");
```
sonnyp: let's be consistent in naming schemes

src/DocumentationViewer.js
```diff
@@ -43,11 +38,20 @@ export default function DocumentationViewer({
     }
   });
 
-  back.connect("clicked", () => {
+  webview.get_back_forward_list().connect("changed", () => {
+    updateButtons();
+  });
```
sonnyp, line 43: this is important because not only load evenets can change `can_go_back` and `can_go_forward`

If you click on a anchor like a title or something that navigates somewhere else in the current page, there is no load event but it is still possible to go back

src/init.js
```diff
@@ -66,12 +66,6 @@ Gio._promisify(
   "enumerate_children_finish",
 );
 
-Gio._promisify(
-  Gio.FileEnumerator.prototype,
-  "next_files_async",
-  "next_files_finish",
```
sonnyp: not used

sonnyp (approved):
Good job!

I found a few more issues that I fixed directly in https://github.com/sonnyp/Workbench/pull/358/commits/c5ba2fb0c350ee360afa6b804846a57b87694da1

Code formatting isn't important, for the rest please look at it and let me know if something doesn't make sense. I have added some comments.

sonnyp:
@AkshayWarrier when you have time, here is the first follow up https://github.com/sonnyp/Workbench/issues/373

In the meantime, I will look into how to proceed with the search/sidebar/deisng.

AkshayWarrier:
Okay!

src/DocumentationViewer.js
AkshayWarrier: Ah I see, got it.

PR #361, opened 2023-06-23
library: Add ListModel Entry

### LINKED ISSUE #294: Add tutorial-demo for GListModel with GTK & Adwaita
Add a demo and tutorial that explains GListModel, and why models are important in GTK and Adwaita.

* [ ] Decide which APIs are most important to explain concepts and usage
* [ ] Use order that introduces concepts incrementally (e.g. model, filtering, sorting, ...)
* [ ] (Maybe) use a very simple UI to present the effect of models

 Relevant classes:

* https://docs.gtk.org/gio/iface.ListModel.html
* https://docs.gtk.org/gtk4/class.StringList.html
* https://docs.gtk.org/gtk4/class.FilterListModel.html
  * https://docs.gtk.org/gtk4/class.Filter.html
* https://docs.gtk.org/gtk4/class.SortListModel.html
  * https://docs.gtk.org/gtk4/class.Sorter.html

SoNiC-HeRE:
Closes #294

### DIFF
--- a/src/Library/demos/List Model/main.blp
+++ b/src/Library/demos/List Model/main.blp
@@ -41,7 +41,7 @@ Adw.StatusPage {
             hexpand: true;
             Frame {
               FlowBox flow_box {
-                width-request: 240;
+                width-request: 360;
                 orientation: horizontal;
                 selection-mode: none;
                 styles ["card"]
--- a/src/Library/demos/List Model/main.blp
+++ b/src/Library/demos/List Model/main.blp
@@ -2,7 +2,7 @@ using Gtk 4.0;
 using Adw 1;
 
 Adw.StatusPage {
-  title: "GList Model";
+  title: "List Model";
   description: _("List models are a simple interface for ordered lists of GObject instances");
 
   Box {
@@ -26,9 +26,8 @@ Adw.StatusPage {
           Box {
             halign: center;
             ListBox list_box {
-              activate-on-single-click: true;
               selection-mode: none;
-              width-request: 360;
+              width-request: 240;
               styles ["boxed-list"]
             }
           };
@@ -41,8 +40,7 @@ Adw.StatusPage {
           Box {
             halign: center;
             FlowBox flow_box {
-              width-request: 360;
-              activate-on-single-click: true;
+              width-request: 240;
               orientation: horizontal;
               selection-mode: none;
               styles ["card"]
@@ -57,17 +55,19 @@ Adw.StatusPage {
         child:
           Box {
             orientation: vertical;
-            spacing: 24;
+            spacing: 12;
             halign: center;
-            ListBox list_box_editable {
-              width-request: 360;
-              activate-on-single-click: true;
-              selection-mode: none;
-              styles ["card"]
-            }
+
             SearchEntry search_entry {
               halign: center;
               placeholder-text: _("Start searching");
+            }
+
+            ListBox list_box_editable {
+              width-request: 240;
+              activate-on-single-click: true;
+              selection-mode: none;
+              styles ["card"]
             }
           };
       }
@@ -92,9 +92,3 @@ Adw.StatusPage {
     }
   }
 }
-
-
-
-
-
-
--- a/src/Library/demos/List Model/main.blp
+++ b/src/Library/demos/List Model/main.blp
@@ -15,7 +15,6 @@ Adw.StatusPage {
     }
 
     Gtk.Stack stack {
-      transition-duration: 300;
       transition-type: crossfade;
       vexpand: true;
 
@@ -55,34 +54,35 @@ Adw.StatusPage {
         child:
           Box {
             orientation: vertical;
-            spacing: 12;
+            spacing: 18;
             halign: center;
 
-            SearchEntry search_entry {
+            Box {
               halign: center;
-              placeholder-text: _("Start searching");
+              styles ["linked"]
+
+              SearchEntry search_entry {
+                hexpand: true;
+                placeholder-text: _("Start searching");
+              }
+
+              Button add {
+                icon-name: "list-add-symbolic";
+                tooltip-text: _("Add Item");
+              }
+
+              Button remove {
+                icon-name: "list-remove-symbolic";
+                tooltip-text: _("Remove Item");
+              }
             }
 
             ListBox list_box_editable {
               width-request: 240;
-              activate-on-single-click: true;
-              selection-mode: none;
-              styles ["card"]
+              selection-mode: single;
+              styles ["boxed-list"]
             }
           };
-      }
-    }
-
-    Box {
-      styles ["linked"]
-      halign: center;
-      margin-bottom: 18;
-
-      Button add {
-        label: _("Add Items");
-      }
-      Button remove {
-        label: _("Remove Items");
       }
     }
 
@@ -92,3 +92,4 @@ Adw.StatusPage {
     }
   }
 }
+
--- a/src/Library/demos/List Model/main.blp
+++ b/src/Library/demos/List Model/main.blp
@@ -77,6 +77,7 @@ Adw.StatusPage {
               Button remove {
                 icon-name: "list-remove-symbolic";
                 tooltip-text: _("Remove Item");
+                sensitive: false;
               }
             }
 
@@ -96,4 +97,5 @@ Adw.StatusPage {
     }
   }
 }
+
 
--- a/src/Library/demos/List Model/main.blp
+++ b/src/Library/demos/List Model/main.blp
@@ -65,14 +65,9 @@ Adw.StatusPage {
               selection-mode: none;
               styles ["card"]
             }
-            Entry query_for_filter {
+            SearchEntry search_entry {
               halign: center;
               placeholder-text: _("Start searching");
-            }
-            Button submit {
-              halign: center;
-              label: _("Search");
-              styles ["suggested-action"]
             }
           };
       }
@@ -97,6 +92,7 @@ Adw.StatusPage {
     }
   }
 }
+
 
 
 
--- a/src/Library/demos/List Model/main.blp
+++ b/src/Library/demos/List Model/main.blp
@@ -4,98 +4,101 @@ using Adw 1;
 Adw.StatusPage {
   title: "List Model";
   description: _("List models are a simple interface for ordered lists of GObject instances");
+  valign: start;
 
-  Box {
-    orientation: vertical;
-    spacing: 24;
+  Adw.Clamp {
+    maximum-size: 640;
+    Box {
+      orientation: vertical;
+      spacing: 24;
 
-    Gtk.StackSwitcher {
-      stack: stack;
-      halign: center;
-    }
-
-    Gtk.Stack stack {
-      transition-type: crossfade;
-      vexpand: true;
-
-      Gtk.StackPage {
-        name: "listbox";
-        title: _("List Box");
-        child:
-          Box {
-            halign: center;
-            ListBox list_box {
-              width-request: 270;
-              selection-mode: none;
-              styles ["boxed-list"]
-            }
-          };
+      LinkButton{
+        label: "Documentation";
+        uri: "https://gjs.guide/guides/gio/list-models.html";
       }
 
-      Gtk.StackPage {
-        name: "flowbox";
-        title: _("Flow Box");
-        child:
-          Box {
-            halign: center;
-            hexpand: true;
+      Gtk.StackSwitcher {
+        stack: stack;
+        halign: center;
+      }
+
+      Gtk.Stack stack {
+        transition-type: crossfade;
+        vexpand: true;
+
+        Gtk.StackPage {
+          name: "listbox";
+          title: _("List Box");
+          child:
+            Box {
+              ListBox list_box {
+                hexpand: true;
+                valign: start;
+                selection-mode: none;
+                styles ["boxed-list"]
+              }
+            };
+        }
+
+        Gtk.StackPage {
+          name: "flowbox";
+          title: _("Flow Box");
+          child:
             Frame {
+              hexpand: true;
+              valign: start;
               FlowBox flow_box {
-                width-request: 360;
                 orientation: horizontal;
                 selection-mode: none;
                 styles ["card"]
               }
-            }
-          };
-      }
+            };
+        }
 
 
-      Gtk.StackPage {
-        name: "Filter";
-        title: _("Edit");
-        child:
-          Box {
-            orientation: vertical;
-            spacing: 18;
-            halign: center;
-
+        Gtk.StackPage {
+          name: "Filter";
+          title: _("Edit");
+          child:
             Box {
-              halign: center;
-              styles ["linked"]
+              orientation: vertical;
+              spacing: 12;
 
-              SearchEntry search_entry {
+              Box {
+                halign: center;
+                styles ["linked"]
+
+                SearchEntry search_entry {
+                  placeholder-text: _("Start searching");
+                }
+
+                Button add {
+                  icon-name: "list-add-symbolic";
+                  tooltip-text: _("Add Item");
+                }
+
+                Button remove {
+                  icon-name: "list-remove-symbolic";
+                  tooltip-text: _("Remove Item");
+                  sensitive: false;
+                }
+              }
+
+              ListBox list_box_editable {
                 hexpand: true;
-                placeholder-text: _("Start searching");
+                valign: start;
+                activate-on-single-click: true;
+                selection-mode: single;
+                styles ["boxed-list"]
               }
-
-              Button add {
-                icon-name: "list-add-symbolic";
-                tooltip-text: _("Add Item");
-              }
-
-              Button remove {
-                icon-name: "list-remove-symbolic";
-                tooltip-text: _("Remove Item");
-                sensitive: false;
-              }
-            }
-
-            ListBox list_box_editable {
-              activate-on-single-click: true;
-              selection-mode: single;
-              styles ["boxed-list"]
-            }
-          };
-      }
-    }
-
-    LinkButton{
-      label: "Documentation";
-      uri: "https://gjs.guide/guides/gio/list-models.html";
+            };
+          }
+        }
     }
   }
 }
+
+
 
 
 
--- a/src/Library/demos/List Model/main.blp
+++ b/src/Library/demos/List Model/main.blp
@@ -12,7 +12,7 @@ Adw.StatusPage {
       orientation: vertical;
       spacing: 24;
 
-      LinkButton{
+      LinkButton {
         label: "Documentation";
         uri: "https://gjs.guide/guides/gio/list-models.html";
       }
@@ -50,7 +50,6 @@ Adw.StatusPage {
               FlowBox flow_box {
                 orientation: horizontal;
                 selection-mode: none;
-                styles ["card"]
               }
             };
         }
@@ -92,8 +91,8 @@ Adw.StatusPage {
                 styles ["boxed-list"]
               }
             };
-          }
         }
+      }
     }
   }
 }
--- a/src/Library/demos/List Model/main.blp
+++ b/src/Library/demos/List Model/main.blp
@@ -2,7 +2,7 @@ using Gtk 4.0;
 using Adw 1;
 
 Adw.StatusPage {
-  title: "GList Model";
+  title: "List Model";
   description: _("List models are a simple interface for ordered lists of GObject instances");
 
   Box {
@@ -41,18 +41,6 @@ Adw.StatusPage {
           width-request: 360;
 
           styles ["boxed-list"]
-
-          Adw.ActionRow {
-            title: _("Item 1");
-          }
-
-          Adw.ActionRow {
-            title: _("Item 2");
-          }
-
-          Adw.ActionRow {
-            title: _("Item 3");
-          }
         }
       }
     }
@@ -71,18 +59,6 @@ Adw.StatusPage {
           orientation: horizontal;
           selection-mode: none;
           styles ["card"]
-
-          Adw.ActionRow {
-            title: _("Item 1");
-          }
-
-          Adw.ActionRow {
-            title: _("Item 2");
-          }
-
-          Adw.ActionRow {
-            title: _("Item 3");
-          }
         }
       }
     }
@@ -106,6 +82,7 @@ Adw.StatusPage {
     }
   }
 }
+
 
 
 
--- a/src/Library/demos/List Model/main.blp
+++ b/src/Library/demos/List Model/main.blp
@@ -1,0 +1,111 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "GList Model";
+  description: _("List models are a simple interface for ordered lists of GObject instances");
+
+  Box {
+    orientation: vertical;
+
+    Box {
+      orientation: horizontal;
+      halign: center;
+      margin-bottom: 24;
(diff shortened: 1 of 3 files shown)


SoNiC-HeRE:
What I'm trying to achieve is when i click on add items button it should add rows with labels to the existing listbox

src/Library/demos/List Model/main.js
```diff
@@ -0,0 +1,51 @@
...
+  const isActive = toggle_flow_box.get_active();
+  reveal_flow_box.reveal_child = isActive;
+  if (isActive) {
+    console.log("FlowBox view toggled on");
+  } else {
+    console.log("FlowBox view toggled off");
+  }
+});
+
+const model = new Gio.ListStore();
```
andyholmes, line 35:
```suggestion
const model = new Gtk.StringList([
  "Apple",
  "Banana",
  "Carrot",
  "Dragon Fruit",
]);
```

src/Library/demos/List Model/main.js
```diff
@@ -0,0 +1,51 @@
...
+list_box.bind_model(model, addRows);
```
andyholmes, line 36:
```suggestion
list_box.bind_model(model, createRow);
```

andyholmes (commented):
Does that make more sense?

SoNiC-HeRE:
@andyholmes would it be good to demonstrate search function as well?
I was thinking of adding a search entry to search for an item and then delete it; wdyt?

P.S: 
- Implemented Add and Remove functions for respective model
- A little bit of UI improvement is needed

andyholmes:
There's no `Gio.ListModel` in your demo, that I can see.

The main principle to demonstrate here is to have a single `Gio.ListModel` instance, like `Gtk.StringList` or `Gio.ListStore`, and use it to automatically populate different widgets using `Gtk.ListBox.bind_model()` and `Gtk.FlowBox.bind_model()`.

Your add/remove buttons should call `Gtk.StringList.append()` or `Gio.ListStore.append()` and the `Gtk.ListBoxCreateWidgetFunc` and `Gtk.FlowBoxCreateWidgetFunc` should create the appropriate widget for the container (e.g. `Adw.ActionRow` for `Gtk.ListBox`, `Gtk.Box` for `Gtk.FlowBox`).

Then if you want, you implement searching and filtering for either container by wrapping the `Gio.ListModel` in `Gtk.FilterListModel` and binding the model to that instead.

SoNiC-HeRE:
All right, I'll make the required changes

SoNiC-HeRE:
Implemented for the following changes:
- Used StringList to populate both List and Flow Box
- Used ListModel methods for adding/removing
- Used toggle mode to make the UI minimalistic and depict one usecase at a time

src/Library/demos/List Model/main.blp
```diff
@@ -0,0 +1,88 @@
...
+
+        FlowBox flow_box {
+          width-request: 360;
+          activate-on-single-click: true;
+          orientation: horizontal;
+          selection-mode: none;
+          styles ["card"]
+        }
+      }
+    }
```
andyholmes, line 64: Let's clean this all up with some higher-level widget.
```suggestion
    Gtk.StackSwitcher {
      stack: stack;
      halign: center;
    }
    
    Gtk.Stack stack {
      transition-duration: 300;
      transition-type: crossfade;
      vexpand: true;
    
      Gtk.StackPage {
        name: "listbox";
        title: _("List Box");
        child:
          Box {
            halign: center;

            ListBox list_box {
              activate-on-single-click: true;
              selection-mode: none;
              width-request: 360;

              styles ["boxed-list"]
            }
          };
      }
    
      Gtk.StackPage {
        name: "flowbox";
        title: _("Flow Box");
        child:
          Box {
            halign: center;

            FlowBox flow_box {
              width-request: 360;
              activate-on-single-click: true;
              orientation: horizontal;
              selection-mode: none;
              styles ["card"]
            }
          };
      }
    }
```

src/Library/demos/List Model/main.js
```diff
@@ -0,0 +1,86 @@
...
+const toggle_list_box = workbench.builder.get_object("toggle_list_box");
+const toggle_flow_box = workbench.builder.get_object("toggle_flow_box");
+const reveal_list_box = workbench.builder.get_object("reveal_list_box");
+const reveal_flow_box = workbench.builder.get_object("reveal_flow_box");
+const list_box = workbench.builder.get_object("list_box");
+const flow_box = workbench.builder.get_object("flow_box");
+const add = workbench.builder.get_object("add");
+const remove = workbench.builder.get_object("remove");
+
+//ListModel initialization and binding
```
andyholmes, line 15:
```suggestion

// Model
```

src/Library/demos/List Model/main.js
```diff
@@ -0,0 +1,86 @@
...
+  add.connect("clicked", () => {
+    const newItem = `Item ${item}`;
+    model.append(newItem); // Append the new item as an array to the model
+    item++;
+  });
+  remove.connect("clicked", () => {
+    const length = model.get_n_items();
+    model.remove(length - 1);
+  });
+}
```
andyholmes, line 64: Actually we only need one set of these, since we're "controlling" the "model" in a model-view-controller pattern. Stripping this part down makes that a little bit clearer.

```suggestion
// Controller
add.connect("clicked", () => {
  const new_item = `Item ${item}`;
  model.append(new_item);
  item++;
});

remove.connect("clicked", () => {
  const n_items = model.get_n_items();
  model.remove(n_items - 1);
});
```

src/Library/demos/List Model/main.js
```diff
@@ -0,0 +1,86 @@
...
+// Check if FlowBox is Active
+toggle_flow_box.connect("toggled", () => {
+  const isActive = toggle_flow_box.get_active();
+  reveal_flow_box.reveal_child = isActive;
+  if (isActive) {
+    console.log("FlowBox toggled on");
+  } else {
+    console.log("FlowBox toggled off");
+  }
+});
```
andyholmes, line 86: This is a good idea to identify when the view has changed, but let's keep it simple so it doesn't distract from the rest of the demo.

```suggestion
// View
stack.connect("notify::visible-child", () => {
  if (stack.visible_child === list_box) {
    console.log("View: List Box");
  } else {
    console.log("View: Flow Box");
  }
});
```

src/Library/demos/List Model/main.js
```diff
@@ -0,0 +1,86 @@
...
+  });
+  return listRow;
+}
+
+function createItemForFlowBox(listItem) {
+  const listRow = new Adw.ActionRow({
+    title: listItem.string,
+  });
+  return listRow;
+}
```
andyholmes, line 31: I think we should pick a more squarish/grid-like widget for this one, but we can tweak that a bit later.

src/Library/demos/List Model/main.js
```diff
@@ -0,0 +1,86 @@
+import Gio from "gi://Gio";
+import Gdk from "gi://Gdk";
+import Gtk from "gi://Gtk";
+import Adw from "gi://Adw";
+
+const toggle_list_box = workbench.builder.get_object("toggle_list_box");
+const toggle_flow_box = workbench.builder.get_object("toggle_flow_box");
+const reveal_list_box = workbench.builder.get_object("reveal_list_box");
+const reveal_flow_box = workbench.builder.get_object("reveal_flow_box");
```
andyholmes, line 9: Much simpler :slightly_smiling_face: 
```suggestion
const stack = workbench.builder.get_object("stack");
```

src/Library/demos/List Model/main.blp
```diff
@@ -0,0 +1,88 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "List Model";
+  description: _("List models are a simple interface for ordered lists of GObject instances");
+
+  Box {
+    orientation: vertical;
+
```
andyholmes, line 10: We can let the containing box do the child spacing.
```suggestion
    orientation: vertical;
    spacing: 24;

```

andyholmes (changes requested):
Much better, I think you're getting how model-view-controller works in GTK now!

With regard to your idea about a text entry, let's clean the UI up to use the `Gtk.Stack` as in the review suggestions, then how about adding a third page like `Edit` or something?

A third page could be a list box, flow box, or we could demonstrate the `items-changed` callback if you're feeling adventurous. Either way, for the third view we should put the `Gtk.StringList` inside another model like `Gtk.FilterListModel`, and have the entry change the filter.

SoNiC-HeRE:
Okay, What I'm trying to do here is when the user clicks on the search button i want to check if the searched text matches one of the strings in ```model``` by updating the ```Filter.search``` property and then using ```Filter.match(searched_text)``` to evaluate whether it was found or not but i get encountered by an error which states searched_text is not of type GObject which is true; but I'm not really sure how to use the string for evaluation in filter;Also i created a new Gtk.PropertyExpression as:
 ```js
  const filter = new Gtk.StringFilter({
  expression: Gtk.PropertyExpression.new(model, null, "strings"),
  ignore_case: true,
  match_mode: Gtk.StringFilterMatchMode.EXACT,
});
```
 
but got an error 
```
gtk_string_filter_set_expression: assertion 'expression == NULL || gtk_expression_get_value_type (expression) == G_TYPE_STRING' failed

```

**Comment by andyholmes** [comment]
So I think you're pretty close.

The filter is going to be passed an item from the `GListModel`, which is a `GtkStringList`, so it should be expecting a `GtkStringObject` and we want to get the `GtkStringObject:string` property:

```js
const search_expression = Gtk.PropertyExpression.new(Gtk.StringObject, null, "string");
```

`GtkStringFilter` will use the expression to retrieve the item's string value, then compare it to another string (i.e. the search terms) by calling `Gtk.Filter.match()`, which always expects a GObject. Since we want to be able to filter, not just find, the `match-mode` should be set to `Gtk.StringFilterMatchMode.SUBSTRING` instead of `EXACT`:

```js
const filter = new Gtk.StringFilter({
 expression: search_expression,
 ignore_case: true,
 match_mode: Gtk.StringFilterMatchMode.SUBSTRING,
});
```

When we get to the search entry, let's skip the button and just filter directly:

```js
const search_entry = new Gtk.SearchEntry();

search_entry.connect("search-changed", (editable) => {
  console.log("Search terms changed");

  search_filter.search = editable.get_text();

  // This will not work, because you are passing a string
  // where a GObject is expected. Calling this function is
  // also what the filter does for you, so there's no need
  search_filter.match(editable.get_text());
});
```

Once that's done, I think we've done enough for the demo and we'll just cleanup the UX a bit. Namely I don't think any widgets should be underneath the view stack, because everything jumps around when items are added/removed/filtered.

SoNiC-HeRE:
@andyholmes I've applied the suggested changes; talking about UX - i was thinking to maybe change the widget that populates all the 3 models? Also we need to take care of the void space created cause of the SearchEntry widget, wdyt?

src/Library/demos/List Model/main.js
```diff
@@ -75,17 +77,15 @@ remove.connect("clicked", () => {
...
-    console.log("Item exists");
-  } else  {
-    console.log("Item does not exist");
-  }
+  filter.search = search_entry.get_text();
+
+  /*  This will not work, because you are passing a string
+   where a GObject is expected. Calling this function is
+   also what the filter does for you, so there's no need
+   search_filter.match(editable.get_text()); */
```
andyholmes, line 88: You can remove this.

andyholmes:
> @andyholmes I've applied the suggested changes; talking about UX - i was thinking to maybe change the widget that populates all the 3 models?

How do you mean change the widget? Do you mean have a separate `GListModel` in addition to the `GtkStringList` and swap them in and out?

> Also we need to take care of the void space created cause of the SearchEntry widget, wdyt?

I think the reasonable place to have the entry is above the views, because they are what will change size the most. Either a margin, or just allowing the widget to shift when the stack page changes I think is okay.

> P.S: I've also kept the comment about filter.match() property in the code - thought it would be useful, i can remove it if you say

Yeah, I think remove it. We're just demonstrating how to use GListModel here and how they can be wrapped in other models. We can cover actual use of filter models and sorting models in another demo.

SoNiC-HeRE:
> > @andyholmes I've applied the suggested changes; talking about UX - i was thinking to maybe change the widget that populates all the 3 models?
> 
> How do you mean change the widget? Do you mean have a separate `GListModel` in addition to the `GtkStringList` and swap them in and out?
> 
> > Also we need to take care of the void space created cause of the SearchEntry widget, wdyt?
> 
> I think the reasonable place to have the entry is above the views, because they are what will change size the most. Either a margin, or just allowing the widget to shift when the stack page changes I think is okay.
> 
> > P.S: I've also kept the comment about filter.match() property in the code - thought it would be useful, i can remove it if you say
> 
> Yeah, I think remove it. We're just demonstrating how to use GListModel here and how they can be wrapped in other models. We can cover actual use of filter models and sorting models in another demo.

I meant regarding this https://github.com/sonnyp/Workbench/pull/361#discussion_r1245710719

src/Library/demos/List Model/main.blp
```diff
@@ -0,0 +1,94 @@
...
+            SearchEntry search_entry {
+              halign: center;
+              placeholder-text: _("Start searching");
+            }
+
+            ListBox list_box_editable {
+              width-request: 240;
+              activate-on-single-click: true;
+              selection-mode: none;
```
andyholmes, line 69: Let's set this to `single` and mark the remove button `active` when there is a selected row, then remove the item from the model with at the row's index position. I think that's sensible UX.

src/Library/demos/List Model/main.blp
```diff
@@ -0,0 +1,94 @@
...
+      halign: center;
+      margin-bottom: 18;
+
+      Button add {
+        label: _("Add Items");
+      }
+      Button remove {
+        label: _("Remove Items");
+      }
+    }
```
andyholmes, line 87: Let's move this above the "Edit" page's list, and connect the whole thing so it looks something like this:

```js
Box {
  orientation: vertical;
  spacing: 18;

  Box {
    styles ["linked"]

    SearchEntry search_entry {
      hexpand: true;
      placeholder-text: _("Start searching");
    }
    Button add {
      icon-name: "list-add-symbolic";
      tooltip-text: _("Add Item");
    }
    Button remove {
      icon-name: "list-remove-symbolic";
      tooltip-text: _("Remove Item");
    }
  }

  ListBox list_box_editable {
    activate-on-single-click: true;
    selection-mode: single;
    styles ["boxed-list"]
  }
}
```

src/Library/demos/List Model/main.blp
```diff
@@ -0,0 +1,94 @@
...
+    orientation: vertical;
+    spacing: 24;
+
+    Gtk.StackSwitcher {
+      stack: stack;
+      halign: center;
+    }
+
+    Gtk.Stack stack {
+      transition-duration: 300;
```
andyholmes, line 18:
```suggestion
```

Not needed.

src/Library/demos/List Model/main.blp
```diff
@@ -0,0 +1,94 @@
...
+  title: "List Model";
+  description: _("List models are a simple interface for ordered lists of GObject instances");
+
+  Box {
+    orientation: vertical;
+    spacing: 24;
+
+    Gtk.StackSwitcher {
+      stack: stack;
+      halign: center;
```
andyholmes, line 14: This one can stay, so the switch is nicely centered about the lists.

src/Library/demos/List Model/main.js
andyholmes: I think what we should do here is return an item like:

```js
new Adw.Bin({
    width_request: 96,
    height_request: 96,
    css_classes: ['card'],
    child: new Gtk.Label({
        label: listItem.string,
        halign: Gtk.Align.CENTER,
        hexpand: true;
        valign: Gtk.Align.CENTER,
    }),
});
```

Then change the `Box` wrapping the `FlowBox` with a `Frame`.

src/Library/demos/List Model/main.js
```diff
@@ -0,0 +1,89 @@
...
+
+// Controller
+
+model.connect("items-changed", (list, position, removed, added) => {
+  console.log(
+    `position: ${position}, Item removed? ${Boolean(
+      removed,
+    )}, Item added? ${Boolean(added)}`,
+  );
+});
```
andyholmes, line 69:
```suggestion
```

src/Library/demos/List Model/main.js
```diff
@@ -0,0 +1,89 @@
...
+const flow_box = workbench.builder.get_object("flow_box");
+const add = workbench.builder.get_object("add");
+const remove = workbench.builder.get_object("remove");
+const list_box_editable = workbench.builder.get_object("list_box_editable");
+const search_entry = workbench.builder.get_object("search_entry");
+
+//Model
+const model = new Gtk.StringList();
+let item = 1;
+
```
andyholmes, line 18:
```suggestion
let item = 1;

model.connect("items-changed", (list, position, removed, added) => {
  console.log(
    `position: ${position}, Item removed? ${Boolean(
      removed,
    )}, Item added? ${Boolean(added)}`,
  );
});
```

andyholmes (commented):
A few more notes, and thoughts I had about UX.

SoNiC-HeRE:
Implemented the suggested changes:
- Improved UI
- Improved UX
- Refined Code

src/Library/demos/List Model/main.js
```diff
@@ -0,0 +1,107 @@
...
+    title: listItem.string,
+  });
+  return listRow;
+}
+
+function createItemForFlowBox(listItem) {
+  const listBox = new Adw.Bin({
+    width_request: 96,
+    height_request: 96,
+    css_classes: ["card"],
```
andyholmes, line 55:
```suggestion
    css_classes: ["card"],
    valign: start;
```

One last tweak.

src/Library/demos/List Model/main.blp
```diff
@@ -0,0 +1,101 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "List Model";
+  description: _("List models are a simple interface for ordered lists of GObject instances");
```
andyholmes, line 6:
```suggestion
  description: _("List models are a simple interface for ordered lists of GObject instances");
  valign: start;
```

src/Library/demos/List Model/main.blp
```diff
@@ -0,0 +1,101 @@
...
+        child:
+          Box {
+            halign: center;
+            hexpand: true;
+            Frame {
+              FlowBox flow_box {
+                width-request: 360;
+                orientation: horizontal;
+                selection-mode: none;
+                styles ["card"]
```
andyholmes, line 47:
```suggestion
```

src/Library/demos/List Model/main.blp
```diff
@@ -0,0 +1,101 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "List Model";
+  description: _("List models are a simple interface for ordered lists of GObject instances");
+
+  Box {
```
andyholmes, line 8: Okay, so let's do this:

Except for the `StackSwitcher` and FlowBox items, go through and remove all the `halign: center` and `width-request` properties. Then put this `Box` inside an `Adw.Clamp` with `maximum-size: 640;` so you have something like:

```js
Adw.Clamp {
  maximum-size: 640;
  child:
    Box current_box {
      // ...
    };
}
```

Then set `hexpand: true` and `valign: start` on `list_box`, `flow_box` and `list_box_editable`. That should even up the whole demo.

src/Library/demos/List Model/main.blp
```diff
@@ -0,0 +1,101 @@
...
+            orientation: vertical;
+            spacing: 18;
+            halign: center;
+
+            Box {
+              halign: center;
+              styles ["linked"]
+
+              SearchEntry search_entry {
+                hexpand: true;
```
andyholmes, line 68:
```suggestion
```

src/Library/demos/List Model/main.js
```diff
@@ -0,0 +1,107 @@
...
+  filter.search = searchText;
+});
+
+// View
+stack.connect("notify::visible-child", () => {
+  console.log("View changed");
+});
+
+list_box_editable.connect("row-selected", () => {
+  remove.sensitive = true;
```
andyholmes, line 106: This is good, but it's possible to "unselect" all rows, so better to do something like:

```suggestion
  remove.sensitive = list_box_editable.get_selected_row() !== null;
```

src/Library/demos/List Model/main.js
```diff
@@ -0,0 +1,107 @@
...
+const stack = workbench.builder.get_object("stack");
+const list_box = workbench.builder.get_object("list_box");
+const flow_box = workbench.builder.get_object("flow_box");
+const add = workbench.builder.get_object("add");
+const remove = workbench.builder.get_object("remove");
+const list_box_editable = workbench.builder.get_object("list_box_editable");
+const search_entry = workbench.builder.get_object("search_entry");
+
+//Model
+const model = new Gtk.StringList();
```
andyholmes, line 16:
```suggestion
const model = new Gtk.StringList({
  strings: [
    "Default Item 1",
    "Default Item 2",
    "Default Item 3",
  ],
});
```

I think this is the easiest way to avoid dealing with empty states in the demo. What do you think?

src/Library/demos/List Model/main.blp
```diff
@@ -0,0 +1,101 @@
...
+          };
+      }
+
+      Gtk.StackPage {
+        name: "flowbox";
+        title: _("Flow Box");
+        child:
+          Box {
+            halign: center;
+            hexpand: true;
```
andyholmes, line 41: You can remove this `Box`, the `Frame` is taking its place.

src/Library/demos/List Model/main.js
```diff
@@ -0,0 +1,107 @@
...
+
+remove.connect("clicked", () => {
+  const selectedRow = list_box_editable.get_selected_row();
+  const index = selectedRow.get_index();
+  model.remove(index);
+
+  const items = model.get_n_items();
+  if (items === 0) {
+    remove.sensitive = false;
+  }
```
andyholmes, line 92: I think this should be handled by the `row-selected` signal, no?

andyholmes (changes requested):
Looking really good.

There are some final UI fixes, then this one is good to go. I think this will be a good template for the `GtkListView`, `GtkGridView` and other GListModel demos.

src/Library/demos/List Model/main.js
SoNiC-HeRE: Yes, it's obviously better than a void space

src/Library/demos/List Model/main.js
SoNiC-HeRE: yes

SoNiC-HeRE:
- Implemented the suggested changes
- Increased the size of flowbox items that visually looks better and occupies the available space more accurately

SoNiC-HeRE:
- Implemented the suggested changes
- Increased the size of flowbox items that visually looks better and occupies the available space more accurately
- Moved the ```linkbutton``` to top since it did'nt look good below, lmk if i should place it to the original place

andyholmes (approved):
This is a really nice demo, well done :+1: 

I think we should definitely use this as a template for the list/grid/column view demos.

andyholmes:
>     * Moved the `linkbutton` to top since it did'nt look good below, lmk if i should place it to the original place

I think that was a good choice. We might want to update and expand the links later, but I think we should get into the other list model widgets first, and see how we want to organize them.

PR #363, opened 2023-06-24
Multi windows / sessions support

sonnyp:
Todo

* [x] Remove "Import File" (not compatible with the new workflow)
* [x] Remove all code related to opening files
* [x] Add multi sessions/windows support
* [x] Migrate the state from v44 to the new format
* [x] Move settings to sessions instead of globals
* [ ] ~Each window needs to be its own process (Console)~
* [x] Fix vala
* [x] no more globals "langs" documets/text
* [x] "// Sorry this is not avaible for Vala yet"
* [x] Remember language for new sessions

Known issue:

* [ ] All sessions / windows share the same Console / stdout
* [ ] Deleted sessions (by closing the window for example) go to the trash and can fill it pretty quickly
* [ ] We should ask for confirmation before closing the session (unless it's an untouched demo)


data/app.gschema.xml
```diff
@@ -22,9 +24,6 @@
     <key name="code-language" type="i">
       <default>0</default>
     </key>
-    <key name="selected-demo" type="s">
-      <default>"Welcome"</default>
-    </key>
```
sonnyp, line 27: unused

src/langs/vala/Compiler.js
```diff
@@ -1,39 +1,33 @@
...
+import { decode } from "../../util.js";
 
-export default function Compiler(data_dir) {
-  const code_file = Gio.File.new_for_path(
-    GLib.build_filenamev([data_dir, "code.vala"]),
-  );
-  const module_file = Gio.File.new_for_path(
-    GLib.build_filenamev([data_dir, "libworkbenchcode.so"]),
-  );
+export default function Compiler({ session }) {
```
sonnyp, line 6: @lw64 ok so basically the idea is that each window is now a "session" with its own "session" folder under `~/.var/app/re.sonny.Workbench.Devel/data/re.sonny.Workbench.Devel/sessions/`

Therefor the compiler now scopes everything to the current session folder.

src/langs/vala/Compiler.js
```diff
@@ -1,39 +1,33 @@
...
       return;
     }
-    await Promise.all([
-      code_file.replace_contents_async(
-        new GLib.Bytes(code || " "),
-        null,
-        false,
-        Gio.FileCreateFlags.NONE,
-        null,
-      ),
```
sonnyp, line 28: I'm not sure why we were duplicating the existing vala code, but I'm not passing directly the buffer file (`sessions/x/main.vala`) to the compiler.

Makes sense?

src/langs/vala/Compiler.js
```diff
@@ -1,39 +1,33 @@
...
-      module_file.delete_async(GLib.PRIORITY_DEFAULT, null).catch(() => {}),
-    ]);
+
+    await module_file.delete_async(GLib.PRIORITY_DEFAULT, null).catch(() => {});
```
sonnyp, line 24: it doesn't look like we need this, the compiler doesn't complain if the module is already there

should we remove this, wdyt?

lw64: Probably can be deleted. If at some point issues arise, it can be added again.

src/langs/vala/Compiler.js
lw64: I think it was a problem when the file contains nothing, so "". That was prevented with replacing it with a " ".

PR #364, opened 2023-06-25
Add CodeFind feature

### LINKED ISSUE #347: Implement code find
# Description

Like most text/code editors, Workbench should have an easy and quick way to find a string in the current text buffer.

Workbench is aimed at newcomers too so we will not have any search option and keep it as simple as possible. We may implement replace in the future.

# Steps

## First

Add a static search bar at the top of `CodeView`. It should have a text entry and a next/preview buttons like so:

image

## Second

Connect the search bar to the buffer using https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/class.SearchContext.html

To keep it simple for users, the search should be case insensitive (`A` matches `a`) without an option to change it.

## Third

Put the search bar in a revealer that is placed on an overlay.  
Add a `Ctrl+F` shortcut to reveal the search bar.

It should behave like the one in GNOME Text Editor.

SoNiC-HeRE:
Closes #347

### DIFF
--- a/src/Library/demos/Code Find/main.blp
+++ b/src/Library/demos/Code Find/main.blp
@@ -3,8 +3,8 @@ using GtkSource 5;
 using Adw 1;
 
 Adw.StatusPage {
-  title: "About Window";
-  description: _("A window showing information about the application.");
+  title: "Code Find Feature";
+  description: _("To implement codefind feature in Workbench");
 
   Box {
     orientation: vertical;
@@ -50,5 +50,6 @@ Adw.StatusPage {
     }
   }
 }
+
 
 
--- a/src/Library/demos/Code Find/main.blp
+++ b/src/Library/demos/Code Find/main.blp
@@ -24,7 +24,7 @@ Adw.StatusPage {
           search-mode-enabled: false;
 
           Box {
-            spacing: 6;
+            spacing: 12;
             SearchEntry search_entry {
               search-delay: 300;
 
--- a/src/Library/demos/Code Find/main.blp
+++ b/src/Library/demos/Code Find/main.blp
@@ -1,0 +1,54 @@
+using Gtk 4.0;
+using GtkSource 5;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "About Window";
+  description: _("A window showing information about the application.");
+
+  Box {
+    orientation: vertical;
+
+    Overlay overlay {
+      ScrolledWindow scroll_view {
+        height-request: 360;
+      }
+      [overlay]
+      Box {
+        halign:end;
+        valign: start;
+
+        Revealer revealer {
+          Box {
+            orientation: horizontal;
+
+            SearchBar search_bar {
+              show-close-button: true;
+              search-mode-enabled: false;
+
+              Box {
+                SearchEntry search_entry {
+                  placeholder-text: _("Start searching");
+                  search-delay: 300;
+                  margin-end: 18;
+                }
+                Box {
+                  valign: center;
+                  styles ["linked"]
+                  Button previous_match {
+                    icon-name: "up";
+                  }
+                  Button next_match {
+                    icon-name: "down";
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
+
+
--- a/src/Library/demos/Code Find/main.blp
+++ b/src/Library/demos/Code Find/main.blp
@@ -31,6 +31,10 @@ Adw.StatusPage {
                 SearchEntry search_entry {
                   placeholder-text: _("Start searching");
                   search-delay: 300;
+
+                  Label occurence_counter {
+                    label: _("");
+                  }
                 }
                 Box {
                   valign: center;
@@ -38,10 +42,12 @@ Adw.StatusPage {
                   Button previous_match {
                     icon-name: "up";
                     tooltip-text: "Move to previous match (Ctrl+Shift+G)";
+                    sensitive: false;
                   }
                   Button next_match {
                     icon-name: "down";
                     tooltip-text: "Move to next match (Ctrl+G)";
+                    sensitive: false;
                   }
                 }
                 Button close_button {
@@ -59,6 +65,7 @@ Adw.StatusPage {
     }
   }
 }
+
 
 
 
--- a/src/Library/demos/Code Find/main.blp
+++ b/src/Library/demos/Code Find/main.blp
@@ -3,8 +3,8 @@ using GtkSource 5;
 using Adw 1;
 
 Adw.StatusPage {
-  title: "Code Find Feature";
-  description: _("To implement codefind feature in Workbench");
+  title: "About Window";
+  description: _("A window showing information about the application.");
 
   Box {
     orientation: vertical;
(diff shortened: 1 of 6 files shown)


SoNiC-HeRE:
The searchbar is not visible at all and when the searchbar is moved outside the template it becomes visible but expands vertically way too much and the codeview isnt implemented.

SoNiC-HeRE:
```  setupSearch() {
    const { buffer } = this;
    this.search_entry.connect("search-changed", () => {
      const searchContext = new Source.SearchContext(buffer);
    });
  }
```
initialising searchcontext throws a bunch of errors;
Using this for reference https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/ctor.SearchContext.new.html

sonnyp:
@SoNiC-HeRE in GJS `new X` is to initialize with properties

try 

```
new SearchContext({
  buffer
})
```

additionally, your code is wrong, I doubt that you're supposed to create a new `SearchContext` on every change

SoNiC-HeRE:
Implemented the requested layout and functionalities suggested in the #347
For the buttons: I need help with the implementation of  ```searchContext.forward_async``` and ```searchContext.backward_async``` functions to iterate through the search results and move the cursor accordingly

SoNiC-HeRE:
> Implemented the requested layout and functionalities suggested in the #347 For the buttons: I need help with the implementation of `searchContext.forward_async` and `searchContext.backward_async` functions to iterate through the search results and move the cursor accordingly

For example: The ```forward_async``` function returns a boolean value if a match was found or not and also returns location for start of match and end of a match. How can I use this info to start my start my iter from first occurence and then move to the next match location iterating through the buffer and moving my cursor.

https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/method.SearchContext.forward_async.html
https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/method.SearchContext.forward_finish.html

SoNiC-HeRE:
``` js  
 this.next_match.connect("clicked", () => {
      settings.search_text = this.search_entry.get_text();
      const searchContext = new Source.SearchContext({ buffer, settings });

      const currentIter = this.buffer.get_iter_at_mark(
      this.buffer.get_insert()
      );

      const [found, iter] = searchContext.forward(currentIter);
        if (found) {
        // Scroll to the previous match
          this.buffer.place_cursor(iter);
          this.buffer.move_mark_by_name("insert", iter);
        // this.source_view.scroll_mark_onscreen(iter);
        } else {
        // Handle no previous match found
        console.log("No previous match found.");
      }});

```

I've implemented this function to make it work such that wherever the cursor is; it get's the next searched-match location by locating highlight and moving down the cursor one step; I've commented the scroll function because it breaks code;

I'm unsure how do i modify this code to make it keep the cursor moving to next match-location not only vertically but horizontally as well

sonnyp (commented):
I pushed a commit that makes next/previous work https://github.com/sonnyp/Workbench/pull/364/commits/dd3b17b443efc84620ff9cd02d4aa03cbafa1564 make sure to `git pull`

This is how I figured it out:

https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/class.SearchContext.html

> The concept of “current match” doesn’t exist yet. A way to highlight differently the current match is to select it.

Ha! 

> In the GtkSourceView source code, there is an example of how to use the search and replace API: see the tests/test-search.c file. It is a mini application for the search and replace, with a basic user interface.

I simply ported the relevant code to GJS see https://gitlab.gnome.org/GNOME/gtksourceview/-/blob/master/tests/test-search.c

Since this isn't easy and there are several things left, I'm asking you again to use a demo for now on which you can iterate quickly and have a short feedback loop. Instead of restarting Workbench every time you try something or make a typo. It will also be easier for me to help you.

Once we're satisfied with the demo and its features, we can move the code to `CodeView`

SoNiC-HeRE:
I'll apply the changes and shift the code to a demo

SoNiC-HeRE:
Ported the feature as a library entry/demo
Successfully ported the features and checked working
- UI is not that great (since this is not a real demo and meant for only testing purposes);
- Also the search and iterate works but is barely visible in dark mode

sonnyp:
> Ported the codefind feature as a library entry/demo

Great, it's much easier now to iterate on the logic and the UI.

> UI is not that great (since this is not a real demo and meant for only testing purposes);

We should iterate on the UI of the search bar in the demo too.

> Also the search and iterate works but is barely visible in dark mode

Works for me (light mode) :D

You can use this:

```
const scheme_manager = Source.StyleSchemeManager.get_default();
const style_manager = Adw.StyleManager.get_default();

function updateScheme() {
    const scheme = scheme_manager.get_scheme(
      style_manager.dark ? "Adwaita-dark" : "Adwaita",
    );
    buffer.set_style_scheme(scheme);
}

updateScheme()
style_manager.connect("notify::dark", updateScheme);
```

It will be useful for the Source View demo anyway.

> What else can i add/improve in the feature?

The ticket says

> It should behave like the one in GNOME Text Editor.

Please make it look and work the same. Including the shortcut for `Esc`. Please pay attention to details such as (but not only)

* the previous/next buttons are disabled if the search entry is empty
* each button has a tooltip with the shortcut

Ignore the buttons "Search & Replace" and "Togge Search options".

SoNiC-HeRE:
Implemented all the functionalities that i saw in the text editor's searchbar.

- I need help with ```line 145 ``` and ```line 155``` as they do not work.
 The updatelabel method works fine when i use the buttons but not using event controllers.

- Also, i tried using ```searchcontext.get_occurrences_count``` function to trigger the sensitivity of buttons to false when it  gives -1 value but it does'nt work.

sonnyp:
There are a bunch of problems. Please compare UI and features with GNOME Text Editor as requested previously.

* The number of occurrences doesn't show before selecting next/previous
* Ctrl+F doesn't focus the entry
* Ctrl+F then Esc then Ctrl+F doesn't work
* The spacing between the elements of the bar are different than GNOME Text Editor
* There is no Clear Entry button in GNOME Text Editor
* GNOME Text Editor doesn't have a placeholder
* The Search Bar corners are rounded in GNOME Text Editor
* "Next match" on the last match should go to the first match (cycles) and vice versa
* After hiding the search (ESC), opening the search bar should immediately enable the search
* Closing the search should not clear the entry

Less obvious ones:

* Enter should be a shortcut to "Move to next Match"
* Shift+Enter should be the a shortcuit to "Move to previous Match"

Let me know if something isn't clear

sonnyp (changes requested):
There are a bunch of problems. Please compare UI and features with GNOME Text Editor as requested previously.

* Ctrl+F doesn't focus the entry (can't type directly after)
* Ctrl+F then Esc then Ctrl+F doesn't work (consider focusing the code view on `ESC`)
* The spacing between the elements of the bar are different than GNOME Text Editor
* There is no Clear Entry button in GNOME Text Editor
* GNOME Text Editor doesn't have a placeholder
* The Search Bar bottom corners are rounded in GNOME Text Editor
* "Next match" on the last match should go to the first match (cycles) and vice versa
* After hiding the search (ESC), opening the search bar should immediately enable the search
* Closing the search should not clear the entry

Less obvious ones:

* Enter should be a shortcut to "Move to next Match"
* Shift+Enter should be the a shortcut to "Move to previous Match"
* The close button should be rounded (check with hovering it with your cursor)

Let me know if something isn't clear

SoNiC-HeRE:
I've successfully implemented the following changes 
-   Ctrl+F doesn't focus the entry (can't type directly after)
-    Ctrl+F then Esc then Ctrl+F doesn't work (consider focusing the code view on ESC)
-    The spacing between the elements of the bar are different than GNOME Text Editor
-    GNOME Text Editor doesn't have a placeholder
-    The Search Bar bottom corners are rounded in GNOME Text Editor
-    "Next match" on the last match should go to the first match (cycles) and vice versa
-    After hiding the search (ESC), opening the search bar should immediately enable the search
-    Closing the search should not clear the entry
-    Shift+Enter should be the a shortcut to "Move to previous Match"
-    The close button should be rounded (check with hovering it with your cursor)

Still Unsolved:
-  There is no Clear Entry button in GNOME Text Editor
   Since we are using Gtk.SearchBar here which precomes with clear entry button; i'm unsure of how to get rid of it/hide it.Also GNOME Text Editor doesn't use Gtk.SearchBar; however using a Gtk.SearchBar is much more useful cause of it's properties and comes with a Gtk.Revealer
-  Enter should be a shortcut to "Move to next Match"
  I've implemented the logic for Enter key shortcut for "Move to next Match" still it doesn't seem to work;However similar code for keybinding Shift+Enter shortcut for "Move to previous Match" works smoothly.
  
  
Other:
  - I tried updating the label of search entry with total no of occurrences whenever a search is made; however the ```get_occurrences_count``` property always returns -1. ```Line no 239 in main.js```

PR #366, opened 2023-06-25
previewer: Fix namespace logic

sonnyp:


src/Library/demos/Web View/main.blp
```diff
@@ -53,7 +54,8 @@ Adw.Clamp {
       }
     }
 
-    Adw.Bin container {
+    WebKit.WebView web_view {
```
sonnyp, line 57: @AkshayWarrier FYI this fixes the bug which prevented using `WebKit.WebView` from Blueprint :)

src/Previewer/utils.js
```diff
@@ -2,13 +2,21 @@ import GObject from "gi://GObject";
...
+    class_name.startsWith(namespace),
+  );
+  if (!namespace) return null;
+
+  const [, name] = class_name.split(namespace);
+  if (!name) return null;
 
-  const [ns, ...rest] = split;
-  return imports.gi[ns]?.[rest.join("")];
+  return imports.gi[namespace]?.[name];
```
sonnyp, line 19: @andyholmes any better idea? The test shows what this function is about:

```js
test("getObjectClass", () => {
  assert.equal(getObjectClass("WebKitWebView"), imports.gi.WebKit.WebView);
});
```

src/Library/demos/Web View/main.blp
AkshayWarrier: Oh this is great! I'll make the same change in https://github.com/sonnyp/Workbench/pull/358 to make use of WebView from Blueprint instead of code once this is merged.

src/Previewer/utils.js
```diff
@@ -2,13 +2,23 @@ import GObject from "gi://GObject";
...
+import GIRepository from "gi://GIRepository";
+// const repository = GIRepository.Repository.get_default();
 
+const namespaces = getNamespaces();
 export function getObjectClass(class_name) {
-  const split = class_name.split(/(?=[A-Z])/);
-  if (split.length < 2) return;
+  const namespace = namespaces.find((namespace) =>
+    class_name.startsWith(namespace),
+  );
```
andyholmes, line 12: This might not be quite good enough. I could imagine a case where `Gtk` matches `GtkSourceCompletionProvider` before `GtkSource`. Maybe the should be sorted first?

sonnyp: good call didn't think of that

fixed in https://github.com/sonnyp/Workbench/pull/366/commits/9a01524a8e24f25367270be1511bbf3fef78a61d

Issue #257, opened 2023-03-29
Add autocompletion


vanillajonathan:
Add auto-completion for GTK class names, properties and styles.

sonnyp:
See https://github.com/sonnyp/Workbench/pull/155

PR #371, opened 2023-06-25
feat: add an internal GI library for Workbench

andyholmes:
Functionality that is static and performance sensitive can be moved here as necessary, but more importantly it can be used to work around problems with introspection in GJS.

cc #257

### DIFF
--- a/src/libworkbench/meson.build
+++ b/src/libworkbench/meson.build
@@ -1,0 +1,116 @@
+# SPDX-License-Identifier: GPL-3.0-only
+# SPDX-FileCopyrightText: Workbench Contributors
+# SPDX-FileContributor: Andy Holmes <andyholmes@gnome.org>
+
+prefix = get_option('prefix')
+datadir = join_paths(prefix, get_option('datadir'))
+includedir = join_paths(prefix, get_option('includedir'))
+libdir = join_paths(prefix, get_option('libdir'))
+
+girdir = join_paths(datadir, 'gir-1.0')
+typelibdir = join_paths(libdir, 'girepository-1.0')
+vapidir = join_paths(datadir, 'vala', 'vapi')
+
+libworkbench_api_version = '0'
+libworkbench_api_name = 'libworkbench-@0@'.format(libworkbench_api_version)
+
+
+#
+# Shared Library
+#
+libworkbench_c_args = [
+  '-Wall',
+  '-Wextra',
+  '-Wfloat-equal',
+  '-Wformat=2',
+  '-Wincompatible-pointer-types',
+  '-Wint-conversion',
+  '-Wint-to-pointer-cast',
+  '-Wmissing-include-dirs',
+  '-Woverflow',
+  '-Wpointer-arith',
+  '-Wpointer-to-int-cast',
+  '-Wredundant-decls',
+  '-Wshadow',
+  '-Wstrict-prototypes',
+  '-Wundef',
+
+  '-Wno-discarded-array-qualifiers',
+  '-Wno-missing-field-initializers',
+  '-Wno-unused-parameter',
+  '-Wno-missing-declarations',
+
+  '-DWORKBENCH_COMPILATION',
+  '-DWORKBENCH_EXPORT=__attribute__((visibility("default"))) extern',
+]
+libworkbench_link_args = []
+libworkbench_deps = [
+  dependency('gio-2.0', version: '>= 2.76.0'),
+  dependency('gtk4', version: '>= 4.10.0'),
+  dependency('gtksourceview-5', version: '>= 5.8.0'),
+]
+
+if get_option('profile') != 'development'
+  libworkbench_c_args += [
+    '-DG_DISABLE_ASSERT',
+    '-DG_DISABLE_CAST_CHECKS',
+  ]
+endif
+
+libworkbench_headers = files([
+  'workbench.h',
+  'workbench-completion-provider.h',
+  'workbench-completion-request.h',
+])
+
+libworkbench_sources = files([
+  'workbench-completion-provider.c',
+  'workbench-completion-request.c',
+])
+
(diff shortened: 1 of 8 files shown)


andyholmes:
So this is a tiny library, which includes a workaround for `GtkSource.CompletionProvider` so it can be easily used with Workbench.

Below is a simple example of how you might use this:

```js
const provider = new Workbench.CompletionProvider();
provider.connect("completion-request", (provider, request) => {
  console.log(`completion-request: ${request.context.get_word()}`);

  /* Pass around the `request` object, calling `request.add()` to append
   * completion proposals, then call `request.state_changed()` with
   * `Workbench.RequestState.CANCELLED` or `Workbench.RequestState.COMPLETE`
   * to finish the request (and async function).
   */
  GLib.idle_add(GLib.PRIORITY_DEFAULT_IDLE, () => {
    request.state_changed(Workbench.RequestState.CANCELLED);
    // request.state_changed(Workbench.RequestState.COMPLETE);

    return GLib.SOURCE_REMOVE;
  });
});
```

Later, we can identify any performance-sensitive code and move it here, when that makes sense.

src/libworkbench/meson.build
```diff
@@ -0,0 +1,116 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# SPDX-FileCopyrightText: Andy Holmes <andrew.g.r.holmes@gmail.com>
```
sonnyp, line 2: FYI Workbench is under GPL 3 but I'm fine with these files under LGPL - looks like they can be very useful for other non-copyleft projects

src/libworkbench/meson.build
```diff
@@ -0,0 +1,116 @@
...
+  '-Wint-conversion',
+  '-Wint-to-pointer-cast',
+  '-Wmissing-include-dirs',
+  '-Woverflow',
+  '-Wpointer-arith',
+  '-Wpointer-to-int-cast',
+  '-Wredundant-decls',
+  '-Wshadow',
+  '-Wstrict-prototypes',
+  '-Wundef',
```
sonnyp, line 36: `-Woof`

src/libworkbench/meson.build
```diff
@@ -0,0 +1,116 @@
...
+libworkbench_link_args = []
+libworkbench_deps = [
+  dependency('gio-2.0', version: '>= 2.76.0'),
+  dependency('gtk4', version: '>= 4.10.0'),
+  dependency('gtksourceview-5', version: '>= 5.8.0'),
+]
+
+if get_option('profile') != 'development'
+  libworkbench_c_args += [
+    '-DG_DISABLE_ASSERT',
```
sonnyp, line 55: interesting

From what I understand Gtk / GLib doesn't do that rigth?

src/libworkbench/meson.build
```diff
@@ -0,0 +1,116 @@
...
+        symbol_prefix: meson.project_name().to_lower(),
+              sources: [libworkbench_headers, libworkbench_sources, libworkbench_enums],
+           extra_args: ['--c-include=workbench.h'],
+             includes: ['Gio-2.0', 'Gtk-4.0', 'GtkSource-5'],
+      install_dir_gir: girdir,
+  install_dir_typelib: typelibdir,
+              install: true,
+)
+
+# Vala API Bindings
```
sonnyp, line 110: thanks :)

src/libworkbench/workbench-completion-provider.c
```diff
@@ -0,0 +1,188 @@
...
+#include "workbench-completion-provider.h"
+#include "workbench-completion-request.h"
+
+
+/**
+ * WorkbenchCompletionProvider:
+ *
+ * A base class for completion providers in Workbench.
+ *
+ * Since: 45.0
```
sonnyp, line 15: GNOME 45 is still 3 months away :disappointed: 

but I guess it's not very important here if we release this under 44.x

src/libworkbench/workbench-completion-provider.c
```diff
@@ -0,0 +1,188 @@
+
```
sonnyp, line 1: did you forget the spdx header in this file?

src/libworkbench/workbench-completion-request.c
```diff
@@ -0,0 +1,450 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+// SPDX-FileCopyrightText: Workbench Contributors
```
sonnyp, line 2:
```suggestion
// SPDX-FileCopyrightText: Andy Holmes <andrew.g.r.holmes@gmail.com>
```

sonnyp (commented):
I will test this with Blueprint ASAP :sparkles:

sonnyp:
Also, please add your name to `developers` in `about.js`.

src/libworkbench/meson.build
andyholmes: oh right, I meant to change those and have them just say `Workbench Contributors`. There's a secondary SPDX key for particular contributors that can be used. GPL-3 is fine for an internal library I think.

src/libworkbench/meson.build
andyholmes: I believe it does based on `debug` or `optimization` meson options, but those don't really apply to Workbench I think.

`g_assert()` failures are hard crashes with line numbers, so they're useful when developing, but as a rule if they fail you made a mistake that shouldn't ever ship.

src/libworkbench/workbench-completion-provider.c
andyholmes: Oh, yeah that's just habit from writing for public libraries. `libworkbench`'s version is actually `0.{meson.project_version()}`, e.g. `0.44.2`.

sonnyp (approved):
Nice! I was able to show completions.

I wonder if it's not worth "upstreaming" one way or the other?

PR #374, opened 2023-06-29
libworkbench: implement GLogWriterFunc and library initialization

andyholmes:
Add `Workbench.init()` to initialize resources and other setup tasks.

Currently, this only implements a `GLogWriterFunc` for silencing
specific messages that Workbench users shouldn't see.

This method has some limitations running on other threads in GJS,
and should be more performant in C anyways.


sonnyp (commented):
Can you add the rational / benefit of this in the PR description?

PR #376, opened 2023-07-01
library: Add Actionbar entry

SoNiC-HeRE:
Kept the entry minimalistic so as to create UI as shown in the documentation:

image

### DIFF
--- a/src/Library/demos/Action Bar/main.blp
+++ b/src/Library/demos/Action Bar/main.blp
@@ -1,29 +1,50 @@
 using Gtk 4.0;
 using Adw 1;
 
-Adw.StatusPage {
+Gtk.Window {
+  default-width: 720;
+  default-height: 360;
   title: _("Action Bar");
-  description: _("A widget designed to present contextual actions.");
 
-  Box {
-    orientation: vertical;
+  Adw.StatusPage {
+    title: "Action Bar";
+    description: _("ActionBar is designed to present contextual actions.");
 
-    ActionBar action_bar {
-      halign: center;
-      revealed: false;
-    }
+    Box {
+      orientation: vertical;
+      spacing: 18;
 
-    Button show_content {
-      halign: center;
-      margin-bottom: 18;
-      margin-top: 24;
-      label: _("Show");
-      styles ["suggested-action", "pill"]
-    }
+      Button show_content {
+        halign: center;
+        label: _("Show Action Bar");
+        styles ["pill", "suggested-action"]
+      }
 
-    LinkButton {
-      label: _("API Reference");
-      uri: _("https://docs.gtk.org/gtk4/class.ActionBar.html");
+      ActionBar action_bar {
+        revealed: false;
+        margin-top: 24;
+
+          [start]
+          Button start_widget {
+            icon-name: "check-round-outline-whole-symbolic";
+          }
+
+          [center]
+          DropDown {
+            model: StringList {
+              strings ["Center Widget","I", "am", "center", "widget"]
+            };
+          }
+
+          [end]
+          Button end_widget {
+            icon-name: "cross-filled-symbolic";
+          }
+      }
+      LinkButton {
+        label: _("API Reference");
+        uri: "https://docs.gtk.org/gtk4/class.ActionBar.html";
+      }
     }
   }
 }
--- a/src/Library/demos/Action Bar/main.blp
+++ b/src/Library/demos/Action Bar/main.blp
@@ -20,9 +20,14 @@ Gtk.Window {
         styles ["pill", "suggested-action"]
       }
 
+      LinkButton {
+        label: _("API Reference");
+        uri: "https://docs.gtk.org/gtk4/class.ActionBar.html";
+      }
+
       ActionBar action_bar {
         revealed: false;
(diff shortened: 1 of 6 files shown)


src/Library/demos/Action Bar/main.js
```diff
@@ -0,0 +1,53 @@
...
+const buttonsBox = new Gtk.Box({
+  orientation: Gtk.Orientation.HORIZONTAL,
+  spacing: 10,
+  homogeneous: true,
+});
+
+buttonsBox.append(prefix_widget);
+buttonsBox.append(suffix_widget);
+
+box.append(buttonsBox);
```
sonnyp, line 39: I don't see why all of this isn't part of the Blueprint

src/Library/demos/Action Bar/main.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "Action Bar",
+  "category": "user_interface",
+  "description": "A widget designed to present contextual actions.",
+  "panels": ["code", "preview"],
```
sonnyp, line 5:
```suggestion
  "panels": ["ui", "preview"],
```

src/Library/demos/Action Bar/main.blp
```diff
@@ -0,0 +1,29 @@
...
+Adw.StatusPage {
+  title: _("Action Bar");
+  description: _("A widget designed to present contextual actions.");
+
+  Box {
+    orientation: vertical;
+
+    ActionBar action_bar {
+      halign: center;
+      revealed: false;
```
sonnyp, line 13: let's show it by default and have a toggle for this

sonnyp (changes requested):
> Kept the entry minimalistic so as to create UI as shown in the documentation:

It doesn't appear to be the case.

And the layout is broken

[Screencast from 2023-07-02 13-26-20.webm](https://github.com/sonnyp/Workbench/assets/19673/dd0102ec-706c-4b40-8357-428f16b13fd3)


From https://docs.gtk.org/gtk4/class.ActionBar.html

> It is expected to be displayed below the content and expand horizontally to fill the area.

I think a better demo would be some kind of fake window (just a box with `card` style class) of 720x360 pixels with the action bar at the bottom

SoNiC-HeRE:
Okay, I'll make the required changes

SoNiC-HeRE:
> > Kept the entry minimalistic so as to create UI as shown in the documentation:
> 
> It doesn't appear to be the case.
> 
> And the layout is broken
> Screencast.from.2023-07-02.13-26-20.webm
> 
> From https://docs.gtk.org/gtk4/class.ActionBar.html
> 
> > It is expected to be displayed below the content and expand horizontally to fill the area.
> 
> I think a better demo would be some kind of fake window (just a box with `card` style class) of 720x360 pixels with the action bar at the bottom
> 
> The fake window could have a status page with a button to toggle reveal on and off

Is this what you meant?

sonnyp:
@SoNiC-HeRE yes but not as a window

If we can avoid it we should at it puts the demo behind a button

sonnyp:
Also I meant 720h and 360w - like a phone

src/Library/demos/Action Bar/main.blp
```diff
@@ -0,0 +1,55 @@
+using Gtk 4.0;
+using Adw 1;
+
+Box {
+  height-request: 72;
+  width-request: 36;
```
sonnyp, line 6: ?

```suggestion
  height-request: 720;
  width-request: 360;
```

src/Library/demos/Action Bar/main.blp
```diff
@@ -0,0 +1,55 @@
...
+  halign: center;
+  margin-top: 48;
+  margin-bottom: 48;
+  orientation: vertical;
+  styles ["card"]
+
+  Adw.StatusPage {
+    title: "Action Bar";
```
sonnyp, line 14: please try to remember

```suggestion

  Adw.StatusPage {
    title: _("Action Bar");
```

src/Library/demos/Action Bar/main.blp
```diff
@@ -0,0 +1,55 @@
...
+    description: _("ActionBar is designed to present contextual actions.");
```
sonnyp, line 15: We shouldn't repeat the title in the description

```suggestion
    description: _("Toolbar for contextual actions.");
```

sonnyp (commented):
Good start.

The action bar should be visible by default so the user doesn't have to click a button to see what it is. Since the bar will be revealed by default, no need for it to be primary action.

> with a button to toggle reveal on and off

The button doesn't toggle on and off. You can make it a "Reveal" ToggleButton.

The chosen icon buttons aren't great. Since this is for contextual actions, we should use icons that are call to actions.

What about `call-start-symbolic` on the left and `padlock2-symbolic` on the right?

For the center a dropdown is fine but it's too long, can you use emojis for the strings ?

sonnyp (approved):
thanks

couple of small fixes https://github.com/sonnyp/Workbench/pull/376/commits/7d1fa76e2f48384d50033dc68e6ea3147e3cb98a

SoNiC-HeRE:
Closes #379

PR #377, opened 2023-07-02
library: add screencast entry

### LINKED ISSUE #342: Add Screencast Demo
https://libportal.org/method.Portal.create_screencast_session.html

halfmexican:
closes #342


src/Library/demos/Screencast/main.js
```diff
@@ -0,0 +1,88 @@
...
+        const source = Gst.ElementFactory.make("pipewiresrc", "source");
+        const queue = Gst.ElementFactory.make("queue", "queue"); // add a queue element
+        const paintable_sink = Gst.ElementFactory.make(
+          "gtk4paintablesink",
+          "paintable_sink",
+        );
+        const glsinkbin = Gst.ElementFactory.make("glsinkbin", "glsinkbin");
+
+        // Set up and Link Pipeline
+        source.set_property("fd", pw_remote); // pw_remote is the file descriptor obtained from libportal
```
bilelmoussaoui, line 62: you should also set the node id but no idea how you would get that with libportal

sonnyp: Thanks, yeah that's also what Georges said, I was going to look into this

sonnyp: @halfmexican you can find the node id from https://libportal.org/method.Session.get_streams.html

sonnyp (changes requested):
It doesn't show the Camera stream anymore :)
But now we need to set the node id on the pipewire source as hinted by Bilal and Georges

https://libportal.org/method.Session.get_streams.html

@halfmexican also please use Gio._Promisify

src/Library/demos/Screencast/main.js
halfmexican: Think I did it

sonnyp (approved):
Good job.

I refactored the JS code to use promisify correctly https://github.com/sonnyp/Workbench/pull/377/commits/f6e2e894e6de6b7a1436d8a7da1b69dde18026d6

sonnyp:
Not merging this just yet as it doesn't work with GNOME 45 for some reason. I need to do more investigation.

PR #386, opened 2023-07-04
library: add Drop Zone Entry

### LINKED ISSUE #368: Drop Zone entry
We should have a demo that demonstrates how to have a drop zone which can be used to drop a file or text or anything from outside the app.

I would suggest making something like this to start with

EDIT: nevermind this design refer to the commet below

image

The zone should highlight or animate when there is a drag over it. 
In GNOME we usually show an inset border of 1px.  See for example with https://apps.gnome.org/app/app.drey.Warp/
There is also a good example here: https://www.dropzone.dev/

Code:

```blp
using Gtk 4.0;
using Adw 1;

Adw.StatusPage {
  title: "Drop Zone";
  description: "Accept data dragged from other applications";
  Box {
    margin-top: 48;
    hexpand: true;
    vexpand: true;
    halign: center;
    valign: center;
    orientation: vertical;

    Box {
      name: "drop";
      width-request: 200;
      height-request: 200;
      Label {
        name: "label";
        label: _("<b>Drop</b>");
        use-markup: true;
        hexpand: true;
      }
    }
  }
}

```

```css
#drop {
  border: dashed 6px @dark_3;
  border-radius: 25%;
  background-color: white;
  
}

#label {
  color: @dark_3;
}
```

halfmexican:
drafting this 

closes #368

### DIFF
--- a/src/Library/demos/Drop Zone/main.blp
+++ b/src/Library/demos/Drop Zone/main.blp
@@ -11,13 +11,12 @@ Adw.StatusPage {
     Adw.Clamp {
       maximum-size: 400;
 
-
-      Adw.Bin container {
+      Adw.Bin Bin {
         styles ["card"]
         height-request: 150;
-
       }
     }
+
     LinkButton {
       label: "API Reference";
       uri: "https://docs.gtk.org/gtk4/drag-and-drop.html";
--- a/src/Library/demos/Drop Zone/main.blp
+++ b/src/Library/demos/Drop Zone/main.blp
@@ -11,12 +11,17 @@ Adw.StatusPage {
     Adw.Clamp {
       maximum-size: 500;
 
-      Adw.Bin bin {
+      Box {
+        orientation: vertical;
         styles ["card"]
-        height-request: 150;
 
-        Label {
-          label: _("Drop files or text here");
+        Adw.Bin bin {
+
+          height-request: 150;
+
+          Label {
+            label: _("Drop files or text here");
+          }
         }
       }
     }
--- a/src/Library/demos/Drop Zone/main.blp
+++ b/src/Library/demos/Drop Zone/main.blp
@@ -11,7 +11,7 @@ Adw.StatusPage {
     Adw.Clamp {
       maximum-size: 400;
 
-      Adw.Bin Bin {
+      Adw.Bin bin {
         styles ["card"]
         height-request: 150;
       }
--- a/src/Library/demos/Drop Zone/main.blp
+++ b/src/Library/demos/Drop Zone/main.blp
@@ -9,7 +9,7 @@ Adw.StatusPage {
     orientation: vertical;
 
     Adw.Clamp {
-      maximum-size: 400;
+      maximum-size: 500;
 
       Adw.Bin bin {
         styles ["card"]
--- a/src/Library/demos/Drop Zone/main.blp
+++ b/src/Library/demos/Drop Zone/main.blp
@@ -21,10 +21,19 @@ Adw.StatusPage {
       }
     }
 
-    LinkButton {
-      label: "API Reference";
-      uri: "https://docs.gtk.org/gtk4/drag-and-drop.html";
+    Box {
       margin-top: 24;
+      orientation: vertical;
+
+      LinkButton {
+        label: "Drag-and-Drop in GTK";
+        uri: "https://docs.gtk.org/gtk4/drag-and-drop.html#drop-targets";
+      }
+
+      LinkButton {
+        label: "DropTarget API Reference";
+        uri: "https://docs.gtk.org/gtk4/class.DropTarget.html";
+      }
     }
   }
 }
--- a/src/Library/demos/Drop Zone/main.blp
+++ b/src/Library/demos/Drop Zone/main.blp
@@ -14,6 +14,10 @@ Adw.StatusPage {
       Adw.Bin bin {
         styles ["card"]
         height-request: 150;
+
+        Label {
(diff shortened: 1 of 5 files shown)


src/Library/demos/Drop Zone/main.blp
```diff
@@ -0,0 +1,33 @@
...
+  title: _("Drop Zone");
+  description: _("A versatile drop area for external content");
+
+  Box {
+    orientation: vertical;
+
+    Adw.Clamp {
+      maximum-size: 500;
+
+      Adw.Bin bin {
```
sonnyp, line 14: `Bin` isn't really meant to be used directly but rather for subclassing

Would using `Box` complicate the code significantly?

src/Library/demos/Drop Zone/main.js
```diff
@@ -0,0 +1,124 @@
...
+  if (!(value instanceof Gio.File)) return false;
+
+  const file_info = value.query_info("standard::content-type", 0, null);
+  const content_type = file_info.get_content_type();
+
+  if (content_type.startsWith("image/")) {
+    try {
+      bin.child = createImageWidget(value);
+    } catch (error) {
+      console.log(`Unable to load image: ${error}`);
```
sonnyp, line 36: always use `logError` in GJS (at least for now, until we fix `console` to deal with it properly)

src/Library/demos/Drop Zone/main.js
```diff
@@ -0,0 +1,124 @@
...
+  } else {
+    try {
+      bin.child = createFileWidget(value);
+    } catch (error) {
+      console.log(`Unable to create file widget: ${error}`);
+    }
+  }
+
+  bin.get_style_context().remove_class("overlay-drag-area");
+});
```
sonnyp, line 53: you can simplify this by moving the content of the signal handler to a new function `onDrop` 

and moving the try/catch upwards:

```
file_drop_target.connect("drop", (self, value, x, y) => {
  try {
    onDrop(value);
  } catch (err) {
    logError(err);
  }
});
```

src/Library/demos/Drop Zone/main.js
```diff
@@ -0,0 +1,124 @@
...
+  widget.append(icon);
+  icon.icon_size = Gtk.IconSize.LARGE;
+
+  const file_name = new Gtk.Label({ label: file.get_basename() });
+  widget.append(file_name);
+
+  return widget;
+}
+
+function createBoxWidget() {
```
sonnyp, line 95: FYI, you could move this to Blueprint but it's fine here too

src/Library/demos/Drop Zone/main.js
```diff
@@ -0,0 +1,124 @@
...
+const string_drop_target = Gtk.DropTarget.new(
+  GObject.TYPE_STRING,
+  Gdk.DragAction.COPY,
+);
+
+bin.add_controller(string_drop_target);
+
+string_drop_target.connect("drop", (self, value, x, y) => {
+  bin.child = createTextWidget(value);
+  bin.get_style_context().remove_class("overlay-drag-area");
```
sonnyp, line 19: `get_style_context` is deprecated.

Use `add_css_class` and `remove_css_class` Widget methods directly

src/Library/demos/Drop Zone/main.js
```diff
@@ -0,0 +1,124 @@
...
+
+function createVideoWidget(file) {
+  const widget = createBoxWidget();
+
+  const video = new Gtk.Video({ file: file });
+  widget.append(video);
+  return widget;
+}
+
+function createFileWidget(file) {
```
sonnyp, line 81: maybe rename the functions to `createFooPreview` ?

src/Library/demos/Drop Zone/main.blp
```diff
@@ -0,0 +1,33 @@
...
+          label: _("Drop files or text here");
+        }
+      }
+    }
+
+    LinkButton {
+      label: "API Reference";
+      uri: "https://docs.gtk.org/gtk4/drag-and-drop.html";
+      margin-top: 24;
+    }
```
sonnyp, line 28: this is not an API reference, here is a suggestion

```
    Box {
      margin-top: 24;
      orientation: vertical;

      LinkButton {
        label: "Drag-and-Drop in GTK";
        uri: "https://docs.gtk.org/gtk4/drag-and-drop.html#drop-targets";
      }
      
      LinkButton {
        label: "DropTarget API Reference";
        uri: "https://docs.gtk.org/gtk4/class.DropTarget.html";
      }
    }
```

sonnyp (changes requested):
Excellent!

Just a couple more things for perfection:

Can you support dragging data too? You can test with dropping an image displayed in your web browser. Dragging text from the browser also does not work but I suspect it's the same problem.

If you look at the paper clip screenshot, you'll notice there is a margin around the blue overlay:

Screenshot from 2023-07-08 23-13-52

src/Library/demos/Drop Zone/main.blp
halfmexican: It's much easier to use a bin as I can just set the child parameter, with a box i would have to handle the removing a child from the box whenever another item is dropped. The document page said that "it is useful for deriving subclasses" does that mean we have to?

sonnyp: Let's use it then :+1:

sonnyp (approved):
Great work, I created a follow up https://github.com/sonnyp/Workbench/issues/440

PR #387, opened 2023-07-05
library: Adds File Dialog Entry

### LINKED ISSUE #382: File Dialog library entry
check the avatar library entry for example

SoNiC-HeRE:
Closes #382 

- Kept the entry minimalistic
- Looking for an early review

### DIFF
--- a/src/Library/demos/File Dialog/main.blp
+++ b/src/Library/demos/File Dialog/main.blp
@@ -1,0 +1,23 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "File Dialog";
+  description: _("It collects the arguments that are needed to present a file chooser dialog to the user");
+
+  Box {
+    orientation: vertical;
+    spacing: 18;
+    vexpand:true;
+
+    Button button {
+      halign: center;
+      label: _("Select a file");
+      styles ["pill", "suggested-action"]
+    }
+    LinkButton {
+      label: _("API Reference");
+      uri: "https://docs.gtk.org/gtk4/class.FileDialog.html";
+    }
+  }
+}
--- a/src/Library/demos/File Dialog/main.js
+++ b/src/Library/demos/File Dialog/main.js
@@ -1,0 +1,26 @@
+import Gtk from "gi://Gtk";
+import Gdk from "gi://Gdk";
+import Gio from "gi://Gio";
+
+Gio._promisify(Gtk.FileDialog.prototype, "open", "open_finish");
+
+const button = workbench.builder.get_object("button");
+
+const dialog = new Gtk.FileDialog({
+  title: "Select an Avatar",
+  modal: true,
+});
+
+button.connect("clicked", () => {
+  onClicked().catch(logError);
+});
+
+async function onClicked() {
+  const file = await dialog.open(workbench.window, null);
+  const info = file.query_info(
+    "standard::*",
+    Gio.FileQueryInfoFlags.NOFOLLOW_SYMLINKS,
+    null,
+  );
+  console.log(`Selected File: ${info.get_name()}`);
+}
--- a/src/Library/demos/File Dialog/main.json
+++ b/src/Library/demos/File Dialog/main.json
@@ -1,0 +1,7 @@
+{
+  "name": "File Dialog",
+  "category": "user_interface",
+  "description": "It collects the arguments that are needed to present a file chooser dialog to the user",
+  "panels": ["code", "preview"],
+  "autorun": true
+}


sonnyp (commented):
I think we should show the different "modes"

open file (with filder)

open  multiple files

select folder

select multiple folders

save file

Maybe one button each?

SoNiC-HeRE:
- Improved UI
- Added separate buttons for different modes
- Looking for an early review

src/Library/demos/File Dialog/main.blp
```diff
@@ -0,0 +1,35 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "File Dialog";
```
sonnyp, line 5:
```suggestion
  title: _("File Dialog");
```

@SoNiC-HeRE :warning:

src/Library/demos/File Dialog/main.blp
```diff
@@ -0,0 +1,35 @@
...
+  description: _("It collects the arguments that are needed to present a file chooser dialog to the user");
```
sonnyp, line 6: This is not in line with the other descriptions. Please try to come up with description that are more inline with the other ones.

```suggestion
  description: _("Present a dialog to open and save files");
```

src/Library/demos/File Dialog/main.js
```diff
@@ -0,0 +1,125 @@
...
+});
+
+button.connect("clicked", () => {
+  switch (selected_mode) {
+    case "Select File":
+      open_file().catch(logError);
+      break;
+    case "Select File (Folder pre-defined)":
+      open_file_folder().catch(logError);
+      break;
```
sonnyp, line 54: I don't think we need both. "Select File" with a pre-defined folder is a good example. If for some reason the user doesn't want a pre-defined folder, it is easy to adjust the code.

sonnyp (changes requested):
Why did you choose not to follow my guidance? It's not necessarily a problem but you should explain the reason that leads you to go a different route.  

In this case:

* why a dropdown instead of a button for each?
*  why 2 examples for Select Files ?

https://github.com/sonnyp/Workbench/pull/387#pullrequestreview-1514773036

SoNiC-HeRE:
> Why did you choose not to follow my guidance? It's not necessarily a problem but you should explain the reason that leads you to go a different route.
> 
> In this case:
> 
>     * why a dropdown instead of a button for each?
> 
>     * why 2 examples for Select Files ?
> 
> 
> [#387 (review)](https://github.com/sonnyp/Workbench/pull/387#pullrequestreview-1514773036)

- I first implemented the layout using different buttons but it makes the UI crowded so i thought of going with Dropdown to minimise UI and give a more cleaner look;if you want to i could rechange the layout to use only buttons instead of Dropdown selection
- I created 2 examples in reference to https://github.com/sonnyp/Workbench/pull/387#pullrequestreview-1514773036
```open file (with filder)``` ;  and i thought it would be great to have 2 examples with different usage;

If any changes are required let me know I'll update it

File: src/Library/demos/File Dialog/main.js [hunk 4]
```diff
@@ -0,0 +1,125 @@
...
+const button = workbench.builder.get_object("button");
+const select_mode = workbench.builder.get_object("select_mode");
+
+let selected_mode = select_mode.selected_item.get_string();
+select_mode.connect("notify::selected-item", () => {
+  selected_mode = select_mode.selected_item.get_string();
+  console.log(`Mode: ${selected_mode}`);
+});
+
+const dialog_for_file = new Gtk.FileDialog({
```
sonnyp, line 32: a new dialog should be created everytime, not re-used

src/Library/demos/File Dialog/main.js
```diff
@@ -0,0 +1,125 @@
...
+      break;
+    case "Save File":
+      save_file().catch(logError);
+      break;
+    default:
+      console.log("Invalid mode selected.");
+  }
+});
+
+async function open_file() {
```
sonnyp, line 72: snake_case for variables
camelCase for functions

added to https://github.com/sonnyp/Workbench/wiki/Contribution-Style-Guide#javascript

sonnyp:
Thanks for the explanations, please make sure to share them proactively next time.

> I first implemented the layout using different buttons but it makes the UI crowded (5-6 buttons) so i thought of going with Dropdown to minimise UI and give a more cleaner look;if you want i could rechange the layout to use only buttons instead of Dropdown selection

I think it would have been useful to share a screenshot and ask for a second opinion, even if just to see for myself. I think the dropdown is fine :+1: 

>  i thought it would be great to have 2 examples with different usage;

Given that the only difference is a single property that can be commented out easily, it's not useful. Please use the example with the default folder.

src/Library/demos/File Dialog/main.js
```diff
@@ -0,0 +1,125 @@
...
+  );
+  console.log(`Selected File: ${info.get_name()}`);
+}
+
+async function open_file_folder() {
+  const folder = Gio.File.new_for_path(pkg.pkgdatadir).resolve_relative_path(
+    "Library/demos/File Dialog/main.blp",
+  );
+  const parent_folder = folder.get_parent();
+  dialog_for_file.initial_folder = parent_folder;
```
sonnyp, line 87: let's use the standard download directory instead, you can get it with GLib `get_user_special_dir`

SoNiC-HeRE:
I'm sorry for not seeking a second opinion before moving on with my user interface; I'll take the feedback more carefully and put them into practise. I'll make the necessary changes to the code for the time being.

sonnyp:
One thing to not is that `GtkFileDialog` doesn't appear to support saving a directory :/

It's available in libportal under https://libportal.org/method.Portal.save_files.html

Not sure what to do about that, maybe we should open a GTK bug about supporting the xdg `SaveFiles` method

sonnyp:
Rectification, `save_files` doesn't save a directory but instead lets the user choose a directory to save multiple files in.

PR #391, opened 2023-07-09
Use diagnostics from gtkcsslanguageserver

### LINKED ISSUE #370: GTK CSS language server
A language server is a standard program that can be used by any editor (which support the Language Server Protocol) to add support for a language or feature.

https://microsoft.github.io/language-server-protocol/

With a GTK CSS Language Server, we would be able to

* show inline CSS errors (we already do but without a Language Server)
* auto-complete / suggest CSS properties
* auto-complete / suggest CSS values

GTK CSS is different from HTML CSS, we cannot use the same language server. This could also be used in GNOME Builder.

A GTK CSS Language Server should be quite straightforward to build, errors can be reported by parsing the CSS with `parsing-error` from https://docs.gtk.org/gtk4/class.CssProvider.html

I don't know if the list of CSS properties is exposed by GTK. But it doesn't change often, so it's fine to build a list. 

Prior work:
* https://github.com/microsoft/vscode-css-languageservice
* https://github.com/Microsoft/vscode/tree/main/extensions/css-language-features/server

JCWasmx86:
Fixes #370

Integrates this language server for GTK CSS flavor: https://github.com/JCWasmx86/GTKCssLanguageServer

Currently the diagnostics are complete, hovering are stubs (As I currently don't have the time to copy-paste the description from MDN), auto-completion is barebones.

This code provided here is enough for showing diagnostics.

### DIFF
--- a/po/POTFILES
+++ b/po/POTFILES
@@ -58,3 +58,4 @@ src/IconLibrary/main.blp
 src/IconLibrary/main.js
 src/DocumentationViewer.blp
 src/DocumentationViewer.js
+src/langs/css/css.js
--- a/re.sonny.Workbench.Devel.json
+++ b/re.sonny.Workbench.Devel.json
@@ -185,6 +185,33 @@
       ]
     },
     {
+      "name" : "jsonrpc-glib",
+      "config-opts" : [
+        "--buildtype=debugoptimized",
+        "-Denable_tests=false"
+      ],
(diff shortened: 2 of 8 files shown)


sonnyp (approved):
Awesome thanks!

Merging this to get the diagnostics. Will do the completions / hovering in follow-ups.

Issue #385, opened 2023-07-02
Label library entry


sonnyp:
Justification, x-align etc

 https://docs.gtk.org/gtk4/class.Label.html

PR #393, opened 2023-07-12
library: add Label entry

halfmexican:
closes: #385

### DIFF
--- a/src/Library/demos/Label/main.js
+++ b/src/Library/demos/Label/main.js
@@ -25,10 +25,10 @@ const style_classes = [
 ];
 
 const short_label =
-  "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore disputandum putant.";
+  "<b>Lorem ipsum</b> dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore disputandum putant.";
 
 const long_label =
-  "     Lorem ipsum dolor sit amet, consectetur adipiscing elit,\n\
+  "     <b>Lorem ipsum</b> dolor sit amet, consectetur adipiscing elit,\n\
   sed do eiusmod tempor incididunt ut labore et dolore magnam aliquam quaerat voluptatem.\n\
   Ut enim mortis metu omnis quietae vitae status perturbatur,\n\
   et ut succumbere doloribus eosque humili animo inbecilloque ferre miserum est,\n\
--- a/src/Library/demos/Label/main.js
+++ b/src/Library/demos/Label/main.js
@@ -1,0 +1,4 @@
+import Gtk from "gi://Gtk";
+import Adw from "gi://Adw";
+
+const justification_row = workbench.builder.get_object("justification_row");
--- a/src/Library/demos/Label/main.js
+++ b/src/Library/demos/Label/main.js
@@ -1,4 +1,53 @@
 import Gtk from "gi://Gtk";
 import Adw from "gi://Adw";
 
+const label = workbench.builder.get_object("label");
 const justification_row = workbench.builder.get_object("justification_row");
+const style_row = workbench.builder.get_object("style_row");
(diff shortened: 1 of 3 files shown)


halfmexican:
marking this ready to get more feedback on what I should add to this demo

src/Library/demos/Label/main.js
```diff
@@ -0,0 +1,53 @@
...
+  "  Lorem ipsum dolor sit amet, consectetur adipiscing elit,\n\
+  sed do eiusmod tempor incididunt ut labore et dolore magnam aliquam quaerat voluptatem.\n\
+  Ut enim mortis metu omnis quietae vitae status perturbatur,\n\
+  et ut succumbere doloribus eosque humili animo inbecilloque ferre miserum est,\n\
+  ob eamque debilitatem animi multi parentes, multi amicos, non nulli patriam,\n\
+  plerique autem se ipsos penitus perdiderunt, sic robustus animus et excelsus omni.";
+
+xalign_spin_button.connect("notify::value", () => {
+  label.xalign = xalign_spin_button.get_value();
+});
```
andyholmes, line 36: I think this one could just be a bound property.

For me, it doesn't seem like the property is correct at startup for some reason, even though you've set it correctly in the UI file. If I bind it, it gets set to `0.0`, but at least the widget is up to date.

andyholmes (commented):
Looking good so far, I like the nice boxed list of properties.

src/Library/demos/Label/main.js
```diff
@@ -0,0 +1,53 @@
...
+  "success",
+  "warning",
+  "error",
+  "heading",
+  "body",
+  "caption-heading",
+  "caption",
+];
+
+label.label =
```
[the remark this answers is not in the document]
halfmexican: I think that is a great idea. We can have a switch to switch between labels

src/Library/demos/Label/main.js
```diff
@@ -0,0 +1,63 @@
...
+  "warning",
+  "error",
+  "heading",
+  "body",
+  "caption-heading",
+  "caption",
+];
+
+const short_label =
+  "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore disputandum putant.";
```
andyholmes, line 28:
```suggestion
  "<b>Lorem ipsum</b> dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore disputandum putant.";
```

Looks good, but needs a bit of markup to demonstrate the feature.

src/Library/demos/Label/main.js
```diff
@@ -0,0 +1,63 @@
...
+
+const long_label =
+  "     Lorem ipsum dolor sit amet, consectetur adipiscing elit,\n\
```
andyholmes, line 31:
```suggestion
  "     <b>Lorem ipsum</b> dolor sit amet, consectetur adipiscing elit,\n\
```

src/Library/demos/Label/main.js
```diff
@@ -0,0 +1,63 @@
...
+  sed do eiusmod tempor incididunt ut labore et dolore magnam aliquam quaerat voluptatem.\n\
+  Ut enim mortis metu omnis quietae vitae status perturbatur,\n\
+  et ut succumbere doloribus eosque humili animo inbecilloque ferre miserum est,\n\
+  ob eamque debilitatem animi multi parentes, multi amicos, non nulli patriam,\n\
+  plerique autem se ipsos penitus perdiderunt, sic robustus animus et excelsus omni.";
+
+label.label = short_label;
```
andyholmes, line 38: Another way to do this might be:

```
const short_label = long_label.replace(/\n/g, "");
```

But then the demo gets *really* wide, so maybe not. Probably okay like this :)

andyholmes (changes requested):
Looks really good and demonstrates all the properties most people will ever need.

Just needs a bit of markup in the label text to demo that feature (pick whatever you want), then its good to go.

andyholmes (approved):

PR #394, opened 2023-07-12
DocumentationViewer: Add missing docs from gtk-doc

### LINKED ISSUE #373: Investigate missing docs
Some docs are missing from the documentation viewer.

The ones I could think of: `GObject`, `Gio`, `Webkit`

As far as I can tell we should show all the docs that

`flatpak run --runtime-version=44 --nofilesystem=host org.gnome.Devhelp` shows.

AkshayWarrier:
Closes #373

### DIFF
--- a/src/DocumentationViewer.js
+++ b/src/DocumentationViewer.js
@@ -127,26 +127,41 @@ async function getDocs(base_path) {
   const dirs = await list(base_path);
   for (const dir of dirs) {
     try {
-      const index = await readDocIndex(base_path, dir);
-      const namespace = `${index["meta"]["ns"]}-${index["meta"]["version"]}`;
+      const results = await Promise.allSettled([
+        readDocIndexJSON(base_path, dir),
+        readDocIndexHTML(base_path, dir),
+      ]);
+
+      let namespace = results.filter((result) => result.status === "fulfilled");
+      if (namespace.length) namespace = namespace[0].value;
+      else continue;
+
       const uri = base_path.get_child(dir).get_child("index.html").get_uri();
       docs.push({
         title: namespace,
         uri: uri,
       });
     } catch (e) {
-      // Ignore the error if the dir does not contain index.json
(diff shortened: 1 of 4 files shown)


src/DocumentationViewer.js
```diff
@@ -87,6 +87,10 @@ export default function DocumentationViewer({ application }) {
   listbox.set_sort_func(sort);
   listbox.invalidate_sort();
   listbox.set_filter_func(filter);
+  listbox.connect("row-selected", (self, row) => {
+    webview.load_uri(row.uri);
+  });
```
sonnyp, line 92: this didn't need to be executed for each doc

src/DocumentationViewer.js
```diff
@@ -97,14 +101,11 @@ export default function DocumentationViewer({ application }) {
...
-      listbox.connect("row-selected", (self, row) => {
-        webview.load_uri(row.uri);
-      });
     })
     .catch(logError);
 
-  function sort(row1, row2) {
-    return row1.title > row2.title;
+  function sort(a, b) {
+    return a.title.localeCompare(b.title);
```
sonnyp, line 108: sort case insensitive

src/DocumentationViewer.js
```diff
@@ -125,25 +126,22 @@ export default function DocumentationViewer({ application }) {
...
   for (const dir of dirs) {
-    try {
-      const results = await Promise.allSettled([
-        readDocIndexJSON(base_path, dir),
-        readDocIndexHTML(base_path, dir),
-      ]);
-
-      let namespace = results.filter((result) => result.status === "fulfilled");
-      if (namespace.length) namespace = namespace[0].value;
-      else continue;
```
sonnyp: confusing :)

sonnyp (approved):
neat, I made some improvements please have a look and let me know if something isn'tt clear https://github.com/sonnyp/Workbench/pull/394/commits/aaac9c510b2fb1231ce081ca75050cdab5160ed1

src/DocumentationViewer.js
AkshayWarrier: Good call

PR #395, opened 2023-07-13
library: Adds ListView Entry

### LINKED ISSUE #365: ListView entry
ListView is really useful for very large list of items. We should have an entry.

SoNiC-HeRE:
Closes #365


src/Library/demos/List View/main.js
```diff
@@ -0,0 +1,75 @@
...
+  listItem.set_child(listRow);
+});
+factory_for_list_view.connect("bind", (factory, listItem) => {
+  const listRow = listItem.get_child();
+  listRow.title = "Demo";
+});
+
+const factory_for_grid_view = new Gtk.SignalListItemFactory();
+factory_for_grid_view.connect("setup", (factory, listItem) => {
+  const listRow = new Adw.ActionRow();
```
andyholmes, line 36: One weird thing I forgot to mention, using `Gtk.ListBoxRow` doesn't work well with listmodel views. Instead we can just use a `Gtk.Box` with `Gtk.Label` inside.

src/Library/demos/List View/main.js
```diff
@@ -0,0 +1,75 @@
...
+  listItem.set_child(listRow);
+});
+factory_for_grid_view.connect("bind", (factory, listItem) => {
+  const listRow = listItem.get_child();
```
andyholmes, line 40: This is one of the confusing bits of these widgets. There are two properties of a `Gtk.ListItem`: `child` (the widget you make) and `item` which is the item from the `Gio.ListModel` (`Gtk.StringList` in our case).

So you want to do something like:

```js
const listRow = listItem.get_child();
const modelItem = listItem.get_item();

listRow.title = modelItem.string;
```

Of course, we'll be using a `Gtk.Box` with a `Gtk.Label` so it would be more like `listRow.child.label = modelItem.string;`, but you get the idea.

src/Library/demos/List View/main.js
```diff
@@ -0,0 +1,75 @@
...
+selection_model.model.connect(
+  "items-changed",
+  (list, position, removed, added) => {
+    console.log(
+      `position: ${position}, Item removed? ${Boolean(
+        removed,
+      )}, Item added? ${Boolean(added)}`,
+    );
+  },
+);
```
andyholmes, line 53: We should keep the original connection to the `Gtk.StringList`, but also have this one to make it obvious the model is inside the selection model.

More important here is to connect to `Gtk.SelectionModel::selection-changed`. I think just printing the title of the item(s) is fine for the callback.

src/Library/demos/List View/main.js
```diff
@@ -0,0 +1,75 @@
...
+// Controller
+add.connect("clicked", () => {
+  const new_item = `New Item ${item}`;
+  selection_model.model.append(new_item);
+  item++;
+});
+
+remove.connect("clicked", () => {
+  selection_model.model.remove(2);
+});
```
andyholmes, line 69: Here we should get the current selection from the `Gtk.SelectionModel`.

andyholmes (commented):
This is a good start. This is a complicated pattern, and I think you're starting to understand it well.

One of these (listview or gridview) we should do a very simple UI-built example, meaning a `Gtk.BuilderListItemFactory`, so have a look into the documentation for that: https://jwestman.pages.gitlab.gnome.org/blueprint-compiler/reference/extensions.html#gtk-builderlistitemfactory-templates

That's maybe a bit more confusing, so do ask for help if you get stuck :slightly_smiling_face:

SoNiC-HeRE:
I have implemented the following changes:
- Implemented UI Build for ListView
- Implemented code build for GridView
- Setup Add Function 
- Setup Remove Function for selected item
- Change the widget for Grid View from ```Adw.ActionRow ``` to ```Gtk.Box``` with ```Gtk.Label``` as child
- Implemented logic for ```selection-changed``` signal

SoNiC-HeRE:
- Currently, I've kept the logic to allow different selections from list and grid view and remove respectively, but I could change it to make it remove from a fixed position from both views. 
- Also, I don't think the add/remove buttons are required for the edit page separately; what if I add them to both pages instead?
-  Furthermore, I believe big UI modifications are possible, and the demo's title should be: ListView/GridView (that's what I was thinking about), but I wanted to ask before proceeding.

Please let me know what changes i should make regarding the UI/Code.

andyholmes:
Just to help me out reviewing, please remember to mark conversations as resolved if you've addressed them :slightly_smiling_face: 

> * Currently, I've kept the logic to allow different selections from list and grid view and remove respectively, but I could change it to make it remove from a fixed position from both views.

Yes, I agree that's the better way to demonstrate this. I see this as a continuation from the List Model demo, where we are applying the model-view-controller pattern to a more complex/powerful widget.

So let's use the same `Gtk.SelectionModel` and `Gtk.StringList` for ListView & GridView. This way when the user/developer switches the same item(s) will be selected. Sort of like the file chooser list/icon view option.

> * I've commented out the code for position updation for stringlist and stack signal handler as i found it way too confusing for user since we are logging multiple properties already

That's a good point, and really the "selection" changes is what we want to show for that model.


> * Also, I don't think the add/remove buttons are required for the edit page separately; what if I add them to both pages instead?

I agree, the separate "Edit" page was good for the List Model demo. Here, let's:

* remove the "Edit" page
* pull the search/edit entry out of the `Gtk.Stack`
* put the search/edit entry between the switch and the stack

This way we have a "Controller" that changes the same `Gtk.StringList` ("Model"), and two different "View" widgets presenting the same data.

> * Furthermore, I believe big UI modifications are possible, and the demo's title should be: ListView/GridView (that's what I was thinking about), but I wanted to ask before proceeding.

I would tend to agree, but I think @sonnyp has a better vision of how to present/organize demos. The GTK documentation puts these under ["List Widget Overview"](https://docs.gtk.org/gtk4/section-list-widget.html), what do you think Sonny?

> Please let me know what changes i should make regarding the UI/Code.

I will add comments as a review for for this :slightly_smiling_face:

andyholmes (changes requested):
Looking good, I'm glad to see you got the factory working in the UI file without trouble.

Once the UI has been updated, let's revisit the code once more.

I don't think this will take many more iterations, so start looking at `Gtk.ColumnView` a bit. That one is more complicated, so we may want to add more to *this* demo if we can. Then when we get to the ColumnView demo it isn't overwhelming.

src/Library/demos/List View/main.blp
```diff
@@ -0,0 +1,105 @@
...
+                  template ListItem {
+                    child: Label {
+                      label: bind template.item as <StringObject>.string;
+                    };
+                  }
+                };
+                model: SingleSelection model_for_listview {
+                  model: StringList  stringlist_for_list_view {
+                    strings ["Default Item 1", "Default Item2", "Default Item 3"]
+                  };
```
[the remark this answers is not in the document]
SoNiC-HeRE: I created same model for populating both list and grid view in the code itself not in the UI

src/Library/demos/List View/main.blp
```diff
@@ -0,0 +1,105 @@
...
+          child:
+            Box {
+              ListView list_view {
+                hexpand: true;
+                valign: start;
+                enable-rubberband: true;
+                factory: BuilderListItemFactory {
+                  template ListItem {
+                    child: Label {
+                      label: bind template.item as <StringObject>.string;
```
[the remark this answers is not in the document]
andyholmes: I'm sorry, I totally forgot my own advice here! I'm not sure why `Gtk.ListView` throws criticals when you use list box rows, but I told you to switch this to box & label and then said to switch it back :see_no_evil:

I guess something like this is pretty close to `Adw.ActionRow`:

```js
template ListItem {
  child: Gtk.Box {
    Gtk.Label {
      label: bind template.item as <StringObject>.string;
      height-request: 50;
      margin-start: 12;
      margin-end: 12;
    }
  };
};
```

What do you think?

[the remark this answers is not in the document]
SoNiC-HeRE: Sure, I think that'd be better

src/Library/demos/List View/main.js
```diff
@@ -0,0 +1,92 @@
...
+  console.log(
+    `Selected item from ListView: ${model.model.get_string(
+      list_view_selected_item,
+    )}`,
+  );
+  console.log(
+    `Selected item from GridView: ${model.model.get_string(
+      grid_view_selected_item,
+    )}`,
+  );
```
andyholmes, line 68: Since we're operating on a single list model, let's trim this to one log message and try to make that more obvious.

```suggestion
  const selected_item = model.get_selected();
  console.log(
    `Model item selected from view: ${model.model.get_string(
      selected_item,
    )}`,
  );
```

src/Library/demos/List View/main.js
```diff
@@ -0,0 +1,92 @@
+import Gio from "gi://Gio";
+import Gdk from "gi://Gdk";
+import GObject from "gi://GObject";
```
andyholmes, line 3:
```suggestion
```

We aren't using these here, I guess, so we can drop them.

src/Library/demos/List View/main.blp
```diff
@@ -0,0 +1,88 @@
...
+
+        Gtk.StackPage {
+          name: "gridview";
+          title: _("Grid View");
+          child:  Frame {
+            valign: start;
+            GridView grid_view {
+              orientation: horizontal;
+              max-columns: 8;
+              enable-rubberband: true;
```
andyholmes, line 69: Not sure why this doesn't work, but let's leave it for now and figure that out later.

andyholmes (changes requested):
This is shaping up quite well, and I'm happy to see it's staying simple. I'm concerned the `Gtk.ColumnView` demo is going to be big and complex, but I think we will be able to keep it simpler by shifting some things into this demo.

Great work so far!

SoNiC-HeRE:
I've applied the suggested changes and also added the code for whenever a change is made in the view the respective position with property whether added or removed is logged; thought it would be useful ;or shall i keep it as it was , wdyt?

Also what properties of `Gtk.ColumnView` do you want to demonstrate here?

src/Library/demos/List View Widgets/main.blp
```diff
@@ -0,0 +1,106 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "List View Widgets";
```
andyholmes, line 5:
```suggestion
  title: _("List View Widgets");
```

src/Library/demos/List View Widgets/main.blp
```diff
@@ -0,0 +1,106 @@
...
+
+
+
+
+
+
+
+
+
+
```
andyholmes, line 106:
```suggestion
```

andyholmes (changes requested):
Looks good to go. Just add that one translatable, remove the whitespace and I think it's ready to merge.

SoNiC-HeRE:
@andyholmes I've applied the changes

andyholmes (approved):

PR #397, opened 2023-07-14
library: Add Actions entry

### LINKED ISSUE #378: Add Actions Library entry

AkshayWarrier:
Closes #378

### DIFF
--- a/src/Library/demos/Actions/main.blp
+++ b/src/Library/demos/Actions/main.blp
@@ -1,0 +1,233 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage demo {
+  title: "Actions";
+  description: _("A high-level interface used to describe a piece of functionality");
+
+  Adw.Clamp {
+    Box {
+      orientation: vertical;
+
+      Box links{
+        halign: center;
+
+        LinkButton {
+          label: "GJS Guide";
+          uri: "https://gjs.guide/guides/gio/actions-and-menus.html#stateful-actions";
+        }
+
+        LinkButton {
+          label: "GTK Documentation";
+          uri: "https://docs.gtk.org/gtk4/actions.html";
+        }
+      }
+
+      Box actions{
+        orientation: vertical;
+        homogeneous: true;
+        spacing: 24;
+
+        Box {
+          orientation: vertical;
+          spacing: 6;
+
+          Label {
+            halign: start;
+            label: _("Activatable Actions");
+
+            styles ["title-4"]
+          }
+
+          Label {
+            halign: start;
+            label: _("Stateless actions that optionally take parameters");
+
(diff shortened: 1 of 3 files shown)


src/Library/demos/Actions/main.blp
```diff
@@ -0,0 +1,234 @@
...
+    label: _("Human Interface Guidelines");
+    action: "demo.open-bookmarks";
+    target: _("Human Interface Guidelines");
+  }
+
+  item {
+    label: _("GNOME Javascript");
+    action: "demo.open-bookmarks";
+    target: _("GNOME Javascript");
+  }
```
andyholmes, line 233: These don't all have to be `app.open_uri`, but I think it's a good opportunity to demo `app` scoped actions. Maybe one could be `app.about`, and another the `demo.open-bookmarks` action?

```suggestion
  item {
    label: _("Developer Documentation");
    action: "app.open_uri";
    target: "https://developer.gnome.org";
  }

  item {
    label: _("Human Interface Guidelines");
    action: "app.open_uri";
    target: "https://developer.gnome.org/hig/";
  }

  item {
    label: _("GNOME Javascript");
    action: "app.open_uri";
    target: "https://gjs-docs.gnome.org";
  }
```

andyholmes (changes requested):
Great job on this, I think you picked up the concepts quite quickly!

src/Library/demos/Actions/main.blp
```diff
@@ -0,0 +1,234 @@
...
+
+            styles ["title-4"]
+          }
+
+          Label {
+            halign: start;
+            label: _("Stateless actions that optionally take parameters");
+
+            styles ["dim-label"]
+          }
```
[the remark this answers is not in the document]
AkshayWarrier: That is much more cleaner, good idea :)

src/Library/demos/Actions/main.blp
AkshayWarrier: So should I change them to ``app.open_uri`` or keep them ``demo.open-bookmarks``? I wanted to show an example of an action that only takes parameters and has no state. I'm not sure how to squeeze in a demo for ``app`` scoped actions though, it might be weird to have an "About App" section under the bookmarks?

src/Library/demos/Actions/main.blp
```diff
@@ -189,4 +189,20 @@ menu bookmarks_menu {
     action: "demo.open-bookmarks";
     target: _("GNOME Javascript");
   }
+
+  section {
+    label: "Actions Documentation";
```
andyholmes, line 194: Just get these three translatables, and it looks good to merge.

andyholmes (changes requested):

andyholmes (approved):

PR #398, opened 2023-07-15
Documentation Viewer: Filter irrelevant docs

### LINKED ISSUE #372: Filter out irrelevant documentation packages
`org.gnome.Sdk` comes with certain things that are there for historical or compatibility reasons. We don't want to expose users to them as it might mislead them.

Let's disable the following namespaces/docs in the documentation viewer:

```js
// For GLib based applications the [GNotification](https://docs.gtk.org/gio/class.Notification.html) API should be used
instead.
// https://gitlab.gnome.org/GNOME/libnotify/-/blob/69aff6e5fa2842e00b409c348bd73188548828b3/README.md
`Notify` ,
// GTK4 applications should use Libadwaita 
`Handy`
```

AkshayWarrier:
Closes #372

### DIFF
--- a/src/DocumentationViewer.js
+++ b/src/DocumentationViewer.js
@@ -91,7 +91,14 @@ export default function DocumentationViewer({ application }) {
     webview.load_uri(row.uri);
   });
 
-  getDocs(base_path)
+  getDocs(base_path, [
+    "atk",
+    "javascriptcoregtk-4.1",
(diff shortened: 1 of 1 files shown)


andyholmes (approved):
LGTM, were you all done with this for now?

If so, lets merge what's here and we can figure out more docs (or mini-prefs) later on.

PR #400, opened 2023-07-16
library: Add EmojiChooser entry

### LINKED ISSUE #380: Add GtkEmojiChooser Demo
You usually get this for free with text extry widgets, but there are certainly other cases where you might want this.

We have in mind a demo that shows a button with a default emoji such as 😃 that opens the GtkEmojiChooser and when an emoji is clicked, the button label is updated to that emoji.

AkshayWarrier:
Closes #380


andyholmes (approved):
Nicely done, I just added a few spaces (we need an auto-formatter :P).

PR #403, opened 2023-07-16
library: Adds file/uri launcher entry

### LINKED ISSUE #390: Add a GtkFileLauncher and GtkUriLauncher Demo
https://docs.gtk.org/gtk4/class.FileLauncher.html

https://docs.gtk.org/gtk4/class.UriLauncher.html

SoNiC-HeRE:
Closes #390

### DIFF
--- a/src/Library/demos/Launcher/main.blp
+++ b/src/Library/demos/Launcher/main.blp
@@ -2,8 +2,8 @@ using Gtk 4.0;
 using Adw 1;
 
 Adw.StatusPage {
-  title: "Launcher";
-  description: _("Present a dialog of apps required to open a File/Uri");
+  title: _("Launcher");
+  description: _("Collect the arguments that are needed to open a File or URI");
 
   Adw.Clamp {
     maximum-size: 640;
@@ -40,10 +40,10 @@ Adw.StatusPage {
                 styles ["linked"]
 
                 Button file_details {
-                  label: _("File Details");
+                  label: _("Open File");
                 }
                 Button file_location {
-                  label: _("File Location");
+                  label: _("Open Folder");
                 }
               }
 
@@ -60,7 +60,7 @@ Adw.StatusPage {
 
         Gtk.StackPage {
           name: "uri_launcher";
-          title: _("Uri Launcher");
+          title: _("URI Launcher");
           child:
             Box {
               orientation: vertical;
@@ -74,8 +74,8 @@ Adw.StatusPage {
                 }
               }
 
-              Button uri_details {
-                label: _("Uri Details");
+              Entry uri_details {
+                placeholder-text: _("Enter Uri");
               }
 
               LinkButton {
--- a/src/Library/demos/Launcher/main.blp
+++ b/src/Library/demos/Launcher/main.blp
@@ -75,7 +75,7 @@ Adw.StatusPage {
               }
 
               Entry uri_details {
-                placeholder-text: _("Enter Uri");
+                placeholder-text: _("Enter URI");
               }
 
               LinkButton {
--- a/src/Library/demos/Launcher/main.blp
+++ b/src/Library/demos/Launcher/main.blp
@@ -30,27 +30,34 @@ Adw.StatusPage {
               spacing: 24;
               Box {
                 halign: center;
+                spacing: 18;
+                orientation: horizontal;
+
                 Button launch_file {
                   label: _("Launch");
                   styles ["suggested-action", "pill"]
                 }
-              }
-
-              Box {
-                styles ["linked"]
-
-                Button file_details {
-                  label: _("File Name");
-                }
                 Button file_location {
-                  label: _("Open Folder");
+                  label: _("View");
+                  styles ["suggested-action", "pill"]
                 }
               }
 
               Button change_file {
-                label: _("Choose another file");
+                label: _("Select File");
+                child: Box {
+                  halign: center;
+                  orientation: vertical;
+                  Label {
+                    label: _("Select File");
+                    styles ["title-4"]
+                  }
+                  Label file_name {
+                    label: _("workbench.txt");
+                    styles ["dim-label", "body"]
+                  }
+                };
               }
-
               LinkButton {
                 label: _("API Reference");
                 uri : "https://docs.gtk.org/gtk4/class.FileLauncher.html";
--- a/src/Library/demos/Launcher/main.blp
+++ b/src/Library/demos/Launcher/main.blp
@@ -1,0 +1,94 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "Launcher";
+  description: _("Present a dialog of apps required to open a File/Uri");
+
+  Adw.Clamp {
+    maximum-size: 640;
+    Box {
+      orientation: vertical;
+      spacing: 24;
+
+      Gtk.StackSwitcher {
+        stack: stack;
+        halign: center;
+      }
+
+      Gtk.Stack stack {
+        transition-type: crossfade;
+        vexpand: true;
+
+        Gtk.StackPage {
+          name: "file_launcher";
+          title: _("File Launcher");
+          child:
+            Box {
+              orientation: vertical;
+              halign: center;
+              spacing: 24;
+              Box {
+                halign: center;
+                Button launch_file {
+                  label: _("Launch");
+                  styles ["suggested-action", "pill"]
+                }
+              }
+
+              Box {
+                styles ["linked"]
+
+                Button file_details {
+                  label: _("File Details");
+                }
+                Button file_location {
+                  label: _("File Location");
+                }
+              }
+
+              Button change_file {
+                label: _("Choose another file");
+              }
+
+              LinkButton {
+                label: _("API Reference");
+                uri : "https://docs.gtk.org/gtk4/class.FileLauncher.html";
+              }
+          };
+        }
+
+        Gtk.StackPage {
+          name: "uri_launcher";
+          title: _("Uri Launcher");
+          child:
+            Box {
+              orientation: vertical;
+              halign: center;
+              spacing: 24;
+              Box {
+                halign: center;
+                Button uri_launch {
+                  label: _("Launch");
+                  styles ["suggested-action", "pill"]
+                }
+              }
+
+              Button uri_details {
+                label: _("Uri Details");
+              }
+
+              LinkButton {
+                label: _("API Reference");
+                uri : "https://docs.gtk.org/gtk4/class.UriLauncher.html";
+              }
+          };
+        }
+      }
+    }
+  }
+}
+
+
+
+
--- a/src/Library/demos/Launcher/main.blp
+++ b/src/Library/demos/Launcher/main.blp
@@ -40,7 +40,7 @@ Adw.StatusPage {
                 styles ["linked"]
 
                 Button file_details {
-                  label: _("Open File");
+                  label: _("File Name");
                 }
                 Button file_location {
                   label: _("Open Folder");
@@ -70,12 +70,13 @@ Adw.StatusPage {
                 halign: center;
                 Button uri_launch {
                   label: _("Launch");
+                  sensitive: false;
                   styles ["suggested-action", "pill"]
                 }
               }
 
               Entry uri_details {
-                placeholder-text: _("Enter URI");
+                placeholder-text: _("Enter Uri");
               }
 
               LinkButton {
@@ -88,7 +89,3 @@ Adw.StatusPage {
     }
   }
 }
-
-
-
-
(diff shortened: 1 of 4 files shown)


SoNiC-HeRE:
[Kooha-2023-07-16-20-04-06.webm](https://github.com/sonnyp/Workbench/assets/96797205/b6ede0eb-d0dd-42d2-a0c4-d888f2eab0f3)

SoNiC-HeRE:
I had another idea for Uri Launcher to dynamically change uri using a file from FileChooser and getting its uri similar to file launcher; however when i use this : ```uri_launcher.set_uri(file.get_uri());```
 I get the warning: ```Gjs-WARNING **: 20:21:56.432: JS ERROR: Gtk.DialogError: The application launch failed
```
and it does'nt open the file

**Comment by andyholmes** [comment]
> ...and it does'nt open the file

This is probably because the `Gtk.UriLauncher` doesn't know to add the file to the documents portal, but it's hard to say. If I set it to a URL it seems to work fine.

**Comment by SoNiC-HeRE** [comment]
> > ...and it does'nt open the file
> 
> This is probably because the `Gtk.UriLauncher` doesn't know to add the file to the documents portal, but it's hard to say. If I set it to a URL it seems to work fine.

Is the current demo fine? or should i try fixing the above error?  An early review might help me with it

File: src/Library/demos/Launcher/main.blp [hunk 1]
```diff
@@ -0,0 +1,94 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "Launcher";
```
andyholmes, line 5:
```suggestion
  title: _("Launcher");
```

andyholmes:
> Is the current demo fine? or should i try fixing the above error? An early review might help me with it

I think just remove that part.

I'm only guessing why that's happening, but we should encourage developers to use `Gtk.FileLauncher` for any file-related operations since they're almost always going to be local and restricted by any sandboxing that's active.

src/Library/demos/Launcher/main.blp
```diff
@@ -0,0 +1,94 @@
...
+  description: _("Present a dialog of apps required to open a File/Uri");
```
andyholmes, line 6:
```suggestion
  description: _("Collect the arguments that are needed to open a File or URI");
```

The documentation says what happens depends on the desktop environment and might not open a dialog, so I think this is a bit better.

src/Library/demos/Launcher/main.blp
```diff
@@ -0,0 +1,94 @@
...
+                halign: center;
+                Button uri_launch {
+                  label: _("Launch");
+                  styles ["suggested-action", "pill"]
+                }
+              }
+
+              Button uri_details {
+                label: _("Uri Details");
+              }
```
andyholmes, line 79: How about a `Gtk.Entry` here instead? You could connect to the `changed` signal if you want and call `Gio.Uri.parse()` on the result to ensure it's valid.

src/Library/demos/Launcher/main.blp
```diff
@@ -0,0 +1,94 @@
...
+              LinkButton {
+                label: _("API Reference");
+                uri : "https://docs.gtk.org/gtk4/class.FileLauncher.html";
+              }
+          };
+        }
+
+        Gtk.StackPage {
+          name: "uri_launcher";
+          title: _("Uri Launcher");
```
andyholmes, line 63:
```suggestion
          title: _("URI Launcher");
```

I'm just nitpicking about capitalization here; it's not really mistake.

src/Library/demos/Launcher/main.blp
```diff
@@ -0,0 +1,94 @@
...
+                  label: _("Launch");
+                  styles ["suggested-action", "pill"]
+                }
+              }
+
+              Box {
+                styles ["linked"]
+
+                Button file_details {
+                  label: _("File Details");
```
andyholmes, line 43: I think rather than a button, just a label that's set when the file is changed would be good.

src/Library/demos/Launcher/main.blp
```diff
@@ -0,0 +1,94 @@
...
+                }
+                Button file_location {
+                  label: _("File Location");
```
andyholmes, line 46: Since this one also "opens" something, how about one button that says "Open File" and another that says "Open Folder"?

src/Library/demos/Launcher/main.blp
```diff
@@ -0,0 +1,94 @@
...
+              Box {
+                halign: center;
+                Button uri_launch {
+                  label: _("Launch");
+                  styles ["suggested-action", "pill"]
+                }
+              }
+
+              Button uri_details {
+                label: _("Uri Details");
```
andyholmes, line 78:
```suggestion
                label: _("URI Details");
```

src/Library/demos/Launcher/main.js
```diff
@@ -0,0 +1,91 @@
...
+
+Gio._promisify(Gtk.FileLauncher.prototype, "launch", "launch_finish");
+Gio._promisify(
+  Gtk.FileLauncher.prototype,
+  "open_containing_folder",
+  "open_containing_folder_finish",
+);
+Gio._promisify(Gtk.FileDialog.prototype, "open", "open_finish");
+Gio._promisify(Gtk.UriLauncher.prototype, "launch", "launch_finish");
+
```
andyholmes, line 22: All the code here great, let's just separate them into two chunks with a comment like:

```js
// File Launcher
// < launcher code here >

// URI Launcher
// < launcher code here >

**Review by andyholmes — CHANGES_REQUESTED** [review]
Looking good!

As usual, I've got some "opinions" about the UI, but I think you can get this one finished in the next run :)

File: src/Library/demos/Launcher/main.js [hunk 9]
```diff
@@ -86,6 +88,7 @@ uri_launch.connect("clicked", () => {
   launchesUri().catch(logError);
 });
 
-uri_details.connect("clicked", () => {
-  console.log(`File: ${uri_launcher.get_uri()}`);
+uri_details.connect("changed", () => {
+  text = uri_details.get_text();
+  uri_launcher.set_uri(text);
```
andyholmes, line 93:
```suggestion
  uri_launcher.set_uri(text);
  
  try {
    uri_launch.sensitive = GLib.Uri.is_valid(text, GLib.UriFlags.NONE);
  } catch {
    uri_launch.sensitive = false;
  }
```

Let's tie the sensitivity of the button to the URI.

src/Library/demos/Launcher/main.js
```diff
@@ -0,0 +1,94 @@
...
+  const result = await file_launcher.open_containing_folder(
+    workbench.window,
+    null,
+  );
+  if (result) {
+    console.log("Parent folder opened");
+  }
+}
+
+async function changesFile() {
```
andyholmes, line 48:
```suggestion
async function changesFile() {
```

Let's do away with these empty wrappers.

src/Library/demos/Launcher/main.js
```diff
@@ -0,0 +1,94 @@
...
+  const details = file_info.query_info(
+    "standard::*",
+    Gio.FileQueryInfoFlags.NOFOLLOW_SYMLINKS,
+    null,
+  );
+  console.log(`File: ${details.get_name()}`);
+});
+
+file_location.connect("clicked", () => {
+  launchesParentFolder().catch(logError);
```
andyholmes, line 75:
```suggestion
  file_launcher.open_containing_folder(workbench.window, null).catch(logError);
```

src/Library/demos/Launcher/main.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "Launcher",
+  "category": "controls",
+  "description": "Present a dialog of apps required to open a file/uri",
```
andyholmes, line 4:
```suggestion
  "description": "Collect the arguments that are needed to open a File or URI",
```

src/Library/demos/Launcher/main.js
```diff
@@ -0,0 +1,94 @@
...
+  const file_info = file_launcher.get_file();
+  const details = file_info.query_info(
+    "standard::*",
+    Gio.FileQueryInfoFlags.NOFOLLOW_SYMLINKS,
+    null,
+  );
+  console.log(`File: ${details.get_name()}`);
+});
+
+file_location.connect("clicked", () => {
```
andyholmes, line 74:
```suggestion
file_location.connect("clicked", () => {
```

I think this button sort just confuses what the "Launch" button does, let's skip it.

src/Library/demos/Launcher/main.blp
```diff
@@ -0,0 +1,94 @@
...
+              Box {
+                styles ["linked"]
+
+                Button file_details {
+                  label: _("Open File");
+                }
+                Button file_location {
+                  label: _("Open Folder");
+                }
+              }
```
andyholmes, line 48:
```suggestion
              Button file_location {
                label: _("Open Folder");
              }
```

Removing the "Open File", which conflicts with the meaning of the "Launch" button. The "Open Folder" folder button is a good idea, but I think it should be somehow connected or associate with the "Launch" button; any ideas?

andyholmes (changes requested):
Lots of cleanups here, to keep the demo focused on the launcher APIs.

The only thing left I'd really like to cleanup is the "Launch"/"Open Folder" buttons, which should be associated with each other.

src/Library/demos/Launcher/main.blp
SoNiC-HeRE: I thought a workaround for this; I changed the open file to file name instead so that it doesnt conflict with launch button; I had another idea how about renaming this to File details wherein i log the details of the file such as size, name etc

andyholmes: I'm just not sure what purpose it serves, and it seems a bit confusing that it is presented with the same importance as "Open Folder".

"Launch" and "Open Folder" both represent methods of the `Gtk.FileLauncher` class, while the other button simply prints the file name to the console. Maybe instead the file selection button could have a caption with the current file name?

src/Library/demos/Launcher/main.js
```diff
@@ -0,0 +1,80 @@
...
+Gio._promisify(Gtk.UriLauncher.prototype, "launch", "launch_finish");
+
+const file_launcher = new Gtk.FileLauncher();
+const uri_launcher = new Gtk.UriLauncher();
+const file = Gio.File.new_for_path(pkg.pkgdatadir).resolve_relative_path(
+  "Library/demos/Launcher/workbench.txt",
+);
+file_launcher.set_file(file);
+let text = "";
+
```
andyholmes, line 30:
```suggestion

```

This is never used outside the callback (anymore), so lets just make it `const` where it's used.

src/Library/demos/Launcher/main.js
```diff
@@ -0,0 +1,80 @@
...
+    logError(err);
+  }
+});
+
+// URI Launcher
+uri_launch.connect("clicked", () => {
+  uri_launcher.launch(workbench.window, null).catch(logError);
+});
+uri_details.connect("changed", () => {
+  text = uri_details.get_text();
```
andyholmes, line 71:
```suggestion
  const text = uri_details.get_text();
```

andyholmes (changes requested):

src/Library/demos/Launcher/main.blp
andyholmes: Here's what I would do:

* Move the constructors for both launchers to their respective sections in the code view.
* Connect to `notify::file` and query the file name to set a caption on the "Choose another file" button
* Set the default file and URI after the signals are connected, so they populate at first run
* Make an equally prominent button for "View"

Here is what I'd imagine that looking like:

[Screencast from 2023-08-05 16-52-36.webm](https://github.com/sonnyp/Workbench/assets/1265208/b6e910d8-d40a-48fb-a487-52cc932adbbd)

src/Library/demos/Launcher/main.js
```diff
@@ -0,0 +1,80 @@
...
+  file_launcher.open_containing_folder(workbench.window, null).catch(logError);
+});
+
+change_file.connect("clicked", () => {
+  try {
+    const dialog_for_file = new Gtk.FileDialog({
+      title: _("Select File"),
+      modal: true,
+    });
+    const new_file = dialog_for_file.open(workbench.window, null);
```
andyholmes, line 56: Sorry, I missed this in the review. This is a `Promise`, and you're not actually passing the resulting file to the launcher. Something like this makes more sense here:

```js
  let new_file = null;

  const dialog_for_file = new Gtk.FileDialog({
    title: _("Select File"),
    modal: true,
  });
  dialog_for_file
    .open(workbench.window, null)
    .then((result) => {
      new_file = result;
    })
    .catch((err) => {
      console.debug(err.message);
    })
    .finally(() => {
      file_launcher.set_file(new_file);
    });
```

andyholmes (changes requested):
Looks good, but the "Open Folder" still needs to be emphasized as part of the `Gtk.FileLauncher` class.

The "File Name" button could just be a label instead, and the "Open Folder" button placed beside the "Launch" button.

andyholmes (approved):
Nicely done!

PR #425, opened 2023-07-17
library: Add Tooltip entry

### LINKED ISSUE #422: Gtk Tooltip
Should be demonstrated with `set_tooltip_text` / `set_tooltip_markup` but also with a custom one.

https://docs.gtk.org/gtk4/class.Tooltip.html

AkshayWarrier:
Closes #422

I'm having some trouble with the custom tooltip. It usually works fine the first time, but then occasionally I get these errors and the tooltip stops working.
```
(re.sonny.Workbench.Devel:4): Gtk-CRITICAL **: 18:00:43.428: gtk_box_append: assertion 'gtk_widget_get_parent (child) == NULL' failed

(re.sonny.Workbench.Devel:4): Gtk-CRITICAL **: 18:00:45.182: gtk_box_remove: assertion 'gtk_widget_get_parent (child) == (GtkWidget *)box' failed
```

### DIFF
--- a/src/Library/demos/Tooltip/main.blp
+++ b/src/Library/demos/Tooltip/main.blp
@@ -1,0 +1,81 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: _("Tooltip");
+  description: _("Show additional information about controls or app content");
+
+  Box {
+    halign: center;
+    valign: center;
+    spacing: 24;
+    orientation: vertical;
+
+    Box {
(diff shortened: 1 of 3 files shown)


andyholmes (approved):
I think this is all good.

I did encounter some different warnings, but I don't see that you're doing anything incorrect. It seems like the custom tooltip widget doesn't need to be replaced on each `query-tooltip` emission, but there's no way to query the child widget, so I don't think that's an issue either.

Good job!

PR #426, opened 2023-07-17
library: Add vala demo for Scale

SpikedPaladin:
Adds Vala code demo for Scale

### DIFF
--- a/src/Library/demos/Scale/main.vala
+++ b/src/Library/demos/Scale/main.vala
@@ -18,9 +18,9 @@ public void main () {
   scale_one.value_changed.connect(() => {
     var scale_value = scale_one.get_value ();
     if (scale_value == scale_one.adjustment.upper) {
-      stdout.printf ("Maximum value reached");
+      message ("Maximum value reached");
     } else if (scale_value == scale_one.adjustment.lower) {
-      stdout.printf ("Minimum value reached");
+      message ("Minimum value reached");
     }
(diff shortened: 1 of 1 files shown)


src/Library/demos/Scale/main.vala
```diff
@@ -0,0 +1,37 @@
...
+    var scale_value = scale_one.get_value ();
+    if (scale_value == scale_one.adjustment.upper) {
+      stdout.printf ("Maximum value reached");
+    } else if (scale_value == scale_one.adjustment.lower) {
+      stdout.printf ("Minimum value reached");
+    }
+  });
+
+  scale_two.value_changed.connect(() => {
+    var scale_value = scale_two.get_value ();
```
lw64, line 28: I wouldn't use var here. What type is this?

src/Library/demos/Scale/main.vala
```diff
@@ -0,0 +1,37 @@
...
+    scale_value /= 50;
+    if (scale_value.to_string().contains(".")) return;
```
lw64, line 30: This line is not very clear. What is the code trying to do here? Checking if scale_value is a whole number? Either put a comment or use another method that makes it better understandable.

lw64 (changes requested):
Everything else looks alright!

SpikedPaladin:
@lw64 updated

PR #427, opened 2023-07-17
library: Add separator entry

### LINKED ISSUE #420: Gtk Separator
https://docs.gtk.org/gtk4/class.Separator.html

SoNiC-HeRE:
Closes #420

### DIFF
--- a/src/Library/demos/Separator/main.blp
+++ b/src/Library/demos/Separator/main.blp
@@ -1,0 +1,90 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: "Separator";
+  description: _("Separates one widget from another");
+
+  Box {
+    orientation: vertical;
+    halign: center;
+
+    Box {
+      orientation: horizontal;
+      halign: center;
+      spacing: 18;
+      margin-bottom: 18;
+      homogeneous: false;
+
+      Box {
+        orientation: vertical;
+        halign: center;
+        spacing: 12;
+
+        Picture picture_one {
+          width-request: 180;
+          height-request: 180;
+        }
+      }
+
+      Separator {
+        margin-start: 6;
+        margin-end: 6;
+      }
+
+      Box {
+        orientation: vertical;
+        halign: center;
+        spacing: 12;
+
+        Picture picture_two {
+          width-request: 180;
+          height-request: 180;
+        }
+      }
+    }
+
+    Separator {
+      margin-bottom: 24;
+    }
+
+    Label {
+      label:_("Separator with toolbar class");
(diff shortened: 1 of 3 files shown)


SoNiC-HeRE:
image

src/Library/demos/Separator/main.blp
```diff
@@ -0,0 +1,90 @@
...
+      Box {
+        orientation: vertical;
+        halign: center;
+        spacing: 12;
+
+        Picture picture_one {
+          width-request: 180;
+          height-request: 180;
+        }
+      }
```
andyholmes, line 28: Let's drop the boxes wrapping the `Picture` widgets.

src/Library/demos/Separator/main.blp
```diff
@@ -0,0 +1,90 @@
...
+
+  Box {
+    orientation: vertical;
+    halign: center;
+
+    Box {
+      orientation: horizontal;
+      halign: center;
+      spacing: 18;
+      margin-bottom: 18;
```
andyholmes, line 16:
```suggestion
```

Let's remove this `margin: 18;` and make it `spacing: 18;` in the parent box.

src/Library/demos/Separator/main.blp
```diff
@@ -0,0 +1,90 @@
...
+
+        Picture picture_one {
+          width-request: 180;
+          height-request: 180;
+        }
+      }
+
+      Separator {
+        margin-start: 6;
+        margin-end: 6;
```
andyholmes, line 32:
```suggestion
```

We can remove this, and just increase the parent's `spacing` if necessary.

src/Library/demos/Separator/main.blp
```diff
@@ -0,0 +1,90 @@
...
+
+        Picture picture_two {
+          width-request: 180;
+          height-request: 180;
+        }
+      }
+    }
+
+    Separator {
+      margin-bottom: 24;
```
andyholmes, line 48:
```suggestion
```

Same with the these `margin-bottom` properties; much easier to use and adjust the `spacing` property.

This is a good habit to get into, because later in a project it makes it easier to find out where/what properties are responsible for widget layout.

andyholmes (changes requested):
Nice one! A good simple example of separators, that's easy to copy-paste.

Just cleanup those margins, and replace them with `spacing` which is far easier to maintain in the long-term.

src/Library/demos/Separator/main.blp
SoNiC-HeRE: yes, I've updated the alignments with spacing property

andyholmes (approved):
Great job!

PR #433, opened 2023-07-18
library: Add Link Button entry

### LINKED ISSUE #416: Gtk LinkButton
https://docs.gtk.org/gtk4/class.LinkButton.html

let's demonstrate "visited"

AkshayWarrier:
Closes #416

### DIFF
--- a/src/Library/demos/Link Button/main.blp
+++ b/src/Library/demos/Link Button/main.blp
@@ -1,0 +1,12 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: _("Link Button");
+  description: _("A button with a hyperlink");
+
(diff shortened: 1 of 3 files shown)


andyholmes (changes requested):
Nice simple example, let's just add a quick example of the [`activate-link`](https://docs.gtk.org/gtk4/signal.LinkButton.activate-link.html) signal, since it's the only other thing this does.

Something like:

```js
linkbutton.connect('activate-link', (button) => {
   console.log(`About to activate ${button.uri}`);

   // Return true if handling the link manually, or
   // false to let the default behavior continue
   return false;
});
```

If you want, you could do something fun like re-direct the link, but just adding a comment about it is okay too.

andyholmes (approved):

PR #434, opened 2023-07-19
library: Add Scrolled Window entry

### LINKED ISSUE #419: Gtk Scrolled window
https://docs.gtk.org/gtk4/class.ScrolledWindow.html

AkshayWarrier:
Closes #419
There is an edge case where if the Start/End buttons are pressed when it's already at the edge, the buttons get disabled for the animation duration. Is that problematic and if so, what might be the simplest way to handle it?

### DIFF
--- a/src/Library/demos/Scrolled Window/main.js
+++ b/src/Library/demos/Scrolled Window/main.js
@@ -6,21 +6,20 @@ const container = workbench.builder.get_object("container");
 const toggle_orientation = workbench.builder.get_object("toggle_orientation");
 const button_start = workbench.builder.get_object("button_start");
 const button_end = workbench.builder.get_object("button_end");
+let auto_scrolling = false;
 
 button_start.sensitive = false;
 
-const scrollbars = [
-  scrolled_window.get_hscrollbar(),
-  scrolled_window.get_vscrollbar(),
-];
-let orientation = 0;
+const scrollbars = {
+  [Gtk.Orientation.HORIZONTAL]: scrolled_window.get_hscrollbar(),
+  [Gtk.Orientation.VERTICAL]: scrolled_window.get_vscrollbar(),
+};
+
 toggle_orientation.connect("toggled", () => {
   if (toggle_orientation.active) {
     container.orientation = Gtk.Orientation.HORIZONTAL;
-    orientation = 0;
   } else {
     container.orientation = Gtk.Orientation.VERTICAL;
-    orientation = 1;
   }
 });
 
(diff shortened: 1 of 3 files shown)


sonnyp (changes requested):
Great demo :+1: 

Sorry for taking so long to review.

> Is that problematic and if so, what might be the simplest way to handle it?

When the button is pressed, you can simply check if the scrolled area is already at the edge and do nothing.

I think an even better solution is to disable the corresponding start or end button when it's already at the edge.

start edge is adj value = adj lower
end endge is adj value = adj upper

Also please reduce animation duration to 1 sec, it's a bit too slow now.

AkshayWarrier:
> start edge is adj value = adj lower
> end endge is adj value = adj upper

I can't really do this because weirdly enough this is not true, I tried logging adj.upper and adj.value when the scrollbar is at the end edge and I got upper = 3360 and value = 2782. I'm not sure what adj.upper is then, but the scrollbar probably just stops scrolling after 2782. And the scrollbar doesn't have property to let me know whats the actual maximum distance it can scroll.

AkshayWarrier:
Though in my opinion, I think it's fine the way it is because I'm already reducing the animation speed to 1s so the buttons are not disabled that long, and most people looking for a Scrolled Window demo probably are only looking for the UI/Blueprint and the demo shows how to programmatically scroll if necessary. I'm concerned adding extra logic to decide when the buttons are enabled/disabled is very specific to the demo?

sonnyp:
> I can't really do this because weirdly enough this is not true, I tried logging adj.upper and adj.value when the scrollbar is at the end edge and I got upper = 3360 and value = 2782. I'm not sure what adj.upper is then, but the scrollbar probably just stops scrolling after 2782. And the scrollbar doesn't have property to let me know whats the actual maximum distance it can scroll.

Right, you need to do some simple maths to get positions relative to each others.

Something like
upper - page-size = value
2240 - 598 = 1642

> The [GtkAdjustment:value](https://docs.gtk.org/gtk4/property.Adjustment.value.html) field sets the position of the thumb and must be between [GtkAdjustment:lower](https://docs.gtk.org/gtk4/property.Adjustment.lower.html) and [GtkAdjustment:upper](https://docs.gtk.org/gtk4/property.Adjustment.upper.html) - [GtkAdjustment:page-size](https://docs.gtk.org/gtk4/property.Adjustment.page-size.html). The [GtkAdjustment:page-size](https://docs.gtk.org/gtk4/property.Adjustment.page-size.html) represents the size of the visible scrollable area.

See https://docs.gtk.org/gtk4/class.Scrollbar.html

>. most people looking for a Scrolled Window demo probably are only looking for the UI/Blueprint

Maybe but the complexity here is in `Code`.

>  I'm concerned adding extra logic to decide when the buttons are enabled/disabled is very specific to the demo?

It's good to keep demos simple but knowing if the scroll is at the edge or not is something very common. Specially when you have a "scroll to top" button like on timelines that you want to hide or disable when you're already at the top.

I think it's worth our time.

src/Library/demos/Scrolled Window/main.js
```diff
@@ -0,0 +1,90 @@
...
+const button_start = workbench.builder.get_object("button_start");
+const button_end = workbench.builder.get_object("button_end");
+
+button_start.sensitive = false;
+
+const scrollbars = [
+  scrolled_window.get_hscrollbar(),
+  scrolled_window.get_vscrollbar(),
+];
+let orientation = 0;
```
sonnyp, line 16: Why do we need this if there is already `container.orientation` ?

```
const scrollbars = {
  [Gtk.Orientation.HORIZONTAL]: scrolled_window.get_hscrollbar(),
  [Gtk.Orientation.VERTICAL]: scrolled_window.get_vscrollbar(),
}
```

sonnyp (changes requested):
Almost, the start / end buttons should only be disabled if at the edge.

For that you should listen to the property `value` on the adjustment. Not sure what's best between the signals `value-changed` and `notify::value`. Would be curious to know the difference if you can figure it out.

sonnyp (approved):
Nice :+1: !

Did you figure out the difference between `notify::value` and `value-changed` ?

AkshayWarrier:
Ah not really, I just ended up using ``value-changed`` since ``Gtk.Adjustment`` already provides that signal.

PR #435, opened 2023-07-19
Library: Port Stack entry to Vala

halfmexican:


src/Library/demos/Stack/main.vala
```diff
@@ -0,0 +1,60 @@
+#! /usr/bin/env -S vala workbench.vala --pkg gtk4 --pkg libadwaita-1 --pkg gio-2.0
+
+using Gtk;
+using Adw;
```
lw64, line 4: Don't use `using`, but use for example `Gtk.Box`

src/Library/demos/Stack/main.vala
```diff
@@ -0,0 +1,60 @@
...
+using Gdk;
+using GLib;
+
+public void main() {
+  var root_box = workbench.builder.get_object("root_box") as Box;
```
lw64, line 9: Use the `(Box)` instead of `as Gtk.Box` casting method in this case. Also the variables should be global and only initialized in the main function, so that they can be accessed from other functions as well.

src/Library/demos/Stack/main.vala
```diff
@@ -0,0 +1,60 @@
...
+    root_box.prepend(stack_switcher);
+  } else {
+    stack_sidebar = new Gtk.StackSidebar();
+    separator = new Separator(Gtk.Orientation.HORIZONTAL);
+    stack_sidebar.stack = stack;
+    root_box.prepend(separator);
+    root_box.prepend(stack_sidebar);
+  }
+
+  navigation_row.notify["selected"].connect(() => {
```
lw64, line 30: This should be put into its own function, Maybe "on_navigation_row_changed". So it can be called also once above, instead of duplicating the code.

src/Library/demos/Stack/main.vala
```diff
@@ -0,0 +1,60 @@
...
+    if (navigation_row.selected == 0) {
```
lw64, line 31: Can you check if the string in the item is correct, instead of just the position. That would be better readable.

src/Library/demos/Stack/main.vala
```diff
@@ -0,0 +1,60 @@
...
+      separator = new Separator(Gtk.Orientation.HORIZONTAL);
+      stack_sidebar = new StackSidebar();
+      stack_sidebar.stack = stack;
+      root_box.prepend(separator);
+      root_box.prepend(stack_sidebar);
+      root_box.set_orientation(Gtk.Orientation.HORIZONTAL);
+    }
+  });
+
+  stack.set_transition_type(transition_row.selected);
```
lw64, line 55: Why is this not done in the xml?

lw64 (changes requested):
I commented on what issues I have with the port to Vala, but I think there are also other things, that could be improved in the javascript code and xml/blueprint as well.

First I wouldnt create and remove widgets dynamically, but define everything in the xml and only make certain widgets visible and invisible again. A lot can be also done just through property bindings/expressions, so a lot less code would be needed. Also Adw.EnumListModel exists, which could be used for the transitions. Whats your opinion @sonnyp ?

sonnyp:
@lw64 

>  but I think there are also other things, that could be improved in the javascript code and xml/blueprint as well.

This is out of scope for a port PR, or it becomes unmanageable for contributors and discourage porting contributions.
Please file an issue if you'd like to suggest improvements to a demo.

lw64:
Right, Will do that.

lw64 (approved):

PR #436, opened 2023-07-19
library: Add Image entry

### LINKED ISSUE #415: Gtk Image library demo
https://docs.gtk.org/gtk4/class.Image.html

> GtkImage displays its image as an icon, with a size that is determined by the application. See [GtkPicture](https://docs.gtk.org/gtk4/class.Picture.html) if you want to show an image at is actual size.

Would be nice if the demo mentioned to use GtkPicture when not using icons.

We can also refer to the Icon Library in Workbench.

AkshayWarrier:
Closes #415


sonnyp (approved):
Looks good :+1:

PR #437, opened 2023-07-20
library: Adds columnview entry

SoNiC-HeRE:


SoNiC-HeRE:
I've implemented the following:
 - Implemented Column View
 - Used DirectoryList Model
 - Added 3 Columns

src/Library/demos/Column View/main.js
```diff
@@ -0,0 +1,101 @@
...
+      height_request: 50,
+      margin_start: 12,
+      margin_end: 12,
+    });
+    listItem.set_child(label);
+  });
+  factory.connect("bind", (factory, listItem) => {
+    const labelWidget = listItem.get_child();
+    const modelItem = listItem.get_item();
+    labelWidget.label = modelItem.get_name();
```
andyholmes, line 22: Let's use `Gio.FileInfo.get_display_name()`, which may use a more human-readable name and then fallback to the standard `name` if necessary.

src/Library/demos/Column View/main.js
```diff
@@ -0,0 +1,101 @@
...
+const column_view = workbench.builder.get_object("column_view");
+
+function createFirstCol() {
+  const factory = new Gtk.SignalListItemFactory();
+  const columnViewColumn = new Gtk.ColumnViewColumn({
+    factory,
+  });
+  factory.connect("setup", (factory, listItem) => {
+    const label = new Gtk.Label({
+      height_request: 50,
```
andyholmes, line 13:
```suggestion
```

Best to keep these at the height of the label, considering how many items are in the list.

src/Library/demos/Column View/main.js
```diff
@@ -0,0 +1,101 @@
...
+      margin_start: 12,
+      margin_end: 12,
+    });
+    listItem.set_child(label);
+  });
+  factory.connect("bind", (factory, listItem) => {
+    const labelWidget = listItem.get_child();
+    const modelItem = listItem.get_item();
+  });
```
andyholmes, line 64: In this column, we can use `Gio.FileInfo.get_modification_date_time()`, just be sure to set the attributes to `"standard::*,time:modified"` so that attribute is available.

Then something simple like this can go in the label:

```js
const date = modelItem.get_modification_date_time();
labelWidget.label = date.format("%F");
```

src/Library/demos/Column View/main.js
```diff
@@ -0,0 +1,101 @@
...
+  factory.connect("setup", (factory, listItem) => {
+    const label = new Gtk.Label({
+      height_request: 50,
+      margin_start: 12,
+      margin_end: 12,
+    });
+    listItem.set_child(label);
+  });
+  factory.connect("bind", (factory, listItem) => {
+    const labelWidget = listItem.get_child();
```
andyholmes, line 20: Don't forget the project [code style](https://github.com/sonnyp/Workbench/wiki/Contribution-Style-Guide#general-principles-2) uses `camelCase` for function names, and`snake_case` for variable names.

andyholmes (changes requested):
This is starting to shape up.

Let's get the UI a little more polished, because we still need to demonstrate column sorters, resizing, and probably want to do column that has a different type like an icon.

src/Library/demos/Column View/main.blp
```diff
@@ -0,0 +1,27 @@
...
+      LinkButton {
+        label: _("Documentation");
+        uri: "https://docs.gtk.org/gtk4/class.ColumnView.html";
+      }
+
+      ColumnView column_view {
+        show-column-separators: true;
+        show-row-separators: true;
+        styles ['view']
+      }
```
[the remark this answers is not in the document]
SoNiC-HeRE: I've implemented the code for now but i can update it later.

src/Library/demos/Column View/main.js
SoNiC-HeRE: I get date as null ; I also tried using ```fileInfo.get_attribute_uint64()``` but still it shows empty.

src/Library/demos/Column View/main.js
```diff
@@ -0,0 +1,91 @@
...
+    const label = new Gtk.Label({
+      margin_start: 12,
+      margin_end: 12,
+    });
+    list_item.set_child(label);
+  });
+  factory.connect("bind", (factory, list_item) => {
+    const label_widget = list_item.get_child();
+    const model_item = list_item.get_item();
+    label_widget.label = model_item.get_size().toString();
```
andyholmes, line 37: Probably we want to use a human readable size here, but let's leave that for the polish phase.

src/Library/demos/Column View/main.blp
```diff
@@ -0,0 +1,41 @@
...
+        ColumnViewColumn col2 {
+          title: _("Size (in bytes)");
+          factory: SignalListItemFactory {};
+        }
+
+        ColumnViewColumn col3 {
+          title: _("Date");
+          factory: SignalListItemFactory {};
+        }
+      }
```
andyholmes, line 38:
```suggestion
      Frame {
        ColumnView column_view {
          show-column-separators: true;
          show-row-separators: true;

          ColumnViewColumn col1 {
            title: _("Name");
            factory: SignalListItemFactory{};
          }

          ColumnViewColumn col2 {
            title: _("Size (in bytes)");
            factory: SignalListItemFactory {};
          }

          ColumnViewColumn col3 {
            title: _("Date");
            factory: SignalListItemFactory {};
          }
        }
      }
```

Frames are pretty.

andyholmes (changes requested):
Cool, once the date column is reporting a basic date, it's on to the "fun" part: column sorting.

Do you want to work through column sorting together, or if I link you to [`Gtk.ColumnView.get_sorter()`](https://docs.gtk.org/gtk4/method.ColumnView.get_sorter.html) as a hint do you can get the name column sorting?

src/Library/demos/Column View/main.js
```diff
@@ -0,0 +1,123 @@
...
+
+const model = new Gtk.SingleSelection({
+  model: new Gtk.SortListModel({
+    model: new Gtk.DirectoryList({
+      file: dir,
+      attributes: "standard::*,time::modified",
+    }),
+    sorter: column_view.sorter,
+  }),
+});
```
andyholmes, line 80: Cool, now that you've got a handle on this, let's break this into separate variables so it's *really* clear what's going on. This is one of the tougher things for people to grasp.

I would do something like:

```js
const data_model = new Gtk.DirectoryList({ ... });

const sort_model = new Gtk.SortListModel({ ... });

column_view.model = new Gtk.SingleSelection({ ... });
```

This way those reading the code don't have to unravel the nesting "backwards", and its really obvious what's being done.

src/Library/demos/Column View/main.js
```diff
@@ -0,0 +1,123 @@
+import Gtk from "gi://Gtk";
+import Gio from "gi://Gio";
+import GObject from "gi://GObject";
+
+const column_view = workbench.builder.get_object("column_view");
+const col1 = workbench.builder.get_object("col1");
+const col2 = workbench.builder.get_object("col2");
+const col3 = workbench.builder.get_object("col3");
+
+function createFirstCol() {
```
andyholmes, line 10: Let's remove all the function wrappers on these, and add comments like:

```
// View

// View Column 1
...

// View Column 2
...
```

Under each column comment, we will define the factory, then the sorter, since these are all the "view" part of the model-view-controller (there's no controller in this demo though and that's fine).

src/Library/demos/Column View/main.js
```diff
@@ -0,0 +1,123 @@
...
+  ),
+});
+
+col3.sorter = new Gtk.NumericSorter({
+  expression: new Gtk.ClosureExpression(
+    GObject.TYPE_INT64,
+    (fileInfo) => fileInfo.get_modification_date_time().to_unix(),
+    null,
+  ),
+});
```
andyholmes, line 104: Move these up to the previously mentioned "view" sections.

src/Library/demos/Column View/main.js
```diff
@@ -0,0 +1,123 @@
...
+    `position: ${position}, Item removed? ${Boolean(
+      removed,
+    )}, Item added? ${Boolean(added)}`,
+  );
+});
+
+model.connect("selection-changed", () => {
+  const selected_item = model.get_selected();
+  console.log(`Model item selected from view: ${model.model}`);
+});*/
```
andyholmes, line 123:
```suggestion
```

We won't bother with demonstrating the selection model itself, we did that in a previous demo.

src/Library/demos/Column View/main.js
```diff
@@ -0,0 +1,123 @@
...
+    (fileInfo) => fileInfo.get_modification_date_time().to_unix(),
+    null,
+  ),
+});
+
+column_view.model = model;
+createFirstCol();
+createSecCol();
+createThirdCol();
+
```
andyholmes, line 110:
```suggestion
```

These wrappers are being removed, as mentioned above.

src/Library/demos/Column View/main.js
andyholmes: Be sure to double-check my math, but this should be good enough:

```js
if (Math.floor(size / 1_000_000_000) > 0)
  label_widget.label = `${(size / 1_000_000_000).toFixed(1)} GB`
else if (Math.floor(size / 1_000_000) > 0)
  label_widget.label = `${(size / 1_000_000).toFixed(1)} MB`
else if (Math.floor(size / 1000) > 0)
  label_widget.label = `${(size / 1000).toFixed(1)} kB`
else
  label_widget.label = `${(size / 1000)} bytes`
```

andyholmes (changes requested):
Nice job!

This is tough one, and it's going to help a *lot* of people. Let's get the thing polished up, and it's good to merge.

andyholmes (approved):
:tada:

PR #441, opened 2023-07-22
GNOME 45

sonnyp:


re.sonny.Workbench.Devel.json
```diff
@@ -224,7 +224,6 @@
         }
       ],
       "post-install": [
-        "cp -a /usr/share/doc/. /app/share/doc/",
```
sonnyp, line 227: couldn't make this work so far

> cp: cannot overwrite directory '/app/share/doc/./appstream' with non-directory

andyholmes: Maybe `--no-clobber` will allow it to copy without hitting the error? It might just bail though, not sure.

sonnyp: nope,

I don't understand the problem. There is no `appstream` directory in `/app/share/doc/` and the same command in an interactive shell works fine

sonnyp:
https://github.com/sonnyp/Workbench/pull/475

PR #445, opened 2023-07-22
library: Add Menu entry

### LINKED ISSUE #321: Add Menu library entry

AkshayWarrier:
Closes #321

### DIFF
--- a/src/Library/demos/Menu/main.blp
+++ b/src/Library/demos/Menu/main.blp
@@ -1,0 +1,159 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage demo {
+  title: _("Menu");
+  description: _("Display structured menus with items, sections and submenus");
+
+  Box {
+    halign: center;
+    orientation: vertical;
+
+    Box {
+      orientation: vertical;
+      spacing: 42;
+
+      Adw.PreferencesGroup {
+        title: _("Primary and Secondary Menus");
+        description: _("Used with controls such as menu buttons");
+
+        MenuButton {
+          halign: center;
+          label: _("Menu Button");
+          menu-model: demo_menu;
+        }
+      }
+
+      Adw.PreferencesGroup {
+        title: _("Context Menus");
+        description: _("Menus that popup on right-clicking a widget");
+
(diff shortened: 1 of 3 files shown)


src/Library/demos/Menu/main.blp
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  section {
+    label: _("Settings");
+
+    submenu {
+      label: _("Appearance");
+
+      item {
+        label: _("Force System Default");
```
andyholmes, line 93:
```suggestion
        label: _("System Default");
```

src/Library/demos/Menu/main.blp
```diff
@@ -0,0 +1,159 @@
...
+    action: "text.bold";
+  }
+
+  submenu {
+    label: _("Font Color");
+
+    item {
+      label: _("White");
+      action: "text.color";
+      target: "white";
```
andyholmes, line 144:
```suggestion
      label: _("Black");
      action: "text.color";
      target: "black";
```

Let's make the default `black`. The default makes it go nearly invisible, and I think that's a bit confusing.

andyholmes (changes requested):
Nice one.

A few teaks and I think this is good to go.

sonnyp:
@AkshayWarrier ping :)

src/Library/demos/Menu/main.blp
AkshayWarrier: Sorry for the late response 😅 

But I was running Workbench in dark mode so I completely forgot that white text would go invisible in light mode, maybe we could choose a color that's neither black nor white?

Maybe gray?

image
image

We could choose any other random color

andyholmes: How about just `green`? I hadn't thought of dark mode myself, so red/green/blue seems safe and legible for all variants.

andyholmes (changes requested):

andyholmes (approved):
No need to wait for a simple fix, since you were up late trying to get Workbench to build.

AkshayWarrier:
Thank you :)

PR #446, opened 2023-07-26
library: Port ListModel and List View Widgets demos to Vala

Diego-Ivan:
Ports ListModel and List View Widgets demos to Vala

### DIFF
--- a/src/Library/demos/List Model/main.vala
+++ b/src/Library/demos/List Model/main.vala
@@ -24,7 +24,7 @@ public void main () {
   );
 
   model.items_changed.connect ((position, removed, added) => {
-    message (@"position: $position, Item Removed? $((bool) removed), Item Added? $((bool) added)");
+    message (@"position: $position, Item Removed? $(removed > 0), Item Added? $(added > 0)");
   });
 
   // Filter Model
@@ -53,7 +53,7 @@ public void main () {
   });
 
   remove_button.clicked.connect (() => {
-    var selected_row = list_box_editable.get_selected_row ();
+    Gtk.ListBoxRow selected_row = list_box_editable.get_selected_row ();
     model.remove (selected_row.get_index ());
   });
 
@@ -80,17 +80,14 @@ private Gtk.Widget create_item_for_flow_box (Object item) {
   var string_object = (Gtk.StringObject) item;
   var label = new Gtk.Label (string_object.string) {
     halign = CENTER,
+    valign = CENTER,
     hexpand = true,
-    valign = CENTER
-  };
-
-  var label_bin = new Adw.Bin () {
-    child = label,
     width_request = 160,
     height_request = 160,
-    css_classes = { "card" }
+    css_classes = {"card"}
   };
-  return label_bin;
+
+  return label;
 }
 
 private Gtk.Widget create_item_for_filter_model (Object item) {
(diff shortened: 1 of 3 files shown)


src/Library/demos/List Model/main.vala
```diff
@@ -0,0 +1,103 @@
...
+  remove_button = (Gtk.Button) workbench.builder.get_object ("remove");
+
+  // Model
+  int item = 1;
+  var model = new Gtk.StringList (
+    {"Default Item 1", "Default Item 2", "Default Item 3"}
+  );
+
+  model.items_changed.connect ((position, removed, added) => {
+    message (@"position: $position, Item Removed? $((bool) removed), Item Added? $((bool) added)");
```
lw64, line 27: Please use a better way to check if removed/added is 0 or not. Vala is not C :)

src/Library/demos/List Model/main.vala
```diff
@@ -0,0 +1,103 @@
...
+    item++;
+  });
+
+  search_entry.search_changed.connect (() => {
+    string search_text = search_entry.text;
+    filter.search = search_text;
+  });
+
+  remove_button.clicked.connect (() => {
+    var selected_row = list_box_editable.get_selected_row ();
```
lw64, line 56: Use the actual type instead of var here please.

src/Library/demos/List Model/main.vala
```diff
@@ -0,0 +1,103 @@
...
+
+private Gtk.Widget create_item_for_flow_box (Object item) {
+  var string_object = (Gtk.StringObject) item;
+  var label = new Gtk.Label (string_object.string) {
+    halign = CENTER,
+    hexpand = true,
+    valign = CENTER
+  };
+
+  var label_bin = new Adw.Bin () {
```
lw64, line 87: I don't think putting the label into a Adw.Bin is not needed. You can put the properties and css class also on the label directly.

src/Library/demos/List View Widgets/main.vala
```diff
@@ -0,0 +1,71 @@
...
+  int item = 1;
+  var string_model = new Gtk.StringList (
+    {"Default Item 1", "Default Item 2", "Default Item 3"}
+  );
+  var model = new Gtk.SingleSelection (string_model);
+
+  var grid_view_factory = new Gtk.SignalListItemFactory ();
+
+  grid_view_factory.setup.connect ((list_item) => {
+    var list_box = new Gtk.Box (HORIZONTAL, 0) {
```
lw64, line 24: This Box is also not needed I think similar to the Adw.Bin

src/Library/demos/List View Widgets/main.vala
```diff
@@ -0,0 +1,71 @@
...
+    var list_box = (Gtk.Box) list_item.child;
+    var label_widget = (Gtk.Label) list_box.get_last_child ();
+
+    var model_item = (Gtk.StringObject) list_item.item;
+    label_widget.label = model_item.string;
+  });
+
+  // View
+  string_model.items_changed.connect ((position, removed, added) => {
+    message (@"position: $position, Item Removed? $((bool) removed), Item Added? $((bool) added)");
```
lw64, line 50: Same here for 0 checking

lw64 (changes requested):
A few changes are needed, but looks good otherwise.

lw64 (approved):

lw64:
Thank you!

PR #447, opened 2023-07-26
library: Add vala demo for Button

SpikedPaladin:
Adds Vala code demo for Button

### DIFF
--- a/src/Library/demos/Button/main.vala
+++ b/src/Library/demos/Button/main.vala
@@ -1,0 +1,22 @@
+#!/usr/bin/env -S vala workbench.vala --pkg gtk4 --pkg libadwaita-1
+
+public void main () {
+  string[] button_ids = {
+    "regular",
+    "flat",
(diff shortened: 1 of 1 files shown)


src/Library/demos/Button/main.vala
```diff
@@ -0,0 +1,22 @@
...
+    "custom",
+    "disabled",
+    "circular-plus",
+    "circular-minus",
+    "pill",
+    "osd-left",
+    "osd-right",
+  };
+
+  foreach (var id in button_ids) {
```
lw64, line 18: Use string instead of var here please.

lw64 (changes requested):
Thanks! Looks good.

src/Library/demos/Button/main.vala
SpikedPaladin: Fixed

sonnyp (approved):

PR #448, opened 2023-07-26
main.blp: add switch row

### LINKED ISSUE #410: Add SwitchRow entry
https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.SwitchRow.html

halfmexican:
closes #410


sonnyp (approved):
LGTM I don't think there is much more to it.

Would be great if we had a way to link a demo to another, here we would link to the `Switch` demo.

PR #449, opened 2023-07-26
window: Set current demo's name as window title

Diego-Ivan:
This PR implements a feature that sets the current demo's name as title of the window. While this is not visible in Workbench's UI, it is in GNOME Shell's activities overview and window switcher. This allows the user to find the demo they were working on more easily if they had multiple windows opened.

The name of the demo is also stored in the session's settings so it can be restored.

### DIFF
--- a/src/window.js
+++ b/src/window.js
@@ -56,6 +56,7 @@ export default function Window({ application, session }) {
     window.add_css_class("devel");
   }
   window.set_application(application);
+  window.set_title(`Workbench — ${settings.get_string("title")}`);
 
   // Popover menu theme switcher
   const button_menu = builder.get_object("button_menu");
--- a/data/app.gschema.xml
+++ b/data/app.gschema.xml
@@ -15,6 +15,9 @@
     </key>
   </schema>
   <schema id="@app_id@.Session" path="/re/sonny/Workbench/">
+    <key name="title" type="s">
+      <default>'Workbench'</default>
+    </key>
     <key name="show-code" type="b">
       <default>false</default>
     </key>
--- a/src/sessions.js
+++ b/src/sessions.js
@@ -62,6 +62,7 @@ export function createSessionFromDemo(demo) {
 
   const { panels } = demo;
   const { settings } = session;
+  settings.set_string("title", demo.name);
   settings.set_boolean("show-code", panels.includes("code"));
   settings.set_boolean("show-style", panels.includes("style"));
   settings.set_boolean("show-ui", panels.includes("ui"));


sonnyp:
This is a great idea :bulb: thanks!

One issue I see is that Workbench encourages users to continue working on the session until it has nothing to do with the original demo anymore. We would need a way for the user to rename the session for it to make sense. 

I don't really see a good way to do this now.

Thankfully the next thing I'm working on is "on-disk" projects which is letting the user choose where to save the session as a directory. Since they'll be able to pick the folder name, we could simply use that. Renaming the folder would rename the session.

Makes sense? and What do you think?

If you agree then we need to change this a bit to use the session folder name instead of adding a setting.

Diego-Ivan:
Sounds great! I think it makes more sense given Workbench's objective. I will be making the changes so that the window title matches the folder name :)

Diego-Ivan:
Hello! I implemented the changes so that the window's title is the base name of the session directory. Nonetheless, I tried to implement a `Gio.FileMonitor` to listen for the changes in name of the directory, but after some testing it did not seem to work. Let me know if I have to make further changes of this MR and I will try to implement whenever I can :smile:

src/window.js
```diff
@@ -57,6 +57,21 @@ export default function Window({ application, session }) {
   }
   window.set_application(application);
 
+  function updateTitle () {
+    window.set_title (`Workbench — ${session.file.get_basename ()}`);
+  }
+
+  // Monitor the session directory and update the window title if it is renamed
+  // FIXME: The monitor does not seem to listen for rename events
+  let monitor = session.file.monitor (Gio.FileMonitorFlags.WATCH_MOVES, null);
```
andyholmes, line 66: You need to hold a reference to this somewhere, or it will just be collected and the signal disconnected with it.

sonnyp: Let's remove the monitor for now.

Currently, if the project is moved or renamed while being open, it'll break anyway.

I have created a follow-up to address the problem more globally https://github.com/sonnyp/Workbench/issues/468

sonnyp (changes requested):
I have added save/open projects in https://github.com/sonnyp/Workbench/pull/466 

I think you can simply use `session.name` as title in `window.js`

Diego-Ivan:
Thanks for the reviews! I've rebased the branch and made the necessary changes :)

Diego-Ivan:
> Maybe we could use the Library entry as name? If the user opens twice the same Library entry, there will be 2 windows with the same name but I don't think it's a big deal and probably more of a edge case.

I think this makes sense. I'm only worried about restored sessions as they would have the same problem as what happens now with Library entries. I was thinking about saving the session name as a key in the gschema settings, so we can restore it later and override it in case the user wants to save the session as a project with a different name. The only side effect this may have is that restored sessions from a previous version of Workbench would have the `Workbench - 20xx- xx xx` title instead of the name of the demo entry, but oh well, I don't think there is a way to guess which demo it was supposed to be.

Diego-Ivan:
I implemented the schema key setting as I mentioned. User projects, newly opened demo entries and restored sessions from this Workbench branch work as they should, but as I said, restored sessions from previous versions use the `Workbench - 20xx- xx xx`  title, but I can't think of any way to solve this.

PR #450, opened 2023-07-29
library: port Spinner demo to Vala

mar-v-in:


sonnyp:
Thanks!

Please add your name to `about.js` in the contributors section :pray:

src/Library/demos/Spinner/main.vala
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env -S vala workbench.vala --pkg gtk4 --pkg libadwaita-1
+
+public void main () {
+  Gtk.init ();
+
+  var button = workbench.builder.get_object ("button") as Gtk.Button;
```
lw64, line 6: Please use here the (Gtk.Button) casting syntax.

lw64 (changes requested):
Looks good overall. Thanks!

sonnyp:
ping @mar-v-in :)

src/Library/demos/Spinner/main.vala
mar-v-in: Changed.

I also checked other examples in the Library, and about half of them are using checked casts for the result of `builder.get_object()` whereas the other half is using unchecked casts. I guess that should be unified to not cause unnecessary confusion to beginners.

lw64: hm. seems like I didnt review them (properly). That should be done in a separate PR though.

lw64 (approved):

PR #457, opened 2023-07-29
Documentation Viewer: Improved sidebar

AkshayWarrier:


AkshayWarrier:
Some things to consider
- There is no section for Function Macros like DevHelp and they are added with the Functions section because both files are named as ``func.xx.html`` 
- We could  try synchronizing the item selected in the sidebar and the doc shown by the webview? Sometimes the user may click on links on the doc that redirect them to some other page.

AkshayWarrier:
For some reason, I'm unable to change the default selected item in the sidebar when the doc viewer is first opened.

Scrolling in the web view is broken and also switching between sections in the documentation (when the sections are in the same page) potentially because of https://bugs.webkit.org/show_bug.cgi?id=260073

PR #458, opened 2023-07-30
Don't mention names in the fixme

Hofer-Julian:


sonnyp (approved):
I remember now, it was Alexander Larsson :)

Anyway, we shouldn't use an abstract socket because it requires network permission and lives outside of the sandbox.

Thanks!

Issue #406, opened 2023-07-16
Add OverlaySplitView entry


AkshayWarrier:
https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.OverlaySplitView.html

sonnyp:
See "Split Views -> Overlay Split View" in libadwaita demo (nightly)

PR #463, opened 2023-08-01
library: add Overlay Split View entry

halfmexican:
Fixes https://github.com/sonnyp/Workbench/issues/406


sonnyp (changes requested):
The window is too small and the sidebar should automatically show/hide depending on the window width.

Could you replicate **exactly** the "Adwaita Demo" example ?

I think it's what users want in 99% of cases.

halfmexican:
[overlay.webm](https://github.com/sonnyp/Workbench/assets/103920890/4292bcb7-6694-41cf-bf9e-58326c29587b)
The automatic hiding of the sidebar appears to work for me.

sonnyp (changes requested):
There were conflicts and unwanted changes in the diff.

I fixed it and merged `main` into your branch, please `git pull`.

So I don't know what's going on but here I get the following now:

[Screencast from 2023-08-09 21-09-20.webm](https://github.com/sonnyp/Workbench/assets/19673/c9279793-7424-4fa5-af62-9ed82a4af2c7)

Are you using the same method as "Adwaita Demo" to set the window size?

halfmexican:
Yes, I thought I forgot the tooltip text but it appears I did included that

sonnyp (approved):
Nice.

I removed `modal: true` because without set_transient_for it feels buggy. 

(Can't interact with Workbench window but there is no feedback as to why)

PR #464, opened 2023-08-01
feat: add D-Bus previewer to libworkbench

andyholmes:


Hofer-Julian:
This looks very useful

andyholmes:
I think this should be 100% functional, I just had some trouble with a namespace collision when trying to port the Vala example to use it.

If you want to give it a shot with Rust only `run()` needs to be implemented on the interface, although the other methods may be overridden if necessary.

Hofer-Julian:
> I think this should be 100% functional, I just had some trouble with a namespace collision when trying to port the Vala example to use it.
> 
> If you want to give it a shot with Rust only `run()` needs to be implemented on the interface, although the other methods may be overridden if necessary.

There's nothing functional with Rust yet. If you get the Vala examples to work, I think it's fine to merge this.

andyholmes:
It may be a while before I get back to this, so I'm going to close it up until I have to work through the conflicts.

Issue #428, opened 2023-07-17
Use relative paths for assets


sonnyp:
Now that we have sessions support and that the assets will be in the session folder, we can use relative paths.

This will make the code simpler, easier to port (to Vala for example) and will allow "important" files into projects.

PR #470, opened 2023-08-04
library: Use relative paths

sonnyp:
Fixes https://github.com/sonnyp/Workbench/issues/428 by adding a new `workbench.resolve('./image.png')` API which returns an absolute file URI.

---

Another way to solve it is that when running a session we can set `pwd` to the session folder and make use of `File.new_for_path('./image.png')` in examples.

One issue is that it can't be used with the "inline" JS preview/runner unless
  - each Workbench window becomes a separate process
  - or
  - we disable inline runner for JS and run in a separate process like Vala

We have a similar problem with the console, it prints stdout/stderr of the Workbench process which is shared between windows. 

I'm not sure yet if this is the right way to go. Since we already have a couple of `workbench.x` APIs, I don't think adding a new one is a huge deal. It might also make the migration path easier because `pwd` is error prone.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,8 +10,5 @@ trim_trailing_whitespace = true
 indent_style = space
 indent_size = 2
 
-[**.vala]
-indent_size = 4
-
 [{Makefile,**.mk}]
 indent_style = tab
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,5 +10,8 @@ trim_trailing_whitespace = true
 indent_style = space
 indent_size = 2
 
+[**.vala]
+indent_size = 4
+
 [{Makefile,**.mk}]
 indent_style = tab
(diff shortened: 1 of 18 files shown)


sonnyp:
What is important is that when using the example in a "real" GNOME app, there is an easy migration path.

In GJS it is [`import.meta.url`](https://gitlab.gnome.org/GNOME/gjs/-/blob/master/doc/ESModules.md#importmetaurl) and `import.meta.resolve` in the future.

Not sure about Vala, C, Rust, ... ?

lw64:
I guess an equivalent is `build_path(get_pwd(), rel_path).to_uri()` ?

sonnyp:
@lw64 `workbench.resolve` resolves relatively to the current session folder, not pwd. 

I suppose we'll have to tell users to use resource uris for assets when they move their code from Workbench to GNOME Builder.

sonnyp:
@andyholmes and @lw64 I updated the description for clarity.

PR #471, opened 2023-08-05
Remove application in workbench-api.vala

Hofer-Julian:
As far as I can tell, it is unused
@lw64


lw64:
The original idea was to use this for sending notifications, but the demo was never ported, because it didnt quite work 100%. I am not sure if this should be removed, but no strong opinion. @sonnyp ?

sonnyp (approved):
Let's remove it for now since it's not used.

PR #472, opened 2023-08-06
Rename workbench-api.vala to workbench.vala

Hofer-Julian:
This original rename in vala.js made it harder to see how the pieces are connected


andyholmes (approved):

andyholmes:
LGTM, I'll just cc @lw64 in case there's something I don't know.

lw64:
I think you also need to adjust the `#!/usr/bin/env ...` in all demos.

Hofer-Julian:
> I think you also need to adjust the `#!/usr/bin/env ...` in all demos.

I don't thinl so. They already all expect `workbench.vala`.

lw64:
I see, alright then.

lw64 (approved):

PR #476, opened 2023-08-07
library: Add vala demo for Text View

SpikedPaladin:
Adds Vala code demo for TextView

### DIFF
--- a/src/Library/demos/Text View/main.vala
+++ b/src/Library/demos/Text View/main.vala
@@ -1,9 +1,5 @@
 #!/usr/bin/env -S vala workbench.vala --pkg gtk4 --pkg libadwaita-1
 
-Gtk.TextBuffer buffer;
-Gdk.Display display;
-Gdk.Clipboard clipboard;
-
 void main () {
   var textview = (Gtk.TextView) workbench.builder.get_object("textview");
   var select = (Gtk.Button) workbench.builder.get_object("select");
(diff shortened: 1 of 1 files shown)


src/Library/demos/Text View/main.vala
```diff
@@ -0,0 +1,76 @@
...
+  clear.clicked.connect(() => {
+    Gtk.TextIter start, end;
+
+    buffer.get_start_iter(out start);
+    buffer.get_end_iter(out end);
+
+    buffer.@delete(ref start, ref end);
+  });
+
+  textview.copy_clipboard.connect(() => {
```
lw64, line 44: If there is only one function call the `{ }` can be left out:
`textview.copy_clipboard.connect(() => message (""));`

src/Library/demos/Text View/main.vala
```diff
@@ -0,0 +1,76 @@
...
+  });
+  textview.cut_clipboard.connect(() => {
+    message("Text cut to clipboard");
+  });
+  textview.paste_clipboard.connect(() => {
+    message("Text pasted");
+  });
+}
+
+void handleCopy() {
```
lw64, line 55: Why are these actual functions, and not like the signal handlers above lambdas?

src/Library/demos/Text View/main.vala
```diff
@@ -0,0 +1,76 @@
...
+  if (buffer.get_has_selection()) {
+    buffer.copy_clipboard(clipboard);
```
lw64, line 57: Always use properties if they are available. In this case `if (buffer.has_selection) {`

lw64 (changes requested):
Thanks for the port. Looks correct except the few places I left a comment.

lw64:
@SpikedPaladin there is still one request for changes unresolved, just for you to note

lw64 (approved):

PR #485, opened 2023-08-11
library: Adds Navigation View Entry

### LINKED ISSUE #404: Add NavigationView entry
https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.NavigationView.html

SoNiC-HeRE:
Closes #404

### DIFF
--- a/src/Library/demos/Navigation View/main.blp
+++ b/src/Library/demos/Navigation View/main.blp
@@ -24,7 +24,7 @@ Adw.StatusPage {
         icon-name: "go-next-symbolic";
         styles ["circular", "suggested-action"]
       }
-    }
+    }    
     Box {
       height-request: 360;
       width-request: 360;
@@ -33,27 +33,22 @@ Adw.StatusPage {
       styles ["card"]
 
       Adw.NavigationView nav_view {
-        animate-transitions: false;
         Adw.NavigationPage nav_pageone {
-          title: bind title.label;
           Adw.StatusPage {
             title: _("Hello");
           }
         }
         Adw.NavigationPage nav_pagetwo {
-          title: bind title.label;
           Adw.StatusPage {
             title: _("from");
           }
         }
         Adw.NavigationPage nav_pagethree {
-          title: bind title.label;
           Adw.StatusPage {
             title: _("Workbench :)");
           }
         }
         Adw.NavigationPage nav_pagefour {
-          title: bind title.label;
           Adw.StatusPage {
             title: _("Navigation View");
             description: _("A page-based navigation container.");
@@ -71,13 +66,14 @@ Adw.StatusPage {
       }
     }
     Adw.PreferencesGroup{
-      styles ["boxed-list"]
       Adw.SwitchRow decisive_button_transition {
-        title: _("Animate transitions");
+        title: _("Animate Transitions");
+        active: bind nav_view.animate-transitions bidirectional;
       }
       Adw.SwitchRow decisive_button_poponescape {
         title: _("Pop on Escape");
-      }
+        active: bind nav_view.pop-on-escape bidirectional;
+      }      
     }
   }
 }
--- a/src/Library/demos/Navigation View/main.blp
+++ b/src/Library/demos/Navigation View/main.blp
@@ -10,14 +10,19 @@ Adw.StatusPage {
     CenterBox {
       [start]
       Button previous_button {
-        icon-name: _("go-previous-symbolic");
+        icon-name: "go-previous-symbolic";
         sensitive: false;
-        styles ["circular","suggested-action"]
+        styles ["circular", "suggested-action"]
+      }
+      [center]
+      Label title {
+        label: _("Page 1");
+        styles ["title-4"]
       }
       [end]
       Button next_button {
-        icon-name: _("go-next-symbolic");
-        styles ["circular","suggested-action"]
+        icon-name: "go-next-symbolic";
+        styles ["circular", "suggested-action"]
       }
     }
     Box {
@@ -30,28 +35,25 @@ Adw.StatusPage {
       Adw.NavigationView nav_view {
         animate-transitions: false;
         Adw.NavigationPage nav_pageone {
-          title: _("Page 1");
+          title: bind title.label;
           Adw.StatusPage {
             title: _("Hello");
-            description: _("page 1");
           }
         }
         Adw.NavigationPage nav_pagetwo {
-          title: _("Page 2");
+          title: bind title.label;
           Adw.StatusPage {
             title: _("from");
-            description: _("page 2");
           }
         }
         Adw.NavigationPage nav_pagethree {
-          title: _("Page 3");
+          title: bind title.label;
           Adw.StatusPage {
             title: _("Workbench :)");
-            description: _("page 3");
           }
         }
         Adw.NavigationPage nav_pagefour {
-          title: _("Page 4");
+          title: bind title.label;
           Adw.StatusPage {
             title: _("Navigation View");
             description: _("A page-based navigation container.");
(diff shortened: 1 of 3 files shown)


SoNiC-HeRE:
[Kooha-2023-08-11-06-12-23.webm](https://github.com/sonnyp/Workbench/assets/96797205/96871f2b-2c4f-441f-a80f-1b077778d87b) 

## Looking for an early review

SoNiC-HeRE:
Currently implemented with
- Default Shortcuts
- No console output
- No Header Bar (tried but makes the design inconsistent )

src/Library/demos/Navigation View/main.blp
```diff
@@ -0,0 +1,90 @@
...
+            label: _("Yes");
+            active: true;
+          }
+          ToggleButton no_button {
+            valign: center;
+            label: _("No");
+            group: yes_button;
+          }
+        }
+      }
```
andyholmes, line 87: This should just be a `Adw.SwitchRow`, bound to the property.

src/Library/demos/Navigation View/main.blp
```diff
@@ -0,0 +1,90 @@
...
+      margin-bottom: 24;
+      styles ["linked"]
+      Button previous_button {
+        label: _("Previous Page");
+        sensitive: false;
+      }
+      Button next_button {
+        label: _("Next Page");
+      }
+    }
```
andyholmes, line 68: This should probably be above the view widget, and I'd suggest using `circular` & `suggested-action` styles, with symbolic icons like `go-previous-symbolic` & `go-next-symbolic`, set in `Gtk.CenterBox` with the buttons in separate sides.

For the "previous" button, should be able to just use `action-name: navigation.pop;`, while the other button would probably need a signal callback or custom action.

I'm not sure how @sonny wants this demo to look in general, but that would be a much more HIG way to do it.

src/Library/demos/Navigation View/main.blp
```diff
@@ -0,0 +1,90 @@
...
+      margin-bottom: 24;
+      orientation: vertical;
+      styles ["card"]
+
+      Adw.NavigationView nav_view {
+        Adw.NavigationPage nav_pageone {
+          title: _("Page 1");
+          Adw.StatusPage {
+            title: _("Hello 👋");
+            description: _("➡️");
```
andyholmes, line 22: These should probably just be standard labels (of some sort), especially considering this widget is probably sensitive to LTR vs RTL language locales.

src/Library/demos/Navigation View/main.blp
```diff
@@ -0,0 +1,90 @@
...
+Adw.StatusPage {
+  Box {
+    halign: center;
+    valign: start;
+    orientation: vertical;
+    Box {
+      height-request: 360;
+      width-request: 360;
+      halign: center;
+      margin-bottom: 24;
```
andyholmes, line 13:
```suggestion
      spacing: 24;
```

You can always use `spacing`, even if there aren't widget below and above.

andyholmes (changes requested):
Early review as requested :slightly_smiling_face:

src/Library/demos/Navigation View/main.blp
SoNiC-HeRE: image

Looks more clean and better

src/Library/demos/Navigation View/main.js
```diff
@@ -0,0 +1,38 @@
...
+});
+
+previous_button.connect("clicked", () => {
+  nav_view.pop();
+  next_button.sensitive = true;
+  if (nav_view.get_visible_page() === nav_pageone) {
+    previous_button.sensitive = false;
+  }
+});
+
```
andyholmes, line 35: I think it would be better to use `notify::visible-page` on the view to set the sensitivity. Something like `previous_button.sensitive = nav_view.visible_page !== nav_pageone` and so on.

src/Library/demos/Navigation View/main.js
```diff
@@ -0,0 +1,38 @@
...
+next_button.connect("clicked", () => {
+  previous_button.sensitive = true;
+  if (nav_view.get_visible_page() === nav_pageone) {
+    nav_view.push(nav_pagetwo);
+  } else if (nav_view.get_visible_page() === nav_pagetwo) {
+    nav_view.push(nav_pagethree);
+  } else if (nav_view.get_visible_page() === nav_pagethree) {
+    nav_view.push(nav_pagefour);
+    next_button.sensitive = false;
+  }
```
andyholmes, line 25: This would make more sense as a `switch` statement, also remember to use properties instead of methods when possible.

andyholmes (changes requested):
Some more notes :)

src/Library/demos/Navigation View/main.js
```diff
@@ -0,0 +1,38 @@
...
+  nav_view.pop();
+  next_button.sensitive = true;
+  if (nav_view.get_visible_page() === nav_pageone) {
+    previous_button.sensitive = false;
+  }
+});
+
+decisive_button.connect("notify::active", () => {
+  nav_view.animate_transitions = decisive_button.active;
+});
```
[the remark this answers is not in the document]
SoNiC-HeRE: Did you mean to remove the signal handler or something like ```nav_view.animate_transitions = true```?
I am already using property instead of method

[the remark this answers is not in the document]
andyholmes: I mean:

```js
      Adw.SwitchRow decisive_button {
        title: _("Animate Transitions");
        active: bind nav_view.animate-transitions;
      }
```

src/Library/demos/Navigation View/main.blp
```diff
@@ -0,0 +1,83 @@
...
+              }
+            }
+          }
+        }
+      }
+    }
+    Adw.PreferencesGroup{
+      styles ["boxed-list"]
+      Adw.SwitchRow decisive_button_transition {
+        title: _("Animate transitions");
```
andyholmes, line 76:
```suggestion
        title: _("Animate Transitions");
```

src/Library/demos/Navigation View/main.blp
```diff
@@ -0,0 +1,83 @@
...
+    }
+    Box {
+      height-request: 360;
+      width-request: 360;
+      halign: center;
+      orientation: vertical;
+      styles ["card"]
+
+      Adw.NavigationView nav_view {
+        animate-transitions: false;
```
andyholmes, line 36:
```suggestion
```

andyholmes (changes requested):

src/Library/demos/Navigation View/main.js
```diff
@@ -0,0 +1,51 @@
...
+    case nav_pagetwo:
+      title.label = "Page 2";
+      break;
+    case nav_pagethree:
+      title.label = "Page 3";
+      break;
+    case nav_pagefour:
+      title.label = "Page 4";
+      break;
+  }
```
andyholmes, line 50:
```suggestion
  title.label = nav_view.visible_page.title;
```

The page title should be part of template, then you can just set the `title` widget here. This is what the `Adw.HeaderBar` basically does.

src/Library/demos/Navigation View/main.js
```diff
@@ -0,0 +1,51 @@
...
+const previous_button = workbench.builder.get_object("previous_button");
+const nav_pagetwo = workbench.builder.get_object("nav_pagetwo");
+const nav_pagethree = workbench.builder.get_object("nav_pagethree");
+const nav_pagefour = workbench.builder.get_object("nav_pagefour");
+const decisive_button_transition = workbench.builder.get_object(
+  "decisive_button_transition",
+);
+const decisive_button_poponescape = workbench.builder.get_object(
+  "decisive_button_poponescape",
+);
```
andyholmes, line 13:
```suggestion
```

andyholmes (changes requested):
Just a few more changes, then it looks good to me.

andyholmes (approved):
LGTM

PR #486, opened 2023-08-11
library: Port Event Controllers demo to Vala

Diego-Ivan:
Ports the Event Controllers demo to Vala

### DIFF
--- a/src/Library/demos/Event Controllers/main.vala
+++ b/src/Library/demos/Event Controllers/main.vala
@@ -1,0 +1,90 @@
+#!/usr/bin/env -S vala workbench.vala --pkg libadwaita-1
+
+private bool ctrl_pressed = false;
+
+public void main () {
+  // Gtk.Window hides Gtk.Widget's add_controller method, thus we need to access it as a Widget.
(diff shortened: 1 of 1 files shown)


src/Library/demos/Event Controllers/main.vala
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env -S vala workbench.vala --pkg libadwaita-1
+
+private bool ctrl_pressed = false;
+
+public void main () {
+  // Gtk.Window hides Gtk.Widget's add_controller method, thus we need to access it as a Widget.
+  Gtk.Widget window = workbench.window;
```
lw64, line 7: use `Gtk.Window` and cast it to `Gtk.Widget` when calling the add_controller method

lw64, line 7: use `Gtk.Window` and cast it to `Gtk.Widget` when calling the add_controller method. A comment there would be appropriate

lw64 (changes requested):
Looks good!

lw64 (approved):

lw64:
Thanks!

PR #487, opened 2023-08-11
library: Adds toolbarview entry

### LINKED ISSUE #408: Add ToolbarView entry
https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.ToolbarView.html

SoNiC-HeRE:
Closes #408

### DIFF
--- a/src/Library/demos/Toolbar View/main.blp
+++ b/src/Library/demos/Toolbar View/main.blp
@@ -1,0 +1,116 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  Box {
+    halign: center;
+    orientation: vertical;
+    spacing: 36;
+    Box {
+      halign: center;
+      Adw.PreferencesGroup {
+        styles ["boxed-list"]
+        Adw.ComboRow topbar_select {
+          title: _("Top Bar :");
+          model: StringList {
+            strings ["Header Bar","Tab Bar","View Switcher Bar","Action Bar","PopOverMenu Bar","Search Bar","Gtk Box"]
+          };
+        }
+        Adw.ComboRow bottombar_select{
+          title: _("Bottom Bar :");
+          model: StringList {
+            strings ["Header Bar","Tab Bar","View Switcher Bar","Action Bar","PopOverMenu Bar","Search Bar","Gtk Box"]
+          };
+        }
+        Adw.ComboRow barstyle_select{
+          title: _("Bar Style:");
+          model: StringList {
+            strings ["Flat","Raised","Raised-Border"]
+          };
+        }
+      }
+    }
+    Box {
+      height-request: 300;
+      width-request: 300;
+      halign: center;
+      orientation: vertical;
+      styles ["card"]
+
+      Adw.ToolbarView toolbar_view {
+        vexpand:true;
+        extend-content-to-bottom-edge: true;
+        extend-content-to-top-edge: true;
+        top-bar-style: bind barstyle_select.selected;
+        bottom-bar-style: bind barstyle_select.selected;
+        reveal-top-bars: false;
+        reveal-bottom-bars: false;
+        [top]
+        Adw.HeaderBar header_top {
+          title-widget: Label{
+            label:_("Header Bar");
+          };
+        }
+        content: Adw.StatusPage {
+          Box {
+            halign: center;
+            orientation: vertical;
+            spacing: 6;
+            Label{
+              label: _("Toolbar View");
+              styles ["title-1"]
+            }
+            Label{
+              wrap: true;
+              label: _("A widget containing a page, as well as top and/or bottom bars.");
+            }
+            LinkButton{
+              label: _("API Reference");
+              uri: "https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.ToolbarView.html";
+            }
+          }
+        };
+        [bottom]
+        Adw.HeaderBar header_bottom {
+          title-widget: Label{
+            label: _("Header Bar");
+          };
+        }
+      }
+    }
+    Box {
+      halign: center;
+      spacing: 12;
+      ToggleButton reveal_topbar{
+        label: _("Reveal Top Bar");
+        styles ["pill"]
+      }
+      ToggleButton reveal_bottombar{
+        label: _("Reveal Bottom Bar");
+        styles ["pill"]
+      }
+    }
+  }
+}
+  ActionBar action_bar {
+    revealed: true;
+    valign: end;
+
+    [start]
+    Button start_widget {
+      icon-name: "call-start-symbolic";
+    }
+
+    [center]
+    DropDown {
+      model: StringList {
+        strings ["Center Widget","👁️", "❤️", "💼", "🪑"]
+        };
+    }
+
+    [end]
+    Button end_widget {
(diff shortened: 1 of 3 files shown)


SoNiC-HeRE:
[Kooha-2023-08-15-20-13-43.webm](https://github.com/sonnyp/Workbench/assets/96797205/2a644605-66da-4ba0-aedb-45a25eabdba3)

andyholmes (changes requested):
This one needs a fair amount of cleanup to the UI.

The first thing I would do is move the boxed list and reveal buttons into the `Adw.ToolbarView`. The list rows labels also need to be updated to use labels following the HIG.

I'm thinking this one might need to be in a proper window, because the borders don't look quite right, and the close buttons actually try to close the Workbench window.

sonnyp (commented):
Feedback from alice

> why does the header bar have a GtkLabel in it?

src/Library/demos/Toolbar View/main.blp
```diff
@@ -0,0 +1,220 @@
...
+    bottom-bar-style: bind barstyle_select.selected;
+    reveal-top-bars: false;
+    reveal-bottom-bars: false;
+    [top]
+    Adw.HeaderBar header_top {
+      title-widget: Label{
+      label:_("Header Bar");
+      };
+    }
+    content: Adw.StatusPage {
```
andyholmes, line 24: Switch this to an `Adw.ViewStack`

src/Library/demos/Toolbar View/main.blp
```diff
@@ -0,0 +1,220 @@
...
+          styles ["title-1"]
+        }
+        Label{
+          wrap: true;
+          label: _("A widget containing a page, as well as top and/or bottom bars.");
+        }
+        LinkButton{
+          label: _("API Reference");
+          uri: "https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.ToolbarView.html";
+        }
```
andyholmes, line 40: Put this in an `Adw.ViewStackPage` and `Adw.StatusPage`.

src/Library/demos/Toolbar View/main.blp
```diff
@@ -0,0 +1,220 @@
...
+        Box {
```
andyholmes, line 41: Change this to an `Adw.PreferencesPage` (in an `Adw.ViewStackPage`), so the `Adw.PreferencesGroup` works correctly.

src/Library/demos/Toolbar View/main.blp
```diff
@@ -0,0 +1,220 @@
...
+          halign: center;
+          Adw.PreferencesGroup {
+            styles ["boxed-list"]
+            Adw.ComboRow topbar_select {
+              title: _("Top Bar :");
```
andyholmes, line 46: Fix all these row titles to follow the HIG for [Writing Style](https://developer.gnome.org/hig/guidelines/writing-style.html).

src/Library/demos/Toolbar View/main.blp
```diff
@@ -0,0 +1,220 @@
...
+              model: StringList {
+                strings ["Header Bar","Tab Bar","View Switcher Bar","Action Bar","PopOverMenu Bar","Search Bar","Gtk Box"]
```
andyholmes, line 48: I'm not sure if we're going to keep these, but they should be marked for `_("Translation")`, if so.

src/Library/demos/Toolbar View/main.blp
```diff
@@ -0,0 +1,220 @@
...
+          halign: center;
+          spacing: 12;
+          ToggleButton reveal_topbar{
+            label: _("Reveal Top Bar");
+            styles ["pill"]
+          }
+          ToggleButton reveal_bottombar{
+            label: _("Reveal Bottom Bar");
+            styles ["pill"]
+          }
```
andyholmes, line 75: Let's make these `active: true` by default.

src/Library/demos/Toolbar View/main.blp
```diff
@@ -0,0 +1,220 @@
...
+        }
+      }
+    };
+    [bottom]
+    Adw.HeaderBar header_bottom {
```
andyholmes, line 80: Let's make this an `Adw.ViewSwitcher`, bound to the `Adw.ViewStack` that will be the main child of the `Adw.ToolbarView`.

src/Library/demos/Toolbar View/main.js
```diff
@@ -0,0 +1,79 @@
...
+    changeHeaderBar(workbench.builder.get_object("switcher_bar"));
+  } else if (topbar_select.get_selected() === 4) {
+    changeHeaderBar(workbench.builder.get_object("popover"));
+  } else if (topbar_select.get_selected() === 5) {
+    changeHeaderBar(workbench.builder.get_object("search_bar"));
+  } else if (topbar_select.get_selected() === 6) {
+    changeHeaderBar(workbench.builder.get_object("gtk_box"));
+  } else if (topbar_select.get_selected() === 1) {
+    changeHeaderBar(workbench.builder.get_object("tab_bar"));
+  }
```
andyholmes, line 42: Not sure if we'll keep this, but anything like this should use a `switch` statement.

andyholmes (changes requested):
More notes :slightly_smiling_face:

src/Library/demos/Toolbar View/main.blp
sonnyp: We are keeping them, it was part of the specs.

https://github.com/sonnyp/Workbench/issues/408#issuecomment-1637150356

src/Library/demos/Toolbar View/main.blp
sonnyp: Why? there is nothing to switch to

andyholmes: Sorry, I'm confused now. I thought you wanted to demonstrate all the different bars? Unless I'm mistaken you need an `Adw.ViewStack` for the `Adw.ViewSwitcherBar` to work?

andyholmes: Ah, I guess the idea is to have that other view switcher in the `.blp` file, along with the other widget? I guess I'm confused why not just actually use a view switcher?

SoNiC-HeRE: [Kooha-2023-08-17-18-11-54.webm](https://github.com/sonnyp/Workbench/assets/96797205/b7de120e-6368-455a-bebc-8c99ff57e980)

@sonnyp @andyholmes  Is this what we wanted to achieve?

src/Library/demos/Toolbar View/main.js
SoNiC-HeRE: Changed to switch statement

src/Library/demos/Toolbar View/main.blp
SoNiC-HeRE: I'm not sure if i should do it or not since I want to dynamically change the widgets instead of switching to a new view; but if you'd want it to be that way and seems better I'll definitely do it.Lemme know what you think of this

andyholmes: I'm not sure what the best approach is here, I think I misunderstood the scope of the demo :confused:

PR #499, opened 2023-08-15
Boxed Lists: show off activatable widget property.

halfmexican:
Super simple addition, with the release of the new 1.4 row widgets, we are not showing new users how to use ```activatable-widget``` property which is useful for deriving your own custom row widget and etc.


andyholmes (approved):
:+1:

Issue #497, opened 2023-08-13
Session inhibit


sonnyp:
https://libportal.org/method.Portal.session_inhibit.html

PR #500, opened 2023-08-16
library: Add Session Monitor entry

### LINKED ISSUE #496: Session monitor
https://libportal.org/method.Portal.session_monitor_start.html

AkshayWarrier:
Closes #496 and #497

### DIFF
--- a/src/Library/demos/Session Monitor/main.blp
+++ b/src/Library/demos/Session Monitor/main.blp
@@ -1,0 +1,42 @@
+using Gtk 4.0;
+using Adw 1;
+
+Adw.StatusPage {
+  title: _("Session Monitor");
+  description: _("Monitor the login session state");
+
+  Adw.Clamp {
+    maximum-size: 340;
+    
+    Box {
+      spacing: 12;
+      orientation: vertical;
+
+      Adw.EntryRow entry {
+        title: "Reason";
+        styles ["card"]
+      }
+    
+      Box {
+        spacing: 18;
+        homogeneous: true;
+
+        Button button_start {
+          label: _("Start Session");
+          styles ["suggested-action", "pill"]
+        }
+
+        Button button_stop {
+          label: _("Stop Session");
+          sensitive: false;
+          styles ["destructive-action", "pill"]
+        }
+      }
+
+      LinkButton {
+        label: "API Reference";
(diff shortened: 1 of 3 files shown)


AkshayWarrier:
I have tried to replicate ASPHD demo. I tested the session inhibits for ``LOGOUT`` and ``IDLE`` and they seem to work fine, but I'm not sure how ``USER_SWITCH`` and ``SUSPEND`` is supposed to work. Whenever I try to suspend or switch user, nothing happens. I even tried comparing it to ASPHD Demo and nothing happens there either.

Also, I'm activating the ``session_inihibit``(s) as soon as the session starts as opposed to when ``session-state-change`` is triggered. I'm not sure what's the correct method but when I try to do the latter, the inhibition always happens late and I have to logout a second time to see it.

sonnyp:
> nothing happens there either.

It might simply not be implemented in GNOME? 

> Also, I'm activating the session_inihibit(s) as soon as the session starts as opposed to when session-state-change is triggered. I'm not sure what's the correct method but when I try to do the latter, the inhibition always happens late and I have to logout a second time to see it.

Why would you wait for a session state change?

AkshayWarrier:
> Why would you wait for a session state change?

It's because in multiple places the doc mentioned something like that.
https://libportal.org/enum.LoginSessionState.html
> XDP_LOGIN_SESSION_QUERY_END  The session is in the query end phase, during which applications can save their state or inhibit the session from ending.

https://libportal.org/method.Portal.session_monitor_query_end_response.html

> This method should be called within one second of receiving a [XdpPortal::session-state-changed](https://libportal.org/signal.Portal.session-state-changed.html) signal with the ‘Query End’ state, to acknowledge that they have handled the state change.
> Possible ways to handle the state change are either to call [xdp_portal_session_inhibit()](https://libportal.org/method.Portal.session_inhibit.html) to prevent the session from ending, or to save your state and get ready for the session to end.

sonnyp:
Ha, I thought it was about something else. 

Hmm if ASHPD also exhibits the same problem then maybe it's a bug in the xdg portal ?

Could you open an issue on https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/issues with steps to reproduce using ASHPD?

PR #508, opened 2023-08-18
library: Add Memory Monitor entry

### LINKED ISSUE #455: Memory Monitor demo
https://docs.gtk.org/gio/iface.MemoryMonitor.html

See also @pwithnall talk https://www.youtube.com/watch?v=WVWrllJQJ_s&t=18094s

AkshayWarrier:
Closes #455

### DIFF
--- a/src/Library/demos/Memory Monitor/main.js
+++ b/src/Library/demos/Memory Monitor/main.js
@@ -1,4 +1,3 @@
-import Adw from "gi://Adw";
 import Gio from "gi://Gio";
 
 const memory_monitor = Gio.MemoryMonitor.dup_default();
@@ -9,6 +8,7 @@ cache.set("b", 2);
 cache.set("c", 3);
 
 memory_monitor.connect("low-memory-warning", (monitor, level) => {
+  // Use inequalities for checking as new levels may be added in the future
   if (level >= Gio.MemoryMonitorWarningLevel.LOW) {
     // Processes should free up unneeded resources
     console.log("Warning Level: Low");
--- a/src/Library/demos/Memory Monitor/main.js
+++ b/src/Library/demos/Memory Monitor/main.js
@@ -1,0 +1,27 @@
+import Adw from "gi://Adw";
+import Gio from "gi://Gio";
+
+const memory_monitor = Gio.MemoryMonitor.dup_default();
+
+const cache = new Map();
+cache.set("a", 1);
+cache.set("b", 2);
+cache.set("c", 3);
+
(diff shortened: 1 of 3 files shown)


src/Library/demos/Memory Monitor/main.js
```diff
@@ -0,0 +1,27 @@
+import Adw from "gi://Adw";
```
andyholmes, line 1:
```suggestion
```

Doesn't seem used here

src/Library/demos/Memory Monitor/main.js
```diff
@@ -0,0 +1,27 @@
...
+
+const memory_monitor = Gio.MemoryMonitor.dup_default();
+
+const cache = new Map();
+cache.set("a", 1);
+cache.set("b", 2);
+cache.set("c", 3);
+
+memory_monitor.connect("low-memory-warning", (monitor, level) => {
+  if (level >= Gio.MemoryMonitorWarningLevel.LOW) {
```
andyholmes, line 12: I think there should be a comment explaining why these are using a `>=` comparisons, since that's unexpected for an enumeration.

andyholmes (changes requested):
Looks good :slightly_smiling_face:

andyholmes (approved):
Nice one!

src/Library/demos/Memory Monitor/main.js
```diff
@@ -0,0 +1,27 @@
...
+cache.set("b", 2);
+cache.set("c", 3);
+
+memory_monitor.connect("low-memory-warning", (monitor, level) => {
+  // Use inequalities for checking as new levels may be added in the future
+  if (level >= Gio.MemoryMonitorWarningLevel.LOW) {
+    // Processes should free up unneeded resources
+    console.log("Warning Level: Low");
+    drop_caches();
+  } else if (level >= Gio.MemoryMonitorWarningLevel.MEDIUM) {
```
pwithnall, line 16: This example is great overall! One comment: this should perhaps be a separate `if` rather than an `else if`, otherwise `drop_caches()` will only be called on levels between `LOW` and `MEDIUM`, rather than on all levels `>= LOW`.

AkshayWarrier: Actually wait, now that I think about it, won't it behave in a completely oppositive manner? For all levels ``>= LOW``, the first ``if`` condition would always be satisfied and it would never get to the other ``else if``(s) even if ``level >= MEDIUM``. I can't believe I missed this but still changing the ``else if`` to ``if`` should still fix the issue regardless.

PR #521, opened 2023-08-22
Manuals: Add searching of docs

AkshayWarrier:


AkshayWarrier:
I completely missed this but I found the problem, it should be fixed now :)

PR #525, opened 2023-08-23
manuals: Hide sidebar using CSS

sonnyp:
Reference: 

* https://gitlab.gnome.org/GNOME/devhelp/-/blob/0d6cfe1f08e365745b357accde43c21b4308f842/devhelp/dh-web-view.c#L71
* https://gitlab.gnome.org/GNOME/devhelp/-/blob/0d6cfe1f08e365745b357accde43c21b4308f842/devhelp/dh-web-view.c#L502


AkshayWarrier (approved):
LGTM!

PR #529, opened 2023-08-25
Split file dialog entry

### LINKED ISSUE #514: Split the File Dialog entry
The File Dialog has 3 different features. "Open file(s)", "Select Folder(s)" and "Save File". 

From GTK perspective it makes sense to use the same class. But for user it's not a good fit to have all 3 in the same demo because usually you want to do one of the 3.

Having a demo for each would make discoverability easier and we have a bit of space to improve the example.

Todo

* Split the existing "File Dialog" entry into 3 separates
  * [ ] "Open File(s)"
  * [ ] "Select Folder(s)"
  * [ ] "Save File"

halfmexican:
closes #514


andyholmes (approved):
Nicely done, LGTM.

PR #530, opened 2023-08-25
Add AboutWindow demo for Rust

Hofer-Julian:


Hofer-Julian:
Just checked, seems like main is protected.
So I don't see what's the alternative to PRs :)

PR #532, opened 2023-08-26
library: add HTTP Request entry

### LINKED ISSUE #515: HTTP Request
We should have a demo to make an HTTP request to some API

You can use the Wikipedia API to get the featured article of the day

For example: https://api.wikimedia.org/feed/v1/wikipedia/en/featured/2023/03/04

See https://api.wikimedia.org/wiki/Getting_featured_content_from_Wikipedia_with_Python

(No need to display the image, we already have a demo for that)

halfmexican:
closes #515


halfmexican:
I think the description and preview could be better, but it works

PR #533, opened 2023-08-26
library: Fix FileDialog demos

sonnyp:
All 3 were broken.
Typos broke Code in 2 of them and the third one was in the wrong location.

The good news is that we now have 100 demos \o/


andyholmes:
Sorry! That was my mistake :/

PR #534, opened 2023-08-27
library: Add accessibility demo

### LINKED ISSUE #498: Basic example of custom widget supporting a11y
Adding accessibility support throughout the demos would be nice, but it would distract from the topic of the demos.

Instead, we should add a reasonably good example of implementing accessibility for a custom widget. The widget should demonstrate roles, relationships and dynamic states.

andyholmes:
closes #498


andyholmes:
Since I dropped a link in #Accessibility and no one had any comments, I'd say this one is probably good to merge.

PR #535, opened 2023-08-27
Add Account demo

Hofer-Julian:


src/langs/rust/template/Cargo.toml
```diff
@@ -6,7 +6,10 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+ashpd = "0.6.2"
 libc = "0.2"
+url = "2.4.0"
+
```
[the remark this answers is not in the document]
Hofer-Julian: > What's the common pratice in Rust GTK apps ?

I am not sure to be honest.
But speaking of it, should ashpd maybe re-export `url`? @bilelmoussaoui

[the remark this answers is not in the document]
bilelmoussaoui: > shouldn't we use GLib?

ashpd doesn't use glib at all and rust libraries uses the url crate as that is what integrates with zbus/zvariant.  I am fine with re-exporting url though.

src/Library/demos/Account/code.rs
```diff
@@ -0,0 +1,45 @@
...
+            }
+        });
+    });
+}
+
+async fn request_user_information() -> Result<(), Box<dyn Error>> {
+    let entry: adw::EntryRow = workbench::builder().object("entry").unwrap();
+
+    let response = UserInformation::request()
+        .reason(entry.text().as_str())
```
bilelmoussaoui, line 24: Do you actually have to call as_str btw?

bilelmoussaoui: also i am sad you don't use window identifier :P

src/Library/demos/Wallpaper/code.rs
```diff
@@ -0,0 +1,36 @@
...
+
+async fn set_wallpaper() -> Result<(), Box<dyn Error>> {
+    let uri = url::Url::parse(&workbench::resolve("./wallpaper.png"))?;
+    let request = WallpaperRequest::default()
+        .set_on(SetOn::Both)
+        .show_preview(true)
+        .build_uri(&uri)
+        .await?;
+
+    if request.response().is_ok() {
```
bilelmoussaoui, line 29: this is not great as you are now missing the response related errors :) so just call resquest.response()?;

src/langs/rust/template/Cargo.toml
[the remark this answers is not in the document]
bilelmoussaoui: Also as you already depend on gtk4, maybe you could enable the gtk4 feature of ashpd

src/Library/demos/Color Picker/color.rs
```diff
@@ -0,0 +1,30 @@
...
+                eprintln!("Could not pick color: {err}")
+            }
+        });
+    });
+}
+
+async fn pick_color() -> ashpd::Result<()> {
+    let color = Color::request().send().await?.response()?;
+
+    let color = gdk::RGBA::new(
```
bilelmoussaoui, line 21: If you enable the gtk4 feature, you can color `gdk::RGBA::from(color)`

src/Library/demos/Account/code.rs
Hofer-Julian: > Do you actually have to call as_str btw?

Yes, that is needed

> also i am sad you don't use window identifier :P

Good point, fixed

Hofer-Julian:
Thanks for the comments @bilelmoussaoui!

@sonnyp feel free to have another look, from my side this is ready to be merged

PR #542, opened 2023-08-28
Add codeowners file

Hofer-Julian:
See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

This way @lw64 should be requested as reviewer for Vala changes, me for Rust and @sonnyp for the rest.

Instead of requiring approval, we could also require reviews from code owners instead.
This should have the side effect, that code owners don't need approvals for code changes they own.
image


lw64:
nice!

PR #545, opened 2023-08-29
Library: ported `Checkbox` in Rust

itsAdee:
Recently Learned about the availability of Rust in the Workbench.
I have added the checkbox implementation in Rust . It works on fine on the builder on my machine. Please Check it.

### DIFF
--- a/src/Library/demos/Checkboxes/code.rs
+++ b/src/Library/demos/Checkboxes/code.rs
@@ -1,0 +1,25 @@
+use crate::workbench;
+use gtk::prelude::*;
+
+pub fn main() {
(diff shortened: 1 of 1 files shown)


Hofer-Julian (commented):
Looks good to me, but let's create closures instead of dedicated functions (similar to the javascript version)

itsAdee:
> Looks good to me, but let's create closures instead of dedicated functions (similar to the javascript version)

Modified the code, please check this version of the code, and let me know of any anomaly.Thanks.

Hofer-Julian (approved):
The code looks good to me.
Will merge it when I checked it locally at home.
If another maintainer has time earlier, they can do it too and merge it then.

Hofer-Julian (approved):
Thank you!

PR #548, opened 2023-08-29
library: ported `Overlay` in rust

mbilal234:
I have added overlay in rust.

Please review

### DIFF
--- a/src/Library/demos/Overlay/code.rs
+++ b/src/Library/demos/Overlay/code.rs
@@ -1,0 +1,8 @@
+use crate::workbench;
+use gtk::gio;
+
+pub fn main() {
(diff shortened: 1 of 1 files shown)


src/Library/demos/Overlay/code.rs
```diff
@@ -0,0 +1,8 @@
+use crate::workbench;
+use gtk::gio;
+
+pub fn main() {
+    let file = gio::File::for_uri(workbench::resolve("./image.png").as_str());
```
Hofer-Julian, line 5:
```suggestion
    let file = gio::File::for_uri(&workbench::resolve("./image.png"));
```

Hofer-Julian (commented):
One nitpick, apart from that it looks good :)

Hofer-Julian (approved):
Thanks!

PR #549, opened 2023-08-29
library: ported `Separator` in rust

mbilal234:
I have added rust code for Separator.

Please review :)


Hofer-Julian (commented):
Apart from the nitpick, it looks good

Hofer-Julian:
Need an approval from you for that one @sonnyp since I've modified a file where you are the code owner :)

PR #558, opened 2023-08-29
library: Ported `NavigationView` to rust

saadulkh:
Adds rust version of the `NavigationView` library.

Noticeable differences:
- `switch` of `JavaScript` is replaced with `if-else` statements.
- `current_page` variable is used at two places to avoid multiple function calls.

Lemme know if the `current_page` is unnecessary for the  use case.

### DIFF
--- a/src/Library/demos/Navigation View/code.rs
+++ b/src/Library/demos/Navigation View/code.rs
@@ -1,0 +1,38 @@
+use crate::workbench;
+use adw::prelude::*;
+
+use crate::glib::clone;
+use gtk::glib;
+
+pub fn main() {
+    let nav_view: adw::NavigationView = workbench::builder().object("nav_view").unwrap();
+    let nav_pageone: adw::NavigationPage = workbench::builder().object("nav_pageone").unwrap();
+    let next_button: gtk::Button = workbench::builder().object("next_button").unwrap();
+    let previous_button: gtk::Button = workbench::builder().object("previous_button").unwrap();
+    let nav_pagetwo: adw::NavigationPage = workbench::builder().object("nav_pagetwo").unwrap();
+    let nav_pagethree: adw::NavigationPage = workbench::builder().object("nav_pagethree").unwrap();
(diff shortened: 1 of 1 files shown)


src/Library/demos/Navigation View/code.rs
```diff
@@ -0,0 +1,38 @@
...
+    next_button.connect_clicked(
+        clone!(@weak nav_view, @weak nav_pageone, @weak nav_pagefour => move |_| {
+            let current_page = nav_view.visible_page().unwrap();
+            if current_page == nav_pageone {
+                nav_view.push(&nav_pagetwo);
+            } else if current_page == nav_pagetwo {
+                nav_view.push(&nav_pagethree);
+            } else if current_page == nav_pagethree {
+                nav_view.push(&nav_pagefour);
+            }
```
Hofer-Julian, line 26: Good opportunity to use `match` :)

Hofer-Julian (commented):
Apart from the comment, it looks good :)

src/Library/demos/Navigation View/code.rs
saadulkh: It may work like this. But not sure if we can apply `match` for a `NavigationPage` variable directly.
```
        match current_page {
            x if x == nav_pageone => nav_view.push(&nav_pagetwo),
            x if x == nav_pagetwo => nav_view.push(&nav_pagethree),
            x if x == nav_pagethree => nav_view.push(&nav_pagefour),
            _ => (),
        }
```
Following gives a behavior where only first arm is executed in any condition.
```
        match current_page {
            nav_pageone => nav_view.push(&nav_pagetwo),
            nav_pagetwo => nav_view.push(&nav_pagethree),
            nav_pagethree => nav_view.push(&nav_pagefour),
            _ => (),
        }
```

Hofer-Julian: Ah yes, fair point!

PR #559, opened 2023-08-30
Ported `Overlay Split View`  in Rust

itsAdee:
Hey. I  ported `Overlay Split View` in Rust.
Tried to follow the Java Script code, but there were ownership errors so i had to use `ref cell` library and `borrow_mut()` function to resolve the ownership  issues for `split view`   . 
Please check this approach it works correctly on my machine and let me know on how can i make it better.


src/Library/demos/Overlay Split View/code.rs
```diff
@@ -0,0 +1,27 @@
+use crate::workbench;
+use adw::prelude::*;
+use std::cell::RefCell;
+use std::rc::Rc;
+
+pub fn main() {
+    let overlay_split_view: Rc<RefCell<adw::OverlaySplitView>> = Rc::new(RefCell::new(
+        workbench::builder().object("split_view").unwrap(),
+    ));
```
Hofer-Julian, line 9: You should need neither Rc nor RefCell here.
GObjects already behave like `Rc<RefCell<T>>`

src/Library/demos/Overlay Split View/code.rs
```diff
@@ -0,0 +1,27 @@
...
+
+    let start_toggle: gtk::ToggleButton = workbench::builder().object("start_toggle").unwrap();
+    let end_toggle: gtk::ToggleButton = workbench::builder().object("end_toggle").unwrap();
+
+    let overlay_split_view_clone = overlay_split_view.clone();
+    start_toggle.connect_toggled(move |_| {
```
Hofer-Julian, line 15: You can use `glib::clone!` here.

src/Library/demos/Overlay Split View/code.rs
```diff
@@ -0,0 +1,27 @@
...
+        overlay_split_view_clone
+            .borrow_mut()
+            .set_sidebar_position(gtk::PackType::Start);
+    });
+
+    let overlay_split_view_clone = overlay_split_view.clone();
+    end_toggle.connect_toggled(move |_| {
```
Hofer-Julian, line 22: Also use `glib::clone!` here

Hofer-Julian (changes requested):
Reading this chapter should clear some things up.
https://gtk-rs.org/gtk4-rs/stable/latest/book/g_object_memory_management.html

src/Library/demos/Overlay Split View/code.rs
itsAdee: > Also use `glib::clone!` here

Made the changes please check again, and thanks again for the memory management resource it helped a lot. :)

src/Library/demos/Overlay Split View/code.rs
```diff
@@ -0,0 +1,21 @@
...
+use adw::prelude::*;
+
+pub fn main() {
+    let overlay_split_view: adw::OverlaySplitView =
+        workbench::builder().object("split_view").unwrap();
+
+    let start_toggle: gtk::ToggleButton = workbench::builder().object("start_toggle").unwrap();
+    let end_toggle: gtk::ToggleButton = workbench::builder().object("end_toggle").unwrap();
+
+    start_toggle.connect_toggled(glib::clone!(@strong overlay_split_view => move |_| {
```
saadulkh, line 12: Use `@weak` reference to the `overlay_split_view` here.

src/Library/demos/Overlay Split View/code.rs
```diff
@@ -0,0 +1,21 @@
...
+        overlay_split_view
+            .set_sidebar_position(gtk::PackType::Start);
+    }));
+
+    end_toggle.connect_toggled(glib::clone!(@strong overlay_split_view =>move |_| {
```
saadulkh, line 17: `clone!` is not needed here.

Hofer-Julian (approved):
I've changed your code slightly.
You can find the changes here: https://github.com/sonnyp/Workbench/pull/559/commits/6ad72469f09a091c0b3e2c9f96fdb6a4d5531521

saadulkh (commented):
I've added come comments as well. @Hofer-Julian Can you confirm if there's any difference with and without the changes that I suggested..?

Hofer-Julian:
I agree with @saadulkh suggestions

itsAdee:
> I agree with @saadulkh suggestions

Added them , thanks to both of you for your suggestions :)

PR #566, opened 2023-09-02
Handle Rust errors properly

### LINKED ISSUE #557: Cannot inspect Rust panics
If Rust fails, that the error message nicely visible in the console.
However, if Rust panics at runtime, all you get is:

```
https://gitlab.gnome.org/World/Rust/gtk-rust-template/-/blob/e86986f26a1bb5bae0a0a770bacb739365cf2e4d/src/main.rs#L2
```

Hofer-Julian:
This way, the rust shared library returns 1 when a panic is handled.
I tried to modify `previewer.vala` to handle that, but I couldn't get anything to work.
Maybe @lw64 can take over that part? 😇

Fixes #557

### DIFF
--- a/src/Previewer/previewer.vala
+++ b/src/Previewer/previewer.vala
@@ -122,7 +122,7 @@ namespace Workbench {
       Gtk.StyleContext.add_provider_for_display (Gdk.Display.get_default (), this.css , Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
     }
 
-    public void run (string filename, string uri) {
+    public void run (string filename, string uri) throws DBusError {
       if (this.module != null) {
         this.module.close ();
       }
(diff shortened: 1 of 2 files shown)


lw64:
As far as I know crashing the previewer shouldn't be a problem. It just restarts.

Hofer-Julian:
> As far as I know crashing the previewer shouldn't be a problem. It just restarts.

That's what we did before. But that way we got confusing warnings and the relevant (Rust) one wasn't displayed.

Hofer-Julian:
@lw64 see https://github.com/sonnyp/Workbench/issues/557

I assume Vala shows similar behaviour?

lw64:
So  this 1 return value needs to be handled by the workbench process?

Hofer-Julian:
> So this 1 return value needs to be handled by the workbench process?

I think that's the way to go, yes.

src/Previewer/previewer.vala
```diff
@@ -157,7 +157,10 @@ namespace Workbench {
         return;
       }
       var main_function = (MainFunction) function;
-      main_function ();
+      int result = main_function ();
+      if (result != 1) {
+            throw new GLib.DBusError.SPAWN_CHILD_EXITED("demo exited with error code 1");
```
lw64, line 162: This will need also changes in the dbus previewer handling in workbench js.

PR #573, opened 2023-09-05
Manuals: Optimize loading docs

AkshayWarrier:
This PR improves the time it takes to create and load the sidebar.
Thanks @andyholmes :)
https://gitlab.gnome.org/-/snippets/6127


src/DocumentationViewer.js
```diff
@@ -110,117 +108,104 @@ export default function DocumentationViewer({ application }) {
...
+
+async function scanLibraries(model, base_dir) {
+  const libraries = [];
+
+  const iter = await base_dir.enumerate_children_async(
+    "standard::*",
+    Gio.FileQueryInfoFlags.NOFOLLOW_SYMLINKS,
+    GLib.PRIORITY_DEFAULT,
+    null,
+  );
```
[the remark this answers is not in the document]
AkshayWarrier: I haven't measured the times but there's a noticeable performance difference. When opening Manuals, you can still see the items in the sidebar coming in with ``for await`` but when reading 10 files at a time it's almost instant.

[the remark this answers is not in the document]
andyholmes: Yeah, I think the default `async` iterator is usually the better choice, but for a load-time operation like this it ends up deferring too much.

src/DocumentationViewer.js
```diff
@@ -110,117 +108,104 @@ export default function DocumentationViewer({ application }) {
...
+      name: namespace,
+      search_name: namespace,
+      uri: html_file.get_uri(),
+      children: getChildren(index, directory),
+    });
+
+    model.append(page);
+  } catch (error) {
+    if (!error.matches(Gio.IOErrorEnum, Gio.IOErrorEnum.NOT_FOUND))
+      logError(error);
```
andyholmes, line 163: Maybe in context, this should re-`throw` the error instead? I'm not sure what the best behaviour is if something actually goes wrong.

src/DocumentationViewer.js
```diff
@@ -110,117 +108,104 @@ export default function DocumentationViewer({ application }) {
...
+  "libnotify-0",
+  "webkit2gtk-4.1",
+  "webkit2gtk-web-extension-4.1",
+];
+
+async function scanLibraries(model, base_dir) {
+  const libraries = [];
+
+  const iter = await base_dir.enumerate_children_async(
+    "standard::*",
```
andyholmes, line 180:
```suggestion
    "standard::name,standard::type",
```

Since we're here to optimize.

src/DocumentationViewer.js
```diff
@@ -110,117 +108,104 @@ export default function DocumentationViewer({ application }) {
...
+
+    scanLibraries(root_model, base_path)
+      .then(() => {
+        browse_list_view.model.selected = 12;
+        const search_model = flattenModel(root_model);
+        filter_model.model = search_model;
+      })
+      .catch(console.error);
+
+    browse_list_view.model = createBrowseSelectionModel(root_model, webview);
```
andyholmes, line 130:
```suggestion
    try {
      const root_model = Gio.ListStore.new(DocumentationPage);

      scanLibraries(root_model, base_path)
        .then(() => {
          browse_list_view.model.selected = 12;
          const search_model = flattenModel(root_model);
          filter_model.model = search_model;
        })
        .catch(console.error);

      browse_list_view.model = createBrowseSelectionModel(root_model, webview);
    } catch(error) {
      console.error(error);
    }
```

I don't know where it makes most sense to `catch`, but there should either be a `try..catch` here or a `catch()` when it's invoked.

andyholmes (changes requested):
Looks pretty good, just a few tweaks.

AkshayWarrier:
I made the changes and I also realized that ``StringSorter`` wasn't working in the browse model so I switched back to using ``CustomSorter``

andyholmes (approved):
LGTM

PR #579, opened 2023-09-06
library: Add Rust file for demos without code

Hofer-Julian:


Hofer-Julian:
> Why?

Because od the way the previewer is set up. The demos without Javascript code don't compile in Rust. 
The Vala demos have the same problem.

Hofer-Julian:
> I see, the problem is when opening a Library demo that is not ported yet and pressing "Run".

This is about demos that have no javascript code, not demos that are not ported yet :)

Hofer-Julian:
Should have been more specific in the PR, sorry for that.

Hofer-Julian:
For context, @sonnyp and I agreed that the approach in this PR makes the most sense for now.
@lw64 I've also did the same for Vala, seems to work fine as far as I can tell.

lw64 (approved):
If its not too much work for you, it would be nice to havea line brak after the `{`, but its really not important.

lw64:
But thanks for the work :)

PR #589, opened 2023-09-10
library: Add Rust demo Map

M-Sabrina:


Hofer-Julian:
Two things that we've noticed while working on this demo:
- These constants seem to be only accessible through the `ffi` module, since they seem to be useful for normal code, maybe they should be exposed? https://github.com/sonnyp/Workbench/blob/7669c2c40b691896e9965c2cb504ca812692b815/src/Library/demos/Map/code.rs#L6
- This demo errors with 
 ```
(process:406): Gtk-ERROR **: 12:23:48.992: failed to add UI: .:74:1 Invalid object type 'ShumateSimpleMap'
```
and then later crashes with 
```
thread '<unnamed>' panicked at 'Builder instance should already be initialized.', [workbench.rs:11](http://workbench.rs:11/):14
fatal runtime error: failed to initiate panic, error 5
``` 
If there's a bug in the demo or Workbench I wouldn't be able to find it. What is interesting is that equivalent JavaScript demo seems to work fine.

Do you think you could have a look @bilelmoussaoui?

**Comment by bilelmoussaoui** [comment]
> These constants seem to be only accessible through the ffi module, since they seem to be useful for normal code, maybe they should be exposed? 

Please open an issue upstream

**Comment by bilelmoussaoui** [comment]
Regarding the `ShumateSimpleMap`  I guess that happens because the gtype is not registered yet and shumate doesn't provide an init function.

**Comment by Hofer-Julian** [comment]
> > These constants seem to be only accessible through the ffi module, since they seem to be useful for normal code, maybe they should be exposed?
> 
> Please open an issue upstream

Will do.



> Regarding the `ShumateSimpleMap` I guess that happens because the gtype is not registered yet and shumate doesn't provide an init function.

Do I have to register the type beforehand in Workbench or should there be something changed in `libshumate-rs`?

**Comment by bilelmoussaoui** [comment]
> Do I have to register the type beforehand in Workbench, should there be something changed in libshumate-rs, or should I open at libshumate an issue to add an init function?

I guess so, yes. If libshumate can add such function, that would be great.

**Comment by Hofer-Julian** [comment]
I've opened https://gitlab.gnome.org/GNOME/libshumate/-/issues/64 and https://gitlab.gnome.org/World/Rust/libshumate-rs/-/issues/3

Until the former is implemented and released, we will need to register the used gtypes manually.
As far as I can tell, this needs to be done in the `run` function of `previewer.vala`. @lw64 do you know how to register gtypes in Vala?

**Comment by bilelmoussaoui** [comment]
> 

Thanks! Another option we could have is to make `gtk::Builder::object::<T>` call internally `T::ensure_type()` for you =) similar to how `gtk::TemplateChild<T>` does.

**Comment by Hofer-Julian** [comment]
> > 
> 
> Thanks! Another option we could have is to make `gtk::Builder::object::<T>` call internally `T::ensure_type()` for you =) similar to how `gtk::TemplateChild<T>` does.

Thanks for the suggestion, but when the Rust code is executed it is already too late.
GTK gives the error when it is at the Vala part of the code, and as soon as it reaches the Rust there's nothing I can do anymore.

**Comment by Hofer-Julian** [comment]
> > 
> 
> Thanks! Another option we could have is to make `gtk::Builder::object::<T>` call internally `T::ensure_type()` for you =) similar to how `gtk::TemplateChild<T>` does.

Oh, I read your answer too fast @bilelmoussaoui.
That would be great. Should I open an issue for that?

**Comment by bilelmoussaoui** [comment]
> > > 
> > 
> > 
> > Thanks! Another option we could have is to make `gtk::Builder::object::<T>` call internally `T::ensure_type()` for you =) similar to how `gtk::TemplateChild<T>` does.
> 
> Oh, I read your answer too fast @bilelmoussaoui. That would be great. Should I open an issue for that?

sure, why not :)

**Comment by Hofer-Julian** [comment]
> But isnt it a problem that the UI files gets loaded first, and then the code is executed that registers the types?

Yes, I think that is exactly the problem.
This is why I think we need to register the types before the UI is loaded. That happens in the Vala code, right?

**Review by Hofer-Julian — APPROVED** [review]
Works as expected.
I've added a `gtk::init()` call and ensured the `Shumate.SimpleMap` type within Vala.

File: src/Previewer/previewer.vala [hunk 1]
```diff
@@ -118,7 +119,7 @@ namespace Workbench {
         var end = section.get_end_location();
         this.css_parser_error(error.message, (int)start.lines, (int)start.line_chars, (int)end.lines, (int)end.line_chars);
       });
-      this.css.load_from_data (content, -1);
+      this.css.load_from_data (content.data);
```
Hofer-Julian, line 122: Not sure where this change comes from.
If I revert it though, I get the following error:

```
../../../../../../../../../Projekte/github.com/sonnyp/Workbench/src/Previewer/previewer.vala:122.32-122.38: error: Argument 1: Cannot convert from `unowned string' to `unowned uint8[]'
  122 |       this.css.load_from_data (content, -1);
      |                                ^~~~~~~      
../../../../../../../../../Projekte/github.com/sonnyp/Workbench/src/Previewer/previewer.vala:122.7-122.43: error: 1 extra arguments for `void Gtk.CssProvider.load_from_data (uint8[])'
  122 |       this.css.load_from_data (content, -1);
```

lw64:
> Works as expected. I've added a `gtk::init()` call and ensured the `Shumate.SimpleMap` type within Vala.

Id there will be more libraries requiring something similar, a generic solution with a dbus method or so would be appropriate I think

Issue #588, opened 2023-09-10
Moving to Freedesktop 23.08 broke build


M-Sabrina:
The following commit https://github.com/sonnyp/Workbench/commit/e54bbad1d2ea50428ec7c1c012b05b94e613c0e3 broke the Workbench build:
```
Compilation failed: 2 error(s), 14 warning(s)
[30/38] Generating src/libworkbench/Workbench-0.gir with a custom command (wrapped by meson to set env)
/var/home/sabrina/.var/app/org.gnome.Builder/cache/gnome-builder/projects/Workbench/builds/re.sonny.Workbench.Devel.json-flatpak-org.gnome.Sdk-45beta-x86_64-main/src/libworkbench/libworkbench-enums.h:13: syntax error, unexpected identifier in 'WORKBENCH_EXPORT' at 'WORKBENCH_EXPORT'
/var/home/sabrina/Projekte/Workbench/src/libworkbench/workbench-completion-request.h:34: syntax error, unexpected identifier in 'WORKBENCH_EXPORT' at 'WORKBENCH_EXPORT'
/var/home/sabrina/Projekte/Workbench/src/libworkbench/workbench-completion-request.h:37: syntax error, unexpected identifier in 'WORKBENCH_EXPORT' at 'WORKBENCH_EXPORT'
/var/home/sabrina/Projekte/Workbench/src/libworkbench/workbench-completion-request.h:39: syntax error, unexpected identifier in 'WORKBENCH_EXPORT' at 'WORKBENCH_EXPORT'
/var/home/sabrina/Projekte/Workbench/src/libworkbench/workbench-completion-request.h:41: syntax error, unexpected identifier in 'WORKBENCH_EXPORT' at 'WORKBENCH_EXPORT'
/var/home/sabrina/Projekte/Workbench/src/libworkbench/workbench-completion-request.h:43: syntax error, unexpected identifier in 'WORKBENCH_EXPORT' at 'WORKBENCH_EXPORT'
/var/home/sabrina/Projekte/Workbench/src/libworkbench/workbench-completion-request.h:45: syntax error, unexpected identifier in 'WORKBENCH_EXPORT' at 'WORKBENCH_EXPORT'
/var/home/sabrina/Projekte/Workbench/src/libworkbench/workbench-completion-request.h:48: syntax error, unexpected identifier in 'WORKBENCH_EXPORT' at 'WORKBENCH_EXPORT'
[...]

PR #590, opened 2023-09-10
fix: remove -fvisibility=hidden

andyholmes:
This is an internal library anyways and seemed to be causing some problems, or at least warnings, when compiling other parts of Workbench.


andyholmes:
Possibly addresses #588

PR #593, opened 2023-09-14
Attempt to build Rust stuff offline

Hofer-Julian:
cargo-c works fine, but gst-plugin-gtk4 does not
See https://github.com/flathub/io.gitlab.daikhan.stable for existing manifest


Hofer-Julian:
I will give up on that, 2h was my personal limit I wanted to spend on it.
If someone else is interested, feel free to pick it up from where I stopped :)

PR #594, opened 2023-09-15
library: Ported View switcher to rust

saadulkh:
Check button is not working right now. @Hofer-Julian Plz have a look.

### DIFF
--- a/src/Library/demos/View Switcher/code.rs
+++ b/src/Library/demos/View Switcher/code.rs
@@ -26,8 +26,8 @@ pub fn main() {
             .build();
 
         button.connect_clicked(clone!(
-            @strong notification_count, @strong notifications_page,
-            @strong notification_list, @strong notification_row => move |_| {
+            @strong notifications_page, @strong notification_list,
+            @strong notification_row => move |_| {
                 notifications_page.set_badge_number(notifications_page.badge_number() - 1);
                 notification_list.remove(&notification_row);
 
--- a/src/Library/demos/View Switcher/code.rs
+++ b/src/Library/demos/View Switcher/code.rs
@@ -1,18 +1,17 @@
 use crate::workbench;
 use adw::prelude::*;
 
-use crate::glib::clone;
-use std::cell::Cell;
-use std::rc::Rc;
+use glib::clone;
+use gtk::glib;
 
 pub fn main() {
     let notifications_page: adw::ViewStackPage = workbench::builder().object("page3").unwrap();
     let notification_list: gtk::ListBox = workbench::builder().object("notification_list").unwrap();
 
-    let notification_count = Rc::new(Cell::new(5));
-    notifications_page.set_badge_number(notification_count.get());
+    let notification_count = 5;
+    notifications_page.set_badge_number(notification_count);
 
-    for _ in 0..notification_count.get() {
+    for _ in 0..notification_count {
         let notification_row = adw::ActionRow::builder()
             .title("Notification")
             .selectable(false)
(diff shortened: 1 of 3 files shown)


src/Library/demos/View Switcher/code.rs
```diff
@@ -0,0 +1,47 @@
...
+            .halign(gtk::Align::Center)
+            .valign(gtk::Align::Center)
+            .margin_top(10)
+            .margin_bottom(10)
+            .icon_name("check-plain-symbolic")
+            .build();
+
+        button.connect_clicked(clone!(
+            @weak notification_count, @weak notifications_page,
+            @weak notification_list, @weak notification_row => move |_| {
```
Hofer-Julian, line 32: They all need to be strong references instead of weak ones.
Technically, only the first loop entry needs to be strong, but that would be annoying to program here.

src/Library/demos/View Switcher/code.rs
```diff
@@ -0,0 +1,47 @@
...
+use crate::glib::clone;
+use gtk::glib;
+use std::cell::Cell;
+use std::rc::Rc;
+
+pub fn main() {
+    let notifications_page: adw::ViewStackPage = workbench::builder().object("page3").unwrap();
+    let notification_list: gtk::ListBox = workbench::builder().object("notification_list").unwrap();
+
+    let notification_count = Rc::new(Cell::new(5));
```
Hofer-Julian, line 13: That is a 1:1 port of JavaScript, which is good.
Tbh I think the JavaScript version is a bit odd, and it especially shows with the Rust version.
Let's make this a normal integer instead.

src/Library/demos/View Switcher/code.rs
```diff
@@ -0,0 +1,47 @@
...
+            .margin_top(10)
+            .margin_bottom(10)
+            .icon_name("check-plain-symbolic")
+            .build();
+
+        button.connect_clicked(clone!(
+            @weak notification_count, @weak notifications_page,
+            @weak notification_list, @weak notification_row => move |_| {
+                notification_count.set(notification_count.get() - 1);
+                notifications_page.set_badge_number(notification_count.get());
```
Hofer-Julian, line 34: Instead of using notification_count, you could just get the current badge number

src/Library/demos/View Switcher/code.rs
```diff
@@ -0,0 +1,47 @@
+use crate::workbench;
+use adw::prelude::*;
+
+use crate::glib::clone;
```
Hofer-Julian, line 4:
```suggestion
use glib::clone;
```

src/Library/demos/View Switcher/code.rs
saadulkh: I forgot about the usage of `strong` xD.
Thanks!

src/Library/demos/View Switcher/code.rs
saadulkh: It results in a compile time error.

> error[E0432]: unresolved import `glib`
>  --> code.rs:4:5
>   |
> 4 | use glib::clone;
>   |     ^^^^ help: a similar path exists: `crate::glib`
>   |
>   = note: `use` statements changed in Rust 2018; read more at <https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html>

saadulkh: My bad. Yes it can be updated.

saadulkh:
Lemme give JavaScript a try as well

saadulkh:
Should I synchronize the JavaScript version too?
@sonnyp

Hofer-Julian:
Thanks, @saadulkh.
Do you think you could update the Vala example as well?

saadulkh:
> Thanks, @saadulkh. Do you think you could update the Vala example as well?

Lemme give it a try.

Hofer-Julian:
One small warning from the Javascript linter in the CI, then we are ready to go :)

saadulkh:
Some problem with the CI now?

saadulkh:
Plz re-run CI manually.

Hofer-Julian (approved):
Looks good. Thanks a lot! 😁

PR #602, opened 2023-09-17
Use runtime extensions for language dependencies, don't copy em

orowith2os:
Makes it easier to get things like bugfixes for Rust faster.

Needs *lots* of testing, and I'm not sure this'll even work. But it's an attempt.

### DIFF
--- a/build-aux/re.sonny.Workbench.Devel.json
+++ b/build-aux/re.sonny.Workbench.Devel.json
@@ -4,11 +4,22 @@
   "runtime-version": "45beta",
   "sdk": "org.gnome.Sdk",
   "sdk-extensions": [
-    "org.freedesktop.Sdk.Extension.vala",
     "org.gnome.Sdk.Docs",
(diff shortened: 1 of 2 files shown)


UrtsiSantsi:
Copying them after compile really slows the development workflow, using runtime extensions will be wonderful.

orowith2os:
The main issue is, does this even work? Because using a generated bundle, it looks like it doesn't. The Vala extension isn't pulled in when installing it if I omit it from sdk-extensions. It should be, but isn't.

UrtsiSantsi:
Not sure, I was getting some build errors myself last week and now I see #604.
I'll try your changes once 45 is out.

orowith2os:
`add-extensions` is for runtime extensions; `sdk-extensions` is for build-time extensions, at least afaik.

orowith2os:
You still don't get the extensions installed as dependencies when using the
SDK as a runtime.

orowith2os:
I'd rather do this, but whatever works.

PR #605, opened 2023-09-19
build: re-add runtime share directories

andyholmes:


andyholmes:
Doesn't fix anything.

Issue #599, opened 2023-09-17
Build Workbench as part of CI


sonnyp:
Might want to use https://github.com/flatpak/flatpak-github-actions too

PR #606, opened 2023-09-19
build(ci): build flatpak in CI

FineFindus:
Fixes https://github.com/sonnyp/Workbench/issues/599.
Adds a new GitHub action that builds Workbench in the CI. The finished Flatpak is automatically uploaded as an artifact.
As it currently uses the 45beta runtime, the non-beta extensions need to be installed manually.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -1,11 +1,14 @@
 name: CI
 
 on:
+  push:
+    branches: [main]
   pull_request:
     branches: [main]
 
 jobs:
   CI:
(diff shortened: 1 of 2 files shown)


FineFindus:
I mainly enabled it for [testing and developing the action](https://github.com/FineFindus/Workbench/actions), which took a few commits. I squashed them so as not to pollute the main branch.

Having a flatpak build of the main branch can be both an effective way to catch build bugs, and a quick way for designers/users to test the new changes without having to compile the application themselves.
I'm also happy to remove it.

PR #620, opened 2023-09-22
Add Mold and LLVM again

Hofer-Julian:


Hofer-Julian:
As discussed with @sonnyp, we wait a bit before we merge this to see if we can get rid of the LLVM dependency

PR #647, opened 2023-09-25
WIP: library: add CSS Gradient entry

cunidev:
Adds the "psychedelic" CSS gradient from the Berlin45 hackfest

[WIP] because the button's animated gradient does not slide in the Workbench UI for some reason. This is also a bug I encountered in older Gtk versions


cunidev:
I guess this now makes sense to be integrated somewhere on top of #636 (for the "animation" / keyframes part of it) since the other demo is much more complete

PR #655, opened 2023-09-25
Library: Fix Vala code after .blp changes in 'Preferences Window' demo

UrtsiSantsi:
The type of widget was changed from **Gtk.Switch** to **Adw.SwitchRow** in the Blueprint


UrtsiSantsi:
We would need some policy about that - with the addition of new languages it will become increasingly harder for contributors to know all of them.

lw64 (commented):
everything ok.

PR #662, opened 2023-09-30
data: Mark release descriptions as untranslatable

sabriunal:
GNOME automatically excludes release descriptions on Damned Lies (GNOME Translation Platform). It's a good practice to follow the GNOME way.

This can streamline the translation process, allowing translators to focus their efforts on more critical and user-facing aspects of the application.

### DIFF
--- a/data/app.metainfo.xml
+++ b/data/app.metainfo.xml
@@ -43,7 +43,7 @@
   <content_rating type="oars-1.1" />
   <releases>
     <release version="45.1" date="2023-09-29">
-      <description>
+      <description translatable="no">
         <ul>
           <li>Reduce Workbench download size</li>
           <li>Add "Extensions" to enable Rust and Vala support</li>
@@ -59,7 +59,7 @@
       </description>
     </release>
     <release version="45" date="2023-09-22">
-      <description>
+      <description translatable="no">
         <ul>
           <li>Use GNOME 45</li>
           <li>Support Rust</li>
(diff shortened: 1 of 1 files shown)


sabriunal:
My reasons are straightforward:

- I translate many FOSS applications into Turkish, and there are extensive release histories to translate, which becomes challenging due to frequent user interface (UI) changes.
- Developers often forget to notify translators before new releases. As a result, I often discover new releases when I encounter untranslated strings.
- GNOME already excludes these strings from their translation process on DL. 

Release notes are only partially visible when you visit an application's page on GNOME Software. Initially, it displays just four lines of release notes, requiring additional clicks to access the full content. Even more click to see older release notes.

workbench

A similar situation occurs on Flathub. You need to scroll down and click an extra button to view the recent release notes. From Flathub's perspective, there's no value in translating older release notes, as it does not show them.

f1
f2
f3

For instance, LibreOffice displays a notification bar when updating to a major release, such as going from 7.5 to 7.6.

Some GTK applications integrate the latest release notes into the 'About' dialog to make them visible. However, there are still some issues. Few users are aware of the existence of a dedicated release notes page. Often applications don't send a toast about release notes, and you will be lucky to discover them.

It could be beneficial if developers used notifications (like toasts) to inform users about significant changes in the latest release. Users could click on the notification to access the relevant dialog, or it could automatically disappear after 10 seconds.

Your text is clear, but here are some minor corrections for grammar and clarity:

As a result:

- You can keep the latest release note as translatable because they might be visible to some lucky users.
- You can create a subpage in the 'About' dialog to show important changes.
- You can notify the user if there are some important changes.

(Grammar mistakes are corrected by ChatGPT)

sabriunal:
Eyedropper

e1
e2
e3



https://github.com/FineFindus/eyedropper/blob/master/src/widgets/about_window.rs#L101-L107

PR #663, opened 2023-09-30
Update extensions window

bragefuglseth:
Updates the visuals of the extensions window and gives it an illustration.

## Screenshots
 
Screenshot from 2023-09-30 10-48-55

Screenshot from 2023-09-30 10-48-29


bragefuglseth:
@sonnyp: I'm not very emotionally tied to any of the code I've written here, so if you have any changes you want made to it, and you think it would be better/quicker if you just made them yourself, feel free to do so. Or just mention any problems / design issues to me as normal, and I'll have a look a them.

PR #665, opened 2023-10-01
library: Improve JavaScript demo Accessibility

M-Sabrina:
Modified inversion of state in function toggleButton for better clarity. Also, corresponds better to suggested Rust example (https://github.com/sonnyp/Workbench/pull/664).


andyholmes (approved):
LGTM

PR #671, opened 2023-10-07
library: Port Media, Image and Picture demos to Vala

Diego-Ivan:
This PR ports the Audio, Image, Picture and Video demos to Vala.


lw64 (approved):

lw64:
Thanks for the ports!

PR #672, opened 2023-10-07
library: Add WebView demo in Rust

Hofer-Julian:
- Ensure types of WebKitGtk 6.0 in previewer.vala
- Add WebView demo

### DIFF
--- a/src/Previewer/previewer.vala
+++ b/src/Previewer/previewer.vala
@@ -59,10 +59,15 @@ namespace Workbench {
       return true;
     }
 
-    public void update_ui (string content, string target_id, string original_id = "") {
+    // This registers GObjects so that they can be found by GtkBuilder
+    void ensure_types() {
       typeof (Shumate.SimpleMap).ensure();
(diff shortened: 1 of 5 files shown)


M-Sabrina (approved):

Hofer-Julian:
Since CI also fails at https://github.com/sonnyp/Workbench/pull/671, I am going ahead and merge this.

src/Previewer/previewer.vala
```diff
@@ -59,10 +59,15 @@ namespace Workbench {
       return true;
     }
 
-    public void update_ui (string content, string target_id, string original_id = "") {
+    // This registers GObjects so that they can be found by GtkBuilder
+    void ensure_types() {
```
lw64, line 63: This should have the `private` modifier.

Hofer-Julian: Private is not the default?
Then why do we put public on all the other functions?

lw64: It is, but it should be explicit.

PR #680, opened 2023-10-09
manuals: Fix a couple of bugs

sonnyp:
* Select the text on search focus (ctrl+k) so the user can type something new to replace the previous search directly
* Fix libadwaita warnings about navigation page not having titles
* Fix an issue where you can't load (because of autoselect) the first filter result
* Reset Manuals when the window has been hidden and reopened (hide-on-close)
  * Focus the search
  * Reset search


AkshayWarrier (approved):
All these look like nice improvements, LGTM!

> Fix an issue where you can't load (because of autoselect) the first filter result

Oo I noticed this as well, and I was trying to figure out why it was the case. Good catch.

Issue #640, opened 2023-09-24
Extract Manuals into a standalone app


sonnyp:
We should extract the offline documentation viewer “Manuals” into a standalone app so that

1. it benefits everybody, not just Workbench users
2. it's easier to work on it by removing Workbench complexity

To-do

1. [x] Create a new GitHub repo with Workbench code as template @sonnyp 
2. [x] Remove everything that's not related to Manuals @sonnyp 
3. [x] Adjust metadata / credits / doap / metainfo / ... @sonnyp 
4. [x] Add collaborators 
5. [x] Publish to Flathub :)
6. [x] Remove Manuals code from Workbench
7. [ ] Add Manuals as a module to Workbench

I will co-maintain with @AkshayWarrier - @andyholmes would you be interested too?

Once we have a repo we can create issues for the good ideas that have been floating around:

* [x] Mobile https://github.com/sonnyp/Workbench/issues/627
* [ ] Tabs
* [ ] Browser 
* [ ] DevDocs
* [x] Bookmarks to Rust/GJS/Vala/PyGObject/...
* [ ] Support redirecting "https://docs.gtk.org/gtk4/class.AppChooserButton.html" to local offline doc
* [ ] Support navigating to http URLs
* [ ] Support the docset format https://zealdocs.org/
* [ ] Support download/viewing multiple sdk docs version
* [x] #703
* [ ] Add “Find” (in current page)
* [ ] Add full text search

andyholmes:
It seems like there might be some overlap with Devhelp here. Is the plan to replace it, or maybe use libdevhelp somehow?

andyholmes:
Maybe we should see what Christian had planned for Builder? Seems like a library that could be shared between Workbench/Manuals/Builder would be a good way to share efforts.

PR #682, opened 2023-10-12
Manuals: Sync the sidebar to search view and webview

AkshayWarrier:
This PR makes the following changes
- Selecting an item in the search view will select the item in the browse view
- Any time the webview's uri changes due to going back/forward or clicking on a link within a documentation to another API reference will update the browse view as well
- The tree list model is now created in a sorted manner using https://docs.gtk.org/gio/method.ListStore.insert_sorted.html, instead of sorting it after the model is created

### DIFF
--- a/src/Manuals/DocumentationViewer.js
+++ b/src/Manuals/DocumentationViewer.js
@@ -178,7 +178,7 @@ export default function DocumentationViewer({ application }) {
   search_model.connect("selection-changed", () => {
     const uri = search_model.selected_item.uri;
     const sidebar_path = URI_TO_SIDEBAR_PATH[uri];
-    selectSidebarItem(browse_selection_model, sidebar_path, browse_list_view);
+    selectSidebarItem(browse_list_view, sidebar_path);
   });
 
   let promise_load;
--- a/src/Manuals/DocumentationViewer.js
+++ b/src/Manuals/DocumentationViewer.js
@@ -50,6 +50,9 @@ const DocumentationPage = GObject.registerClass(
   class DocumentationPage extends GObject.Object {},
 );
 
+const URI_TO_SIDEBAR_PATH = {};
+let sync_sidebar = false;
+
 export default function DocumentationViewer({ application }) {
   const builder = Gtk.Builder.new_from_resource(resource);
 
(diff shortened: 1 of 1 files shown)


AkshayWarrier:
It does scroll to the right position but it only scrolls until the item gets into view (which is usually when the item is at the end of the view), and the problem is that the item gets covered by the Search Entry. Maybe I could either manually offset it a little more than it needs or have the search entry not hide any times. There is also this struct https://docs.gtk.org/gtk4/struct.ScrollInfo.html which lets you modify the scroll behaviour and it is taken by this function https://docs.gtk.org/gtk4/method.ListView.scroll_to.html but idk how to use it.

AkshayWarrier:
So the scrolling always works for the browse view, whenever the item is barely visible after ``scroll_to`` is called, it will adjust to make it visible. For the search view, I'm not exactly sure but it looks like it fails to scroll to the item on the first selection but seems to work after that. I'd like you to confirm this, it might be an issue with ``scroll_to``?

If I'm being honest, I feel like this is good enough considering it works for the most part. I could just specifically check for a condition where I do something if it's the first time someone is selecting a search result, but that feels odd.

AkshayWarrier:
Actually sorry, I think there's still a bug with this. It looks like ``value-changed`` of the scrollbar adjustment doesn't always get triggered which is a problem. I'll have to figure out a way to fix that.

AkshayWarrier:
> Actually sorry, I think there's still a bug with this. It looks like `value-changed` of the scrollbar adjustment doesn't always get triggered in an edge case which is a problem. I'll have to figure out a way to fix that.

Okay I think this should be fixed now? It looks like it but I still don't have an answer for

> For the search view, I'm not exactly sure but it looks like it fails to scroll to the item on the first selection but seems to work after that.

AkshayWarrier:
Oh wait, I'm not syncing the items in the search view and the web view. Whenever you select an item on the search view, it gets selected in the browse view when you go back (so that item should be scrolled into view when you go back), that's what I mean by syncing in the context of search view. I don't think you need to sync the items in the search view itself as I don't see that as being helpful? DevHelp doesn't do it either.

AkshayWarrier:
> If I research multiple thing and jump back and forth between the pages in random order it will be a lot easier to use the sidebar and click on one of the already expanded items instead of going back or forward multiple times with the buttons. With this change all other items are collapsed and it is impossible to use it in that way.

That is a valid point, I'll have to think of how to do that.

> The same for search - if I open a page, then a linked page and so on a few times in a row and want to go back to the initial page I searched that is still in the sidebar I will not be able to do it without a clicking back a few times or searching again. Even if there are not a lot of paged to go back it is still easier and more natural (and bigger target) to click on the searched item.

So ig you would expect the sidebar to not collapse all the rows and keep its state? Yeah, I noticed now that DevHelp doesn't reset and collapses all the rows. Maybe we could give a separate option to collapse all the rows? Because sometimes opening a section (for ex Gtk-4.0 -> Classes) will just introduce a lot of rows and make it harder to browse.

DevHelp has a small context menu for that
image


> 1. The close button on the left feels strange.

Yeah I think we will change that in the future when Manuals becomes a separate app
> 2. Why is the keyboard shortcuts button down there and next to the search? I would expect search button or nothing.

Yes that is also just a temporary fix, Manuals doesn't have a primary menu, and it will when it becomes a standalone app.
> 3. Why is it called "Focus global search" if there is no local search? And can we enable local (page) search?

That is one of the features we plan to add in the future https://github.com/sonnyp/Workbench/issues/640

PR #689, opened 2023-10-16
library: Remove the need for noop code files 

sonnyp:
Noop code files were introduced in https://github.com/sonnyp/Workbench/pull/579
So that "Run" on demos with node code in Vala or Rust would work.

This is an alternative solution that

* Remove the noop files
* Does not attempt to compile if there is no code
* Remove all tags on the Library demos without code
* Remove the placeholders - they would show in user projects too
* Opens the demo in JavaScript if it's not available in the user preferred (last used) language

image

The downside is that if the user opens the "Code" panel on these, there isn't much to extend from. But it's also an issue with the noop files and since we have multi-windows, they can check an example with "JavaScript", "Rust" or "Vala".


Hofer-Julian:
I am fine with the changed behavior. I didn't test the branch locally yet, though.
Let me know if you want me to do that.

Hofer-Julian:
> @Hofer-Julian I'm confident it works well.
> 
> I'd like to hear your thoughts on
> 
> > The downside is that if the user opens the "Code" panel on these, there isn't much to extend from. But it's also an issue with the noop files and since we have multi-windows, they can check an example with "JavaScript", "Rust" or "Vala".

I would mind if the code panel can be used, just for users to notice that it doesn't do anything.
If you make it somehow hidden or not-targetable, I am fine with it.

Hofer-Julian:
The branch doesn't seem to work as expected for me

[Bildschirmaufzeichnung vom 2023-10-17, 17-37-10.webm](https://github.com/sonnyp/Workbench/assets/30049909/82102f2c-b982-47a1-80a1-2a6815260d18)

Hofer-Julian:
> @Hofer-Julian thanks for catching, I forgot to remove the placeholder which was inadequate because it would also show in user saved projects when switching lang. Done.

Thanks, can try it again in the evening.

Hofer-Julian (approved):
Works as expected

PR #692, opened 2023-10-17
demo: update Elastic URL

sabriunal:
Use the new simplified URL.


sabriunal:
@sonnyp Simple link correction in a demo.

Issue #209, opened 2023-03-06
Use a path instead of abstract socket for dbus connection


sonnyp:
Abstract sockets = bad
require network permissions and are set up outside the sandbox (AFAIK)

Let's use a temporary file instead

https://dbus.freedesktop.org/doc/dbus-specification.html#addresses

PR #697, opened 2023-10-21
Use a file instead of abstract socket for the DBus connection

sonnyp:
Fixes https://github.com/sonnyp/Workbench/issues/209

Also:

• Rename Vala previewer to Module previewer
• Simplify Library tags creation


Hofer-Julian (approved):
Changes look good to me!

PR #698, opened 2023-10-21
Manuals: Add fuzzy searching and tags for search view

AkshayWarrier:
This PR makes the following changes
- Implement fuzzy searching by making use of ``fzy.js``, the module that comes with every gi-docgen documentation to provide fuzzy searching for the documentation.
- Add tags to the search results to show the kind of symbol it is
- You can search using tags by typing ``tag: search term`` to only keep search results of that ``tag``
- Show a status page for when no search results are found
- Remove crossfade transition to make search feel more snappy


AkshayWarrier:
> Since you didn't write the code in fzy.js, you'll need to include a copyright and license header.

Ah okay, I was actually confused about that because the original file did have a license header, but since I modified it a little I didn't know if I had to include it or not.

AkshayWarrier:
> Tags look good. Could we also do something when the entry is deprecated? Maybe dim or strike-through ? Or use the "deprecated" tag

I actually can't find a straightforward way to do this because ``index.json`` does not contain information about deprecation so the only way to know that is by parsing the HTML doc which is not feasible for every doc.

AkshayWarrier:
Yeah that was one of the ideas I had as well, we could do that.

PR #704, opened 2023-10-26
Use GLib.Error matches

sonnyp:


andyholmes (approved):
Good idea!

LGTM, however you can drop the conditional chaining if you want (`matches?.()`), since it's guaranteed to be on all `Error` objects (see [GJS override](https://gitlab.gnome.org/GNOME/gjs/-/blob/master/modules/core/overrides/GLib.js#L280-285)).

Issue #110, opened 2022-06-04
Allow using actions in the external preview


sonnyp:
WIP https://github.com/sonnyp/Workbench/pull/108

sonnyp:
Some ideas:

* Proxy/relay all dbus stuff somehow through the Workbench process
* custom WorkbenchApplication object, that delegates stuff over dbus to the workbench ui. Also stuff like application.quit() could be handled safely then. (by @lw64 )
* Start the vala/previewer process outside the sandbox (probably no)
* Do not use actions at all in Workbench demos (probably no)

sonnyp:
Doing some investigation.

Both applications are registered

![image](https://user-images.githubusercontent.com/19673/204095729-f1855c47-96ab-41ea-ba2b-eb73d637c5b8.png)

but when running the Vala desktop notification demo:

```vala
#!/usr/bin/env -S vala workbench.vala --pkg gtk4 --pkg libadwaita-1

public void main () {
  var application = workbench.application as GLib.Application;

  // this is required now for some reason?
  application.register(null);

  var notification = new Notification ("Lunch is ready");
  notification.set_body (
    "Today we have pancakes and salad, and fruit and cake for dessert"
  );
  notification.set_default_action ("app.notification-reply");
  notification.add_button ("Accept", "app.notification-accept");
  notification.add_button ("Decline", "app.notification-decline");

  var icon = new ThemedIcon ("object-rotate-right-symbolic");
  notification.set_icon (icon);

  var simple_button = workbench.builder.get_object ("button_simple") as Gtk.Button;
  simple_button.clicked.connect (() => {
    application.send_notification ("lunch-is-ready", notification);
  });

  var action_reply = new SimpleAction ("notification-reply", null);
  action_reply.activate.connect (() => {
    stdout.printf ("Reply");
  });
  application.add_action (action_reply);

  var action_accept = new SimpleAction ("notification-accept", null);
  action_accept.activate.connect (() => {
    stdout.printf ("Accept");
  });
  application.add_action (action_accept);

  var action_decline = new SimpleAction ("notification-decline", null);
  action_decline.activate.connect (() => {
    stdout.printf ("Decline");
  });

  application.add_action (action_decline);
}
```

the notification shows but clicking on an action `Accept` or `Decline` appears to communicate with `/re/sonny/Workbench` instead of `/re/sonny/Workbench/vala_previewer`. 

[...]

PR #705, opened 2023-10-27
library: Port Actions demo to Python

rolandlo:
This PR ports the Actions demo to Python

I have tried to stay as close as possible to the JS code. Not sure if I should have avoided using lambdas as in the Welcome demo, see [comment](https://github.com/sonnyp/Workbench/pull/570/files#r1372677674). I used lambdas because here it's only for a single statement, whereas in the Welcome demo there were two statements.

The `open_uri` actions somehow don't work yet. Is this something that needs to be fixed on the "core" side or is there something wrong with my code?

### DIFF
--- a/src/Library/demos/Actions/main.py
+++ b/src/Library/demos/Actions/main.py
@@ -1,0 +1,70 @@
+import gi
+
+gi.require_version("Gtk", "4.0")
+gi.require_version("Adw", "1")
+from gi.repository import Gtk, Adw, GLib, Gio
+import workbench
+
+demo: Adw.StatusPage = workbench.builder.get_object("demo")
+
+demo_group = Gio.SimpleActionGroup()
+demo.insert_action_group("demo", demo_group)
+
+# Action with no state or parameters
+simple_action = Gio.SimpleAction(
+  name = "simple"
+)
+
+simple_action.connect("activate",
+  lambda action, _ : print(f"{action.get_name()} action activated")
+)
+
+demo_group.add_action(simple_action)
+
+# Action with parameter
+bookmarks_action = Gio.SimpleAction(
+  name = "open-bookmarks",
+  parameter_type = GLib.VariantType("s"),
+)
+
+bookmarks_action.connect("activate",
+  lambda action, parameter : print(f"{action.get_name()} activated with {parameter.unpack()}")
+)
+
+demo_group.add_action(bookmarks_action)
+
+# Action with state
+toggle_action = Gio.SimpleAction(
+  name = "toggle",
+  # Boolean actions dont need parameters for activation
+  state = GLib.Variant.new_boolean(False),
+)
+
+toggle_action.connect("notify::state",
+  lambda action, _ : print(f"{action.get_name()} action set to {action.get_state().unpack()}")
+)
+
+demo_group.add_action(toggle_action)
+
+# Action with state and parameter
+scale_action = Gio.SimpleAction(
+  name = "scale",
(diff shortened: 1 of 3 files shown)


Hofer-Julian:
>  I used lambdas because here it's only for a single statement

If it's only a single statement, it's fine to use lambdas IMO

rolandlo:
> @rolandlo thanks! please consider adding yourself to the list of contributors in `about.js`
Done!

src/Library/demos/Actions/main.py
```diff
@@ -0,0 +1,70 @@
...
+
+simple_action.connect("activate",
+  lambda action, _ : print(f"{action.get_name()} action activated")
+)
+
+demo_group.add_action(simple_action)
+
+# Action with parameter
+bookmarks_action = Gio.SimpleAction(
+  name = "open-bookmarks",
```
theCapypara, line 26: One example for PEP 8 is that there should be no spaces around the `=` in keyword arguments.

theCapypara (changes requested):
Thanks! In general this looks good and works perfectly, however could you make sure your code is formatted to follow [PEP 8](https://peps.python.org/pep-0008/) and uses 4 spaces for indentation? 

You can use `pycodestyle` to verify. You can ignore error `E402 module level import not at top of file`, since that isn't possible due to the `require_version`.

Also we may want to change the demo's blueprint file to also link to the PyGObject docs (or remove the GJS link), what do you think @sonnyp?

rolandlo:
> Thanks! In general this looks good and works perfectly, however could you make sure your code is formatted to follow [PEP 8](https://peps.python.org/pep-0008/) and uses 4 spaces for indentation?
> 
> You can use `pycodestyle` to verify. You can ignore error `E402 module level import not at top of file`, since that isn't possible due to the `require_version`.
> 

Done! It would be good to add these style requirements to the [style guide](https://github.com/sonnyp/Workbench/wiki/Style-Guide), so everybody can learn about them without going through previous PR's.

> Also we may want to change the demo's blueprint file to also link to the PyGObject docs (or remove the GJS link), what do you think @sonnyp?

I have added a commit adding that link. Let me know if I should remove the commit (or make changes to it).

Back to:

> The `open_uri` actions somehow don't work yet. Is this something that needs to be fixed on the "core" side or is there something wrong with my code?

Is it only for me that the Linkbuttons from the bookmarks don't work (the links are not opened) with the Python code and the Actions Documentation section is disabled?
grafik

theCapypara:
Oh, that's very odd, I somehow didn't even notice that the first time around. 
Must be an issue in PyGObject, somehow? It's really odd since there's no code in either the JS/Python demo related to that, but I can't really imagine how the language bindings would break this.

grafik

However looking at the inspector it seems the "sensitive" property for the generated ModelButtons is somehow filled with the label...

theCapypara:
Oh wow, wait, disregard my previous comment even though I double and tripple checked before sending (and was very confused how this is even possible) I just was off by one line. That is the text.

theCapypara:
Okay this is not a bug in PyGObject, this is a general issue with the way the external previewers work at the moment: https://github.com/sonnyp/Workbench/issues/715

theCapypara (approved):

rolandlo:
Should I squash all the commits (or at least the first two commits) into one?

theCapypara:
IMO we could still merge this. There are already some other demos broken right now because of #715 / #110

rolandlo:
Ok then, let's squash and merge.

PR #711, opened 2023-10-28
Extensions: Add button to install Sdk Extensions

lw64:


lw64:
grafik

lw64:
grafik

Hofer-Julian:
> Needs a designer input :)
> 
> The Rust case where 2 buttons are needed isn't worth it IMHO. I think we should remove the buttons there until we figure out a way to use mold with ggc instead of llvm.
> 
> @Hofer-Julian agree?
> 
> @lw64 in which case the code would be too complicated - we can make a special case for Vala only

I would prefer buttons everywhere or nowhere. I also don't really see the benefits of having two buttons just because there are two instructions tbh

lw64:
> I would prefer buttons everywhere or nowhere. I also don't really see the benefits of having two buttons just because there are two instructions tbh

Each sdk extension is a separate `appstream://` uri, there is no one that does both at the same time. Thats why both are needed. Maybe the second one only appears when the first sdk is already installed, so you can do it one after the other? But that's also a bit confusing.

lw64:
Looks good to me!

lw64:
Thanks! :)

PR #718, opened 2023-10-29
More python examples

gregorni:


gregorni:
So after talking with @sonnyp on Matrix, I'm going to do one PR per demo. I'd like to keep this list here though, so that I can keep track of them myself.

theCapypara:
In that case I will mark this as draft and remove reviews. Please close this when you don't need it anymore.

gregorni:
> In that case I will mark this as draft and remove reviews. Please close this when you don't need it anymore.

Ah, you don't have to do that, just merge ahead. I'll find this either way.

theCapypara:
Wait, but you said you were going to open separate PRs. In that case this PR is not meant to be merged.

gregorni:
Well, we said since it's just these five demos, it's fine, but with everything else I'll do per-demo PRs.

src/Library/demos/Avatar/main.py
```diff
@@ -0,0 +1,31 @@
+import gi
```
theCapypara, line 1: For some reasons the Python demo does not load the default avatar image for me, showing just a placeholder. But I can see in the inspector that the correct icon is placed there, so either something is wrong in my dev setup or this is an issue with the previewer process or PyGObject.

theCapypara (approved):

Issue #716, opened 2023-10-29
Implement Python code formatting


sonnyp:
With JavaScript and Rust (and soon Vala) we do 2 things:

1. Workbench auto formats the code when pressing "Run"
2. We have a pre-commit hook that auto formats staged files (see `lint-staged` in `package.json` )

We should implement the same thing for Python so that

1. We don't ask contributors to make bothersome code formatting changes
2. Improve the experience of Workbench user

PR #719, opened 2023-10-29
Library: Port 'Carousel' to Python

gregorni:


theCapypara:
Thanks!
Just a note: This is blocked until #716 is done (prepare for some re-formatting :stuck_out_tongue: :sweat_smile:).

theCapypara:
Note: I pushed a commit to apply black formatting now that #716 is done.

gregorni:
Thanks! I think this is done then.

By the way, I used a global variable once in the example because it fixes a problem that I couldn't solve otherwise. If you know a better solution, feel free to correct.

src/Library/demos/Carousel/main.py
```diff
@@ -0,0 +1,67 @@
...
+from gi.repository import Adw, Gtk
+import workbench
+
+root_box = workbench.builder.get_object("root_box")
+carousel = workbench.builder.get_object("carousel")
+ls_switch = workbench.builder.get_object("ls_switch")
+sw_switch = workbench.builder.get_object("sw_switch")
+indicator_row = workbench.builder.get_object("indicator_row")
+orientation_row = workbench.builder.get_object("orientation_row")
+indicators = None
```
theCapypara, line 15: This should not be needed. Can you check and remove?

src/Library/demos/Carousel/main.py
```diff
@@ -0,0 +1,67 @@
...
+    indicators = Adw.CarouselIndicatorDots(carousel=carousel)
+else:
+    indicators = Adw.CarouselIndicatorLines(carousel=carousel)
+
+indicators.set_orientation(carousel.get_orientation())
+root_box.append(indicators)
+
+
+def on_indicator_selected(_widget, _item):
+    global indicators
```
theCapypara, line 43: This is OK in this case, you need to do this since you are below declaring a local variable with the same name as the global one. If you did not have a local variable with the same name, Python would automatically use the global. 
But since you do set a variable "indicators" below Python assumes all usage of that variable to be local, giving you an error about undefined locals on the next line. By declaring indicators as global you explicitly tell Python that your setting of indicators is not declaring a local variable, it is modifying the global one and all usages of indicators are again referring to that global variable.

In real Python code you'd probably avoid doing this to avoid confusion, but since this is a small demo and we want to keep these close to the JavaScript demos this is OK. Personally I'd probably use something like a partial to pass the indicators into the function.

gregorni: >  Personally I'd probably use something like a partial to pass the indicators into the function.

So do I keep it as is, or do this?

theCapypara: I would keep it like is for now to stay closer to the JS demo.

gregorni: Then I think this can be merged.

theCapypara (approved):

PR #720, opened 2023-10-29
ci: Run black to format Python code

### LINKED ISSUE #716: Implement Python code formatting

theCapypara:
I hope the README.md I added is OK, it's a bit of an unfortunate situation but it wasn't too bad to get it working.

Closes #716

### DIFF
--- a/black.sh
+++ b/black.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+# This is a small wrapper around black - the Python code formatter.
+# The script will setup and activate a virtual env in .venv if black is not
+# found in PATH.
+#
+# Still requires Python 3.
+set -e
+
+if ! which black &>/dev/null; then
+  if [ ! -d ".venv" ]; then
+    echo "Note: black was not found on your system. A virtualenv is created in .venv."
+    python3 -m venv .venv
+  fi
+  source .venv/bin/activate
+  if ! which black &>/dev/null; then
+    pip3 install black
+  fi
+fi
+
+exec black "$@"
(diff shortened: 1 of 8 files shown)


black.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
theCapypara, line 1: This is for local execution. See package.json.

src/Library/demos/Welcome/main.py
```diff
@@ -7,10 +7,7 @@
```
theCapypara, line 7: This is the pre-commit hook doing it's magic :)

black.sh
```diff
@@ -0,0 +1,20 @@
...
+    echo "Note: black was not found on your system. A virtualenv is created in .venv."
+    python3 -m venv .venv
+  fi
+  source .venv/bin/activate
+  if ! which black &>/dev/null; then
+    pip3 install black
+  fi
+fi
+
+exec black "$@"
```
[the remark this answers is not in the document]
theCapypara: Would probably be useful to have in the SDK. I feel like pre-building the Workbench Flatpak is a bit convoluted and error prone.

PR #723, opened 2023-10-29
Move PyGObject url to Bookmarks

sonnyp:


theCapypara (approved):
Maybe we should move the GJS link as well then? It's just a bit odd seeing it there so prominently when running the demo via Python.

Issue #713, opened 2023-10-28
Unable to run Python demos with SourceView in the ui file


UrtsiSantsi:
(process:101): Gtk-ERROR **: 01:30:56.799: failed to add UI: .:8:1 Invalid object type 'GtkSourceView'

theCapypara:
I can't reproduce this, are you using the Flatpak based development environment and did you import GtkSource?
```py
import gi
gi.require_version("GtkSource", "5")
from gi.repository import GtkSource

print(GtkSource.View)
```
prints
```
<class 'gi.repository.GtkSource.View'>
```

theCapypara:
If you are running Workbench via GNOME Builder it will use the Flatpak. 

That is odd, so with that exact snippet I shared you get an error?

theCapypara:
Why do you have a `$` in front of GtkSourceView? I'm pretty sure it should be just `Gtk.SourceView`, also note the `.`.

PR #724, opened 2023-10-29
Library: Port 'CSS Gradients' to Python

UrtsiSantsi:
Marking as draft, since it doesn't work for me - see #713

### DIFF
--- a/src/Library/demos/CSS Gradients/main.py
+++ b/src/Library/demos/CSS Gradients/main.py
@@ -19,7 +19,7 @@ gtksource_buffer = workbench.builder.get_object("gtksource_buffer")
 button_copy_css = workbench.builder.get_object("button_copy_css")
 
 
-def generateCss():
+def generate_css():
     angle_string = spin_row_angle.get_value()
     first_color_string = button_color_1.get_rgba().to_string()
     second_color_string = button_color_2.get_rgba().to_string()
@@ -59,7 +59,7 @@ def generateCss():
     return css.lstrip()
 
 
-def updateCssProvider(css):
+def update_css_provider(css):
     display = Gdk.Display.get_default()
 
     css_provider = Gtk.CssProvider()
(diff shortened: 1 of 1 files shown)


theCapypara:
I'm not sure I understand, your commit passed CI. Black expects a new line at the end and your file has one.

src/Library/demos/CSS Gradients/main.py
```diff
@@ -0,0 +1,110 @@
...
+)
+spin_row_angle = workbench.builder.get_object("spin_row_angle")
+button_color_1 = workbench.builder.get_object("button_color_1")
+button_color_2 = workbench.builder.get_object("button_color_2")
+button_color_3 = workbench.builder.get_object("button_color_3")
+gtksource_buffer = workbench.builder.get_object("gtksource_buffer")
+button_copy_css = workbench.builder.get_object("button_copy_css")
+
+
+def generateCss():
```
theCapypara, line 22: Please use snake case names (`generate_css`)

src/Library/demos/CSS Gradients/main.py
```diff
@@ -0,0 +1,110 @@
...
+    angle_string = spin_row_angle.get_value()
+    first_color_string = button_color_1.get_rgba().to_string()
+    second_color_string = button_color_2.get_rgba().to_string()
+    third_color_string = button_color_3.get_rgba().to_string()
+
+    css = ""
+    if combo_row_gradient_type.get_selected() == 0:
+        css = """
```
theCapypara, line 30: This (and the others) are not valid format strings. You want something like:
```py
f""""
{angle_string}deg
""""
```

theCapypara (changes requested):
There really seems to be something wrong with importing SourceView, I can reproduce this in your example. It might be that we need to make sure GtkSource is loaded in the previewer too, since it's defined in the Builder XML, I did not check that yet. Will have a look.

gregorni:
@UrtsiSantsi Is there any way I can contact you outside GitHub? We're both working on Python demos right now, in parallel, so I thought it would be smart if we talked together to make sure we don't accidentally port the same demo.

If you're on Matrix, you can find me under https://matrix.to/#/@gregorni:gnome.org

theCapypara:
I'll add the Adw and other initialization code to the previewer as well, but please also keep it in the demo, that's good to have for people that try to copy the demos into their own apps.

theCapypara (approved):

PR #730, opened 2023-10-30
Library: Port 'Toolbar View' to Python

UrtsiSantsi:


src/Library/demos/Toolbar View/main.py
```diff
@@ -0,0 +1,74 @@
...
+    global bottom_bar
+    new_bottom_bar = workbench.builder.get_object(name)
+    if bottom_bar:
+        toolbar_view.remove(bottom_bar)
+    toolbar_view.add_bottom_bar(new_bottom_bar)
+    bottom_bar = new_bottom_bar
+
+
+def selectTopBar():
+    match topbar_select.get_selected():
```
theCapypara, line 35: The match statement isn't really meant to be used like switch from other languages, it's pretty slow from what I heard. In general in Python you would probably use if-else chains. But to be honest I prefer this much more too for a demo and the match statement is so new, I couldn't really find any real words on if this is a good idea to use it like this. The PEP just says:

> That last feature helps with some kinds of code which look more like the “switch” statement as present in other languages:
> [...]
> Although this will work, it’s not necessarily what the proposal is focused on, and the new syntax has been designed to best support the destructuring scenarios.

theCapypara: No you can keep it. It takes a while to go through all and I can only do a few per day at the moment.

theCapypara (approved):
LGTM! There were a few minor code styling issues, I added a commit for those.

PR #732, opened 2023-10-30
Library: Port 'Toasts' to Python

UrtsiSantsi:


src/Library/demos/Toasts/main.py
```diff
@@ -0,0 +1,54 @@
+import gi
+
+gi.require_version("Gtk", "4.0")
+gi.require_version("GLib", "2.0")
+gi.require_version("Gio", "2.0")
```
gregorni, line 5:
```suggestion
```

`GLib` and `Gio` don't require specifying the version, only `Gtk` and `Adw` do.

theCapypara: Don't think I did, it's only required for libraries that have more than one version (Gtk, GtkSource, etc.) and it's recommended for Adw.

It's also not needed for direct dependencies of Gtk, such as Gdk. 
You can take Workbench itself as a reference for what needs versions: https://github.com/sonnyp/Workbench/blob/main/src/init.js#L1-L7

gregorni: I usually just leave them out until Python complains, and it's only ever complain with Gtk and Adw for me.

theCapypara (changes requested):
Hi! Thank you! The demo unfortunately does not launch for me:

```py

Traceback (most recent call last):
  File "/app/share/re.sonny.Workbench.Devel/gdbus_ext.py", line 319, in on_method_call
    result = method_func(obj, *args)
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/share/re.sonny.Workbench.Devel/python-previewer.py", line 135, in run
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/home/marco/.var/app/re.sonny.Workbench.Devel/data/re.sonny.Workbench.Devel/sessions/2023-11-11 17-13-05/main.py", line 52, in <module>
    workbench.window.add_action(action_console)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Window' object has no attribute 'add_action'

```

That part is probably not supportable yet in Python either, see also the Vala code and the issue it's refering: https://github.com/sonnyp/Workbench/issues/110

theCapypara:
Oh, hm! Sorry about that, I think GitHub didn't show me your comments yesterday :monocle_face:

PR #735, opened 2023-10-30
Library: Port 'Text Colors' to Python

UrtsiSantsi:


theCapypara:
Very sorry about the CI issues, that was another demo that caused this. Will be fixed by #736, you may want to update your branches after this.
With all these many PRs one slipped through that had bad formatting.

src/Library/demos/Text Colors/main.py
```diff
@@ -0,0 +1,60 @@
+# Pango is a text layout library. It can e.g. be used for formatting text
+# https://gjs-docs.gnome.org/pango10~1.0/
+import gi
+
+gi.require_version("Gtk", "4.0")
+gi.require_version("Pango", "1.0")
```
theCapypara, line 6: Specifying the version for Pango is not necessary, there is only one version.

theCapypara (changes requested):

src/Library/demos/Text Colors/main.py
theCapypara: Oh sorry, I wasn't aware Pango 2 was a thing. In that case we should keep this :+1: 
See my comment here: https://github.com/sonnyp/Workbench/pull/732#discussion_r1382240886

theCapypara: See also #738 - I'll think about / do some research if specifying the Pango version makes sense, but in doubt we keep it.

theCapypara (approved):
LGTM! There were a few minor code styling issues, I added a commit for those.

There was also a debug print still in there, removed that as well.

PR #743, opened 2023-10-30
Library: Port 'Dialogs' to Python

gregorni:
I had to deviate from the JS demo a bit because Python can't handle async as nicely

### DIFF
--- a/data/app.gschema.xml
+++ b/data/app.gschema.xml
@@ -26,6 +26,9 @@
     <key name="show-style" type="b">
       <default>false</default>
     </key>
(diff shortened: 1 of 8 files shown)


gregorni:
Terribly sorry, I don't know what's causing that. I'll try to fix it anyways

gregorni:
@sonnyp It seems using merge to pull fixes the problem, as opposed to rebasing. That just means you'll need to squash all of my merge requests.

theCapypara (approved):
Thanks, LGTM except for some styling issues that I fixed in a commit I added.

PR #744, opened 2023-10-30
Library: Port 'Flow Box' to Python

gregorni:
Affected by #745

### DIFF
--- a/src/Library/demos/Flow Box/main.py
+++ b/src/Library/demos/Flow Box/main.py
@@ -34,4 +34,3 @@ def on_item_selected(_flowbox, item):
 
 
 flowbox.connect("child-activated", on_item_selected)
-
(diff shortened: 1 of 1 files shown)


src/Library/demos/Flow Box/main.py
```diff
@@ -0,0 +1,36 @@
...
+
+gi.require_version("Gtk", "4.0")
+gi.require_version("Adw", "1")
+from gi.repository import Adw, Gtk
+import workbench
+
+flowbox = workbench.builder.get_object("flowbox")
+
+
+def addEmoji(_flowbox, unicode):
```
theCapypara, line 11: Please use `snake_case`.

theCapypara (changes requested):
Thanks, LGTM and works, except one minor nitpick :)

src/Library/demos/Flow Box/main.py
gregorni: Ah, woops, that's some leftover JS.

theCapypara (approved):

PR #746, opened 2023-10-30
Library: Port 'Memory Monitor' to Python

gregorni:


src/Library/demos/Memory Monitor/main.py
```diff
@@ -0,0 +1,31 @@
...
+        print("Warning Level: Medium")
+    if level >= Gio.MemoryMonitorWarningLevel.CRITICAL:
+        # system will start terminating processes to reclaim memory
+        print("Warning Level: Critical")
+
+
+memory_monitor.connect("low-memory-warning", evaluate_memory)
+
+
+def drop_caches():
```
theCapypara, line 30: Weird to have this as an extra function but it's the same for JS so I think it's OK for consistentcy.

theCapypara (approved):
Thanks!!

PR #747, opened 2023-10-30
Library: Port 'Label' to Python

gregorni:


src/Library/demos/Label/main.py
```diff
@@ -0,0 +1,61 @@
...
+Ut enim mortis metu omnis quietae vitae status perturbatur,
+et ut succumbere doloribus eosque humili animo inbecilloque ferre miserum est,
+ob eamque debilitatem animi multi parentes, multi amicos, non nulli patriam,
+plerique autem se ipsos penitus perdiderunt, sic robustus animus et excelsus omni."""
+
+label.set_label(short_label)
+
+single_line_switch.connect(
+    "notify::active",
+    lambda *_: label.set_label(
```
theCapypara, line 38: I think this is a bit too complicated for a lambda, please use a function and proper if-else.

theCapypara (changes requested):

theCapypara (approved):
LGTM, thanks!

PR #750, opened 2023-10-30
Library: Port 'Status Page' to Rust

UrtsiSantsi:


Hofer-Julian (commented):
Thanks @UrtsiSantsi the code looks good to me.

However, I think the demo is broken in general.
In both Rust and Javascript, removing the code doesn't change the behaviour.
With the code it gives the following warning:

```
Adwaita-Critical: adw_status_page_set_child: assertion 'gtk_widget_get_parent (child) == NULL' failed
```

@sonnyp could you have a look since you've been the last one to touch this demo?

PR #751, opened 2023-10-30
Library: Port 'Menu' to Python

gregorni:


theCapypara (approved):
LGTM. The differences from the JS demos make sense for Python (and are arguably better).

PR #758, opened 2023-10-31
Library: Port 'Spell Checker' to Python

UrtsiSantsi:


theCapypara (approved):
LGTM! The demo works just as well as the JS demo. The issue with the underline is (probably?) a bug in libspelling.

Issue #745, opened 2023-10-30
Preview window doesn't close sometimes


gregorni:
This shows up with the current state of #744 and #742. When you run the demo, the preview window opens, but then you can't close it, even if you close all other Workbench windows. 

When running such previews, there's a warning message in the terminal:
```
(process:196): Gtk-WARNING **: 18:11:28.728: A window is shown after it has been destroyed. This will leave the window in an inconsistent state.
```

theCapypara:
Looking at the affected demos, this is most likely an issue with demos that have windows as top-level components (meaning the previewer doesn't explicitly have to create it).

PR #760, opened 2023-10-31
python: Fix initializations

theCapypara:
This makes sure `Gtk` and `Adw` are really properly initialized. 
I haven't found any init calls in any of the other demos or Workbench itself, so this should be all.


theCapypara:
This may also fix #745, but I'm not able to reproduce that one.

PR #763, opened 2023-10-31
Manual: Add forward and backward mouse gestures.

huanie:
I have a mouse that has these buttons on the side. Many applications such as browsers use them for going forward and backwards.

I've tried adding the GestureClick to the document_viewer template but it only added the GestureClick to the sidebar for some reason.


huanie:
For some reason I need to double click these buttons (only when mouse is in webview) when I try to change it in the standalone app. Also I've noticed that the shortcuts only work when the mouse is in the sidebar. Is the Webview maybe at fault here?

PR #770, opened 2023-11-01
library: Port Memory Monitor demo to Vala

Diego-Ivan:
Ports Memory Monitor demo to Vala

### DIFF
--- a/src/Library/demos/Memory Monitor/main.vala
+++ b/src/Library/demos/Memory Monitor/main.vala
@@ -10,7 +10,10 @@ public void main () {
 
   var memory_monitor = MemoryMonitor.dup_default ();
   memory_monitor.low_memory_warning.connect ((level) => {
-    // Use inequalities for checking as new levels may be added in the future
+    /*
+     * Use inequalities for checking as new levels may be added in the future
+     * See https://valadoc.org/gio-2.0/GLib.MemoryMonitorWarningLevel.html
+     */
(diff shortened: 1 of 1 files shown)


src/Library/demos/Memory Monitor/main.vala
```diff
@@ -0,0 +1,34 @@
...
+
+public void main () {
+  cache = new HashTable<string, int?> (string.hash, str_equal);
+  cache["a"] = 1;
+  cache["b"] = 2;
+  cache["c"] = 3;
+
+  var memory_monitor = MemoryMonitor.dup_default ();
+  memory_monitor.low_memory_warning.connect ((level) => {
+    // Use inequalities for checking as new levels may be added in the future
```
lw64, line 13: I am not sure if inequalities are appropriate here. Its not immediately understandable, what the inequality means here, and people shouldnt need to look up the API docs for the enum, to see what other entries exist. Also if its just `==`'s then, maybe a switch statement would be better suited? Not sure...
The switch statement could be missing break's to still print the other warnings.

Diego-Ivan: It is a recommendation from the enum docs: https://docs.gtk.org/gio/enum.MemoryMonitorWarningLevel.html. I agree that it isn't obvious, so a link to Valadoc would be useful in this case.

lw64 (approved):

PR #773, opened 2023-11-01
library: Allow the user to a folder in Column View demo

Diego-Ivan:
The demo currently uses pkg.pkgdatadir to resolve a folder path, which makes it impossible to port to other languages. By allowing the user to choose a folder themselves, we solve this issue as well as making the demo a little more interactive.

### DIFF
--- a/src/Library/demos/Column View/main.js
+++ b/src/Library/demos/Column View/main.js
@@ -1,6 +1,5 @@
 import Gtk from "gi://Gtk";
 import GLib from "gi://GLib";
-import Gio from "gi://Gio";
 import GObject from "gi://GObject";
 
 const column_view = workbench.builder.get_object("column_view");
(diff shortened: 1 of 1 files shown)


andyholmes:
> I wonder if we shouldn't use a custom model with a bunch of static or generated data instead.
> 
> Introducing fs in this demo isn't very useful. @andyholmes WDYT?

I agree, some static models (or at least pre-populated) would be useful for a lot of demos. I think the demos should focus on the subject matter to be as clear as possible.

Diego-Ivan:
> Did you forgot the blp changes?

Just noticed I did. Sorry about it, just pushed the changes too.

> I agree, some static models (or at least pre-populated) would be useful for a lot of demos. I think the demos should focus on the subject matter to be as clear as possible.

Sounds good to me. I was thinking about using an Object subclass similar to what is used in the Drop Down demo to list something like books and sort them by title/author/year.

Diego-Ivan:
Just made the changes :smile:

PR #789, opened 2023-11-05
library: Port Column View demo to Vala

Diego-Ivan:
Ports the Column View demo to Vala

### DIFF
--- a/src/Library/demos/Column View/main.vala
+++ b/src/Library/demos/Column View/main.vala
@@ -21,9 +21,7 @@ public void main () {
   var col2 = (Gtk.ColumnViewColumn) workbench.builder.get_object ("col2");
   var col3 = (Gtk.ColumnViewColumn) workbench.builder.get_object ("col3");
 
-  // Create the data model
-  var data_model = new ListStore (typeof (Book));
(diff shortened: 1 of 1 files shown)


src/Library/demos/Column View/main.vala
```diff
@@ -0,0 +1,101 @@
...
+
+public void main () {
+  var column_view = (Gtk.ColumnView) workbench.builder.get_object ("column_view");
+  var col1 = (Gtk.ColumnViewColumn) workbench.builder.get_object ("col1");
+  var col2 = (Gtk.ColumnViewColumn) workbench.builder.get_object ("col2");
+  var col3 = (Gtk.ColumnViewColumn) workbench.builder.get_object ("col3");
+
+  // Create the data model
+  var data_model = new ListStore (typeof (Book));
+  data_model.splice (0, 0, {
```
lw64, line 26: Can you put a comment here to explain the usage of `splice()`? I think its not that obvious. Maybe even using `append()` in a foreach loop on an array would be better instead.

lw64 (changes requested):

lw64 (approved):

PR #791, opened 2023-11-05
library: Port HTTP Request demo to Vala

Diego-Ivan:
Ports the HTTP Request demo to Vala

### DIFF
--- a/src/Library/demos/HTTP Request/main.vala
+++ b/src/Library/demos/HTTP Request/main.vala
@@ -28,63 +28,55 @@ private async void fetch_wikipedia_todays_featured_article () throws Error {
     @"https://api.wikimedia.org/feed/v1/wikipedia/$language/featured/$(date.format ("%Y/%m/%d"))";
 
   var message = new Soup.Message ("GET", url);
-  InputStream input_stream = yield http_session.send_async (message, Priority.DEFAULT, null);
 
+  Bytes message_bytes = yield http_session.send_and_read_async (message, Priority.DEFAULT, null);
   if (message.status_code != 200) {
     throw new FetchError.FAILED_REQUEST (@"Failed Request. HTTP Status: $(message.status_code)");
   }
 
-  string json_data = yield read_string (input_stream);
+  unowned uint8[] data = message_bytes.get_data ();
+  string json_data = (string) data;
   parse_json_response (json_data);
 }
 
-private async string read_string (InputStream input_stream) throws Error {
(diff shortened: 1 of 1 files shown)


src/Library/demos/HTTP Request/main.vala
```diff
@@ -0,0 +1,90 @@
...
+  var date = new DateTime.now_local ();
+  var http_session = new Soup.Session ();
+
+  // https://api.wikimedia.org/wiki/Feed_API/Reference/Featured_content
+  string language = "en";
+  string url =
+    @"https://api.wikimedia.org/feed/v1/wikipedia/$language/featured/$(date.format ("%Y/%m/%d"))";
+
+  var message = new Soup.Message ("GET", url);
+  InputStream input_stream = yield http_session.send_async (message, Priority.DEFAULT, null);
```
lw64, line 31: I think its way easier to use `send_and_read_async()` than using an `InputStream`. Was there a specific reason to do so?

src/Library/demos/HTTP Request/main.vala
```diff
@@ -0,0 +1,90 @@
...
+}
+
+private void parse_json_response (string json) throws Error {
+  // https://valadoc.org/json-glib-1.0/Json.Parser.html
+  var parser = new Json.Parser ();
+  parser.load_from_data (json);
+
+  // https://valadoc.org/json-glib-1.0/Json.Reader.html
+  Json.Node root_node = parser.get_root ();
+  var reader = new Json.Reader (root_node);
```
lw64, line 61: Can you provide a rough structure of the recieved json document in a comment, so the parsing is easier to understand. Also the cursor based API isnt intuitive, so a short explaination how it works would be good I think. (or maybe just accessing it normally with nodes would be easier?)

lw64 (changes requested):

src/Library/demos/HTTP Request/main.vala
Diego-Ivan: Just for consistency with the JS demo, as it uses `send_async` too and reads the response using an `InputStream`, but yeah, `send_and_read_async` is way easier and probably what should be done in the first place. I will use `send_and_read_async` and send a PR to use it in the JS demo later.

lw64 (approved):

src/Library/demos/HTTP Request/main.vala
lw64: Thanks! That is of course the best solution :)

PR #795, opened 2023-11-05
library: Port Web View demo to Vala

Diego-Ivan:
Ports the Web View demo to Vala

### DIFF
--- a/src/Library/demos/Web View/main.vala
+++ b/src/Library/demos/Web View/main.vala
@@ -34,8 +34,6 @@ public void main () {
       case FINISHED:
         message ("Loading page finished");
         break;
-      case COMMITTED:
-      case REDIRECTED:
       default:
         break;
     }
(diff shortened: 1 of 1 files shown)


src/Library/demos/Web View/main.vala
```diff
@@ -0,0 +1,76 @@
...
+
+  web_view.load_changed.connect ((load_event) => {
+    switch (load_event) {
+      case STARTED:
+        message ("Loading page started");
+        break;
+      case FINISHED:
+        message ("Loading page finished");
+        break;
+      case COMMITTED:
```
lw64, line 37: These could just be left out, as they don't do anything, right? Or are they for demonstrative purposes?

Diego-Ivan: I included them because otherwise the compiler would throw a warning saying that there are missing enum values in the switch statement. I preferred to add them to avoid confusion from the users.

lw64: In that case add a `default` case.

lw64: But to replace the FINISHED and COMMITTED cases :)

Diego-Ivan: Ohhhh that makes sense, sorry then haha :)

lw64 (approved):

PR #805, opened 2023-11-14
library: Port Avatar demo to Rust

theCapypara:


src/Library/demos/Avatar/code.rs
```diff
@@ -0,0 +1,38 @@
...
+    button.connect_clicked(move |_| {
+        dialog.open(
+            Some(workbench::window()),
+            None::<&gio::Cancellable>,
+            on_selected,
+        )
+    });
+}
+
+fn on_selected(result: Result<gio::File, glib::Error>) {
```
Hofer-Julian, line 29:
```suggestion
    button.connect_clicked(move |_| {
        dialog.open(
            Some(workbench::window()),
            None::<&gio::Cancellable>,
            on_clicked,
        )
    });
}

fn on_clicked(result: Result<gio::File, glib::Error>) {
```

To use the same naming as the other demos

Hofer-Julian (commented):
Apart from that one comment, it looks good to me!

PR #818, opened 2023-11-18
Library: Port 'Accessibility' to Python

gregorni:


theCapypara (changes requested):
on_key_released does not work:
```
TypeError: on_key_released() takes 2 positional arguments but 4 were given
```
https://docs.gtk.org/gtk4/signal.EventControllerKey.key-released.html

gregorni:
Fixed now. Thanks for the link!

theCapypara (approved):

theCapypara:
Thanks!

PR #819, opened 2023-11-18
Library: Port 'Level Bars' to Python

gregorni:


theCapypara:
@sonnyp There seems to be an issue with either the JavaScript or demo, I think the JS demo has the CSS not applied, but looking at the classes in the inspectors for the battery demo they don't match the classes in the CSS in either Python / JS, so I'm not sure how it's supposed to be.

JS:
grafik

Python:
grafik

Since the Python demo actually changes color, I think that one is probably correct?

theCapypara:
Sure!

PR #821, opened 2023-11-18
Fix various issues with the previewers

### LINKED ISSUE #700: "Open Preview Window" always uses internal previewer
When the "Open Previewer Window" button is pressed while using a language that requires the external previewer, Workbench will just launch the demo in the internal previewer (presumably using the JavaScript demo code?)

theCapypara:
This fixes a bunch of related issues related to switching between previewers.
- Fixes #700
- When switching from a demo using the external previewer to internal previewer in inline mode, the stack is now correctly switched to the "inline" child item
- The button to open the external previewer demo is now no longer shown when switching between languages until the new language was "Run" once (ie. the previewer and window were created)*
- The Vala and Python previewer now don't destroy the preview window when closing it, this allows for the "Show Previewer Window" button to re-open the windows

There are probably still a few issues left, we should consider refactoring the previewer logic to not have so many different bits and pieces that could lead to inconsistent states. Maybe add some sort of central controlling interface that makes sure the previewer is always in a consistent state?

*: There is still a minor issue with this: When opening and running another demo using the external previewer, and then clicking the "Show Preview Window" button in the old but still-opened library window, this will instead show the new demo window, since the external previewer always just has one demo loaded. This is probably minor enough to not worry about until a potential re-write.


src/Previewer/previewer.vala
```diff
@@ -25,12 +25,13 @@ namespace Workbench {
     private void set_window(Gtk.Window window) {
       this.window?.destroy ();
       this.window = window;
+      // Make sure the preview can be re-opened by using the "Show Preview Window" button.
+      this.window.set_hide_on_close(true);
```
lw64, line 29: `this.window.hide_on_close = true;`

src/langs/python/python-previewer.py
```diff
@@ -216,11 +216,12 @@ def set_window(self, the_window: Gtk.Window):
         if self.window is not None:
             self.window.destroy()
         self.window = the_window
+        # Make sure the preview can be re-opened by using the "Show Preview Window" button.
+        self.window.set_hide_on_close(True)
```
lw64, line 220: Here too I guess. Always use the property when its available.

lw64 (changes requested):

src/langs/python/python-previewer.py
theCapypara: With PyGObject it seems to be preferred to use the methods, probably because it's ambiguous if you are setting a Python or GObject property. But I'll update the Vala code!

lw64 (approved):
Thanks!

theCapypara:
I'm closing this, as this has been obsoleted as far as I understand, but feel free to re-open.

PR #822, opened 2023-11-18
Remove Icon Library

### LINKED ISSUE #808: Remove Icon Library
Let's remove the builtin Icon Library in favor of the standalone "Icon Library".

https://flathub.org/apps/org.gnome.design.IconLibrary

We will need a way to "import" the icon into Workbench somehow

gregorni:
Closes #808 

My strategy isn't fool-proof: I just `grep`ed for "icon", and removed everything related to the Icon Library. There might still be remains of it that I missed, but I believe there's no way of accessing it anymore.


gregorni:
Should that happen in a separate PR or in this one?

gregorni:
> Note; in this PR you can also remove the "icon-development-Kit" module.

I tried doing that, but then the Image demo breaks, and some of the icons in the UI go missing. I think it might be easier to keep the development kit than hand-picking, saving, and including the specific ones we need.

gregorni:
That sounds like work for a separate PR

PR #823, opened 2023-11-18
Composite template support for Vala, Rust, Python + "Custom Widget" demo

theCapypara:
This is my attempt to get composite templates working for the other langs.

I say attempt, because it doesn't actually work. There are a lot of reasons, and it boils down to that this probably would require approaching the external previewers completely different.

The main remaining/blocking issue is the fact that the class names need to be counted up to avoid collisions. but Python, Vala and Rust expect the class name to match exactly what's in the demo while the current GJS demo doesn't seem to mind the mismatch.
I worked around that for Python by getting the gtype name from the Workbench API, but for the others this is much more complicated due to the fact that it needs to be compiled. I'm not sure how to get the modified class name in there. Maybe we need to just not add a number to it for those two and restart the previewer every time.
The Python code currently launches once. After that it doesn't anymore, since it tries to re-register the object. Modification would probably be needed to always count up on launch, not just when the XML changes, or the previewer would just also need to be restarted every time. In that case we should just get rid of the number for all external demos and restart the previewer on every launch if templates are involved.

Another issue is that the Vala/Rust code would need to modify the internal state of the previewer to set the "target", but since they are linked in as shared objects there's no easy way to communicate back here.

And I needed to change how the dummy target is handled, replacing it just with an empty box so that the Python and Vala previewers don't crash trying to initialize it's signal handlers. Weirdly this kinda breaks Blueprint, it's trying to display the internal previewer now for non-JS demos. But maybe that's not related.

### DIFF
--- a/src/Library/demos/Custom Widget/code.rs
+++ b/src/Library/demos/Custom Widget/code.rs
@@ -1,0 +1,66 @@
+use crate::workbench;
+use gtk::{glib, subclass::prelude::*};
+
+mod imp {
+    use super::*;
+
+    #[derive(Debug, Default, gtk::CompositeTemplate)]
+    // The file will be provided by Workbench when the demo compiles, it just contains the template.
+    #[template(file = "workbench_template.ui")]
+    pub struct AwesomeButton {}
+
+    #[glib::object_subclass]
+    impl ObjectSubclass for AwesomeButton {
+        const NAME: &'static str = "AwesomeButton";
+        type Type = super::AwesomeButton;
+        type ParentType = gtk::Button;
+
+        fn class_init(klass: &mut Self::Class) {
+            klass.bind_template();
+        }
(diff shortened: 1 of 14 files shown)


theCapypara:
Some solution ideas we had:

- Unregister the template type before the demo starts if it's already registered. This also means: Do not count up the class name for external previewer.
- Instead of having a "workbench.preview" API, expose the target widget as "workbench.target". Have that be an empty GtkBox for template demos. For regular demos have it just be the normal target.

theCapypara:
> there is no unregister equivalent to registerClass and 

Ah shame, that makes unregistering impossible then, shame. If we also don't want to restart then we have to come up with another solution. For Python we could probably monkey-patch Gtk.Template, but for the other two it wouldn't be so simple...

PR #824, opened 2023-11-20
library: Port 'Column View' to Python

theCapypara:
Draft because PyGObject does not support GtkExpression as a fundemental type yet:

```
    expression=Gtk.PropertyExpression.new(Book, None, "title"),
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: No means to translate argument or return value for 'GtkPropertyExpression'
```

Currently checking if this is solved by https://gitlab.gnome.org/GNOME/pygobject/-/merge_requests/250 and we just need to wait for a release (potentially until GNOME 46) or if there's extra work needed for that.

### DIFF
--- a/src/Library/demos/Column View/main.py
+++ b/src/Library/demos/Column View/main.py
@@ -1,0 +1,166 @@
+from typing import cast
+import gi
+
+gi.require_version("Gtk", "4.0")
+from gi.repository import GObject, Gio, Gtk
+import workbench
+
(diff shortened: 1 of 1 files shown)


theCapypara:
I checked with PyGObject's master branch, it works there. We'll have to wait for SDK/Platform 46 before being able to merge this.

theCapypara:
Replaced with https://github.com/workbenchdev/demos/pull/46

PR #826, opened 2023-11-20
library: Port 'List View Widget' to Python

theCapypara:
This is the first demo that would use `typing.cast`. What is your opinion in that?

I feel like with data driven APIs like this, and due to the fact it relies much less on the already established global objects than other demos, showing the user a good way to be able to properly type this is both useful to know for real world apps and also makes some parts of the demo easier to read, since it's now clear from the code how the different objects are related (eg. `ListItem.get_item` returning the `StringObject` because it's a `StringList`)

The downside is that casts can't really be checked statically. So in theory this could break, but in practice even if it would, the demo also wouldn't work properly then anyway.

### DIFF
--- a/src/Library/demos/List View Widgets/main.py
+++ b/src/Library/demos/List View Widgets/main.py
@@ -1,0 +1,91 @@
+from typing import cast
+import gi
+
+gi.require_version("Gtk", "4.0")
+from gi.repository import Gtk
+import workbench
+
+list_view = workbench.builder.get_object("list_view")
+grid_view = workbench.builder.get_object("grid_view")
+add = workbench.builder.get_object("add")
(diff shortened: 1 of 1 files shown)


theCapypara:
typing.cast is a no-op (identity function) at runtime.

theCapypara:
Right now the primary idea is to have a second language option "Python (typed)" that would use the normal Python source, and then "Python" would show the source stripped of type checks, ideally by an automated tool.

I will put this to draft for now.

theCapypara:
I'll probably end up doing that.

theCapypara:
I'll be back tomorrow and do some cleaning up, also have some PRs to review :)

PR #842, opened 2023-12-09
library: Add Rust demo Animation

Hofer-Julian:


Hofer-Julian:
Close as discussed at https://github.com/workbenchdev/Workbench/pull/843#issuecomment-1848368208

PR #843, opened 2023-12-09
Move demos to a separate repository

sonnyp:
https://github.com/workbenchdev/demos


Hofer-Julian:
@sonnyp I just opened this: https://github.com/workbenchdev/Workbench/pull/842
Should I wait before I merge it?

PR #844, opened 2023-12-09
Update gst-plugin-gtk4 to 0.11.2

sonnyp:
Fixes https://github.com/workbenchdev/Workbench/issues/825


gregorni (approved):
Works as expected using the 45.3 manifest

PR #854, opened 2023-12-20
Clean up error handling and GC in LSPClient,js

janvhs:
While working on the implementation for the Python LSP, I realized, that wrongly encoded messages trigger a lot of error messages. Furthermore, LSPs, who exit early or crash, trigger irrelevant error messages, due to their missing output.

This patch fixes this issue and prevents the GJS garbage collector from getting overwhelmed with recursive function calls on LSPClient._read, which can occur, if the message gets processed too fast.


janvhs:
Whop's, there were files, I did not intend to push, yet.

PR #856, opened 2023-12-26
lsp: Exit on unexpected errors

sonnyp:
Fixes https://github.com/workbenchdev/Workbench/pull/855

### DIFF
--- a/src/lsp/LSPClient.js
+++ b/src/lsp/LSPClient.js
@@ -54,7 +54,7 @@ export default class LSPClient {
     this.ready = true;
     this.emit("ready");
 
-    // For testing blueprint language server restart
+    // For testing language server restart
     // setTimeout(() => {
     //   this.stop()
     // }, 5000);
@@ -85,8 +85,8 @@ export default class LSPClient {
 
   async stop() {
     await Promise.all([
-      this.stdin.close_async(null),
-      this.stdout.close_async(null),
+      this.stdin.close_async(GLib.PRIORITY_DEFAULT, null).catch(console.error),
+      this.stdout.close_async(GLib.PRIORITY_DEFAULT, null).catch(console.error),
     ]);
     // this.proc?.force_exit();
     this.proc.send_signal(15);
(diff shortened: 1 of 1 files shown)


src/lsp/LSPClient.js
```diff
@@ -85,8 +85,8 @@ export default class LSPClient {
 
   async stop() {
     await Promise.all([
-      this.stdin.close_async(null),
-      this.stdout.close_async(null),
+      this.stdin.close_async(null).catch(console.error),
```
janvhs, line 88: Idk what argument is missing, but here is a link to the docs: https://gjs-docs.gjsify.org/classes/Gio_2_0.Gio.OutputStreamClass.html#close_async

```
(re.sonny.Workbench.Devel:5): re.sonny.Workbench.Devel-CRITICAL **: 14:15:28.821: TypeError: method Gio.OutputStream.close_async: At least 3 arguments required, but only 2 passed
_promisify/proto[asyncFunc]/<@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:450:35
_promisify/proto[asyncFunc]@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:448:16
stop@resource:///re/sonny/Workbench/lsp/LSPClient.js:88:18
_read@resource:///re/sonny/Workbench/lsp/LSPClient.js:220:19
async*_start_process@resource:///re/sonny/Workbench/lsp/LSPClient.js:164:10
start@resource:///re/sonny/Workbench/lsp/LSPClient.js:49:10
setup@resource:///re/sonny/Workbench/langs/javascript/javascript.js:13:8
JavaScriptDocument@resource:///re/sonny/Workbench/langs/javascript/JavaScriptDocument.js:10:17
Window@resource:///re/sonny/Workbench/window.js:81:31
openDemo@resource:///re/sonny/Workbench/Library/Library.js:105:26
Library/</<@resource:///re/sonny/Workbench/Library/Library.js:42:15
EntryRow/<@resource:///re/sonny/Workbench/Library/EntryRow.js:25:12
_init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:266:34
```

src/lsp/LSPClient.js
```diff
@@ -85,8 +85,8 @@ export default class LSPClient {
...
+      this.stdout.close_async(null).catch(console.error),
```
janvhs, line 89: same here

src/lsp/LSPClient.js
```diff
@@ -204,19 +204,28 @@ export default class LSPClient {
     }
 
     const str = decoder_utf8.decode(uint8);
-    return JSON.parse(str);
+    try {
+      return JSON.parse(str);
+    } catch (err) {
+      await this.stop();
```
janvhs, line 210: Stopping the LSP on an error is a good idea, but not re-throwing the error could potentially make debugging harder?

janvhs (changes requested):
From reading the source, stopping the LSP on error cases, seems to resolve the case where the GC is overwhelmed. I think it is a way better solution compared to what I proposed in #855 ! :)

There is however an error in the type signature, which prevents me from fully trying it.

Issue #804, opened 2023-11-14
Python: Linting


theCapypara:
Currently Python code is not checked against any linting rules. We should check the merged demos and determine some linter configurations that we think would make sense to enforce.

As a linter we should probably use [ruff](https://github.com/astral-sh/ruff), in that case we may also evaluate if it makes sense to replace black with it.

gregorni:
I'm for ruff. In my experience, it's very strict by default, but can be customized pretty easily using config files.

PR #857, opened 2023-12-26
Add pylsp as the python LSP.

janvhs:
This PR, lays the foundation to introduce an LSP for the python language.
Pylsp is packaged as a flatpak module and included in the build tree for the Devel version of Workbench.

Currently, pylsp responds to file input changing, but doesn't publish any diagnostics and I can not figure out why.
It seems to support the same capabilities as the JS client, yet only responds with empty diagnostics on the `notification::textDocument/publishDiagnostics` event.

Another thing, that raised my attention, is the usage of booth `python` and `python3`, when referencing to the language ID throughout the codebase. I don't think it is part of the problem, tho.

Can someone take a look at this and help to figure out, why it does not work?


theCapypara:
Which capabilities do we want?

From the README:
> [Rope](https://github.com/python-rope/rope) for Completions and renaming
> [Pyflakes](https://github.com/PyCQA/pyflakes) linter to detect various errors
>  [McCabe](https://github.com/PyCQA/mccabe) linter for complexity checking
>[pycodestyle](https://github.com/PyCQA/pycodestyle) linter for style checking
> [pydocstyle](https://github.com/PyCQA/pydocstyle) linter for docstring style checking (disabled by default)
>[autopep8](https://github.com/hhatto/autopep8) for code formatting
>  [YAPF](https://github.com/google/yapf) for code formatting (preferred over autopep8)
>  [flake8](https://github.com/pycqa/flake8) for error checking (disabled by default)
> [pylint](https://github.com/PyCQA/pylint) for code linting (disabled by default)

My concern is that this seems to use other tools for code formatting than we currently use.
We considered using Ruff in #804, Ruff has it's own LSP:
https://github.com/astral-sh/ruff-lsp

I think the idea is to combine it with python-lsp:
> ruff-lsp supports surfacing Ruff diagnostics and providing Code Actions to fix them, but is intended to be used alongside another Python LSP in order to support features like navigation and autocompletion.

So should we for now only install Rope for completions and add Ruff to replace Black in a new PR where we would also add ruff-lsp?

theCapypara:
I guess Workbench currently doesn't use completions. I'll still add Rope in case we want that in the future.
For now I'll also add Pyflakes for error checking, but we may want to replace that part with Ruff later.

theCapypara (approved):
For me this looks good for now! See my comments above for questions on how to extend this.

theCapypara:
Since I had time, I started creating a possible follow-up PR that would add Ruff: https://github.com/workbenchdev/Workbench/pull/902

theCapypara:
I removed rope again for now, since we don't support completions anyway yet.

janvhs:
@theCapypara thanks for moving it forward! ^^

@sonnyp I'm glad to do so :)

PR #858, opened 2023-12-27
Add Workbench CLI

sonnyp:
Add a Workbench CLI that uses the language servers to do format and lint.

We will use this for demos CI

image


src/Previewer/previewer.vala
```diff
@@ -1,244 +1,243 @@
...
-
-      // Toplevel windows returned by these functions will stay around
-      // until the user explicitly destroys them with gtk_window_destroy().
-      // https://docs.gtk.org/gtk4/class.Builder.html
-      if (target is Gtk.Window) {
-        ((Gtk.Window) target).destroy();
-      }
-    }
+// This registers GObjects so that they can be found by GtkBuilder
+        void ensure_types () {
```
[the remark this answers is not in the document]
lw64: what do you mean with bug? does it not get invoked?

src/cli/main.js
```diff
@@ -0,0 +1,399 @@
...
+    return 1;
+  }
+
+  if (lang.id === "vala") {
+    const api_file = (
+      GLib.getenv("FLATPAK_ID")
+        ? Gio.File.new_for_path(pkg.pkgdatadir)
+        : current_dir.resolve_relative_path("src/langs/vala")
+    ).get_child("workbench.vala");
+    api_file.copy(
```
lw64, line 39: Why is the `workbench.vala` file copied here?

lw64: vls doesn*t let you format code when there are diagnostics errors?

PR #865, opened 2024-01-14
Fix blank preview on no code demos

sonnyp:
Fixes https://github.com/workbenchdev/Workbench/issues/812
Closes https://github.com/workbenchdev/Workbench/pull/820

@theCapypara thanks a lot for the bug report and PR
Took me a while to wrap my head around what was going on. This is the culprit 

https://github.com/workbenchdev/Workbench/blob/58b40d9802d14cbcc613a4f25489858783b808b2/src/window.js#L268-L272C6 

For that particular code path (javascript selected, demo without js) it wouldn't update the preview

One of the issue is that demos without code shouldn't have `autorun: true`. I will fix that.
But in any case - I don't want Workbench to misbehave even if the demo json is incorrect.

This is a slightly different approach to what you suggested that instead separates `load` and `run` in a way that `load` is always called before `run`.

### DIFF
--- a/src/Library/Library.js
+++ b/src/Library/Library.js
@@ -92,16 +92,23 @@ async function openDemo({ application, demo_name, language }) {
     session.settings.set_boolean("show-code", true);
   }
 
+  const { autorun, languages } = demo;
+
   // Override the user preferred language if the demo doesn't support it
   const lang = session.getCodeLanguage();
-  if (demo.languages.length > 0 && !demo.languages.includes(lang.id)) {
+  if (languages.length > 0 && !languages.includes(lang.id)) {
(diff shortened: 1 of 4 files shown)


theCapypara (approved):
Oh this is much better. 
I was missing the insight into the inner workings of how it all works to properly do that, but I think separating them is exactly the way to go.

You may want to check if that cleanup I did for the console clearing is still necessary with this. When I did the original PR I noticed that compiler errors would immediately get cleared after refactoring, I'm not 100% sure this couldn't happen now as well, but I think it might.

theCapypara:
also the CI failure is strange :thinking:

PR #868, opened 2024-01-19
Add Rust analyzer support

### LINKED ISSUE #520: Add Rust Language Server
https://rust-analyzer.github.io/

sonnyp:
Fixes #520 

Only supports syntax diagnostics for now. I made a follow up for types https://github.com/workbenchdev/Workbench/issues/881

## How to test

* Open Workbench
* From the Library, open a demo in Rust
* Make a syntax error in Rust code
* Wait a minute or so (rust-analyzer takes a while to initialize)

Diagnostics will appear for syntax errors. 

image

### DIFF
--- a/src/langs/rust/Compiler.js
+++ b/src/langs/rust/Compiler.js
@@ -2,10 +2,7 @@ import Gio from "gi://Gio";
 import GLib from "gi://GLib";
 import dbus_previewer from "../../Previewer/DBusPreviewer.js";
 import { copyDirectory, decode, encode } from "../../util.js";
-
-const rust_template_dir = Gio.File.new_for_path(
-  pkg.pkgdatadir,
-).resolve_relative_path("langs/rust/template");
+import { rust_template_dir } from "./rust.js";
 
 export default function Compiler({ session }) {
   const { file } = session;
--- a/src/langs/rust/rust.js
+++ b/src/langs/rust/rust.js
@@ -1,3 +1,5 @@
+import Gio from "gi://Gio";
+
 import { createLSPClient } from "../../common.js";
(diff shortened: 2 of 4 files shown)


src/common.js
```diff
@@ -94,6 +94,12 @@ export const languages = [
     document: null,
     default_file: "code.rs",
     index: 2,
+    // TODO: remove verbose
+    language_server: ["rust-analyzer", "--verbose"],
+    formatting_options: {
+      ...formatting_options,
+      tabSize: 2,
```
Hofer-Julian, line 101: We currently use rustfmt's defaults here, which specificies a tab size of 4.
Unless there's a good reason, we should stick with that.

src/langs/rust/RustDocument.js
```diff
@@ -1,36 +1,30 @@
...
 
-  const [success, stdout, stderr] = rustfmtProcess.communicate_utf8(text, null);
+  async format() {
+    // https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_formatting
+    const text_edits = await this.lspc.request("textDocument/formatting", {
+      textDocument: {
+        uri: this.file.get_uri(),
+      },
+      options: {
+        tabSize: 2,
```
Hofer-Julian, line 20: Same here

Hofer-Julian:
Considering that there's soon a release, I wonder if we should merge this PR (after a small clean up).

Showing syntax errors is already quite useful, and we could look into diagnostics in a follow-up PR.

lw64 (approved):

Hofer-Julian (approved):
Works as expected 🦀
Great work @sonnyp!

PR #877, opened 2024-01-31
Use AdwWindow for the previewer 

sonnyp:
I was working on updating demos for GNOME 46, particularly AboutWindow -> AboutDialog when I realized the previewer needs to be an AdwWindow instead of a GtkWindow if we want to support Adw.Dialog

https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.Dialog.html

> AdwDialog is similar to a window, but is shown within another window. It can be used with [AdwWindow](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.Window.html) and [AdwApplicationWindow](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.ApplicationWindow.html), use [adw_dialog_present()](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/method.Dialog.present.html) to show it.

---

I made a C WorkbenchPreview to experiment with sharing more code between previewers.

It can also be used in JS when we offer previewing in a separate window or process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,10 +26,10 @@
   "lint-staged": {
     "*.{json,md,yaml,yml}": "prettier --write",
     "*.{js,cjs,mjs}": "eslint --fix",
-    "*.css": "./build-aux/fun workbench-cli format",
+    "*.css": "./build-aux/fun workbench-cli format css",
     "*.py": "./build-aux/fun black",
     "*.rs": "./build-aux/fun rustfmt --edition 2021",
-    "*.blp": "./build-aux/fun workbench-cli format",
+    "*.blp": "./build-aux/fun workbench-cli format blueprint",
     "*.vala": "./build-aux/fun workbench-cli format vala"
   }
 }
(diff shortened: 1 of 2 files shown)


andyholmes (approved):
Nice one!

src/langs/python/python-previewer.py
```diff
@@ -22,15 +22,20 @@
...
-from gi.repository import GLib, Gdk, Gtk, Adw, Graphene, Gio, Gsk, GtkSource
+from gi.repository import GLib, Gdk, Gtk, Adw, Graphene, Gio, GtkSource, Workbench
 from gi.repository.Gio import DBusConnection, DBusConnectionFlags
 
 # Make sure all libraries are properly initialized
 Gtk.init()
 Adw.init()
 GtkSource.init()
 
+resource = Gio.Resource.load(
```
theCapypara, line 35: This would break any non-Flatpak installed version of Workbench. I know this isn't really supported, but I think we should probably still try to not "intentionally" break that. I pass in some other file paths as parameters, maybe that could be done here as well?

theCapypara (commented):
Nice! In general LGTM, just have one remark.

PR #878, opened 2024-02-04
Update Rust deps

sonnyp:
Depends on https://github.com/dbus2/zbus/issues/569

Allow using AdwDialog
I wanted to update the other ones but it's not working yet. I will wait for beta when things should stabilize.

### DIFF
--- a/src/application.js
+++ b/src/application.js
@@ -1,6 +1,5 @@
 import Adw from "gi://Adw";
 import Gio from "gi://Gio";
-import GLib from "gi://GLib";
 
 import Window from "./window.js";
 import Actions from "./actions.js";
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
Now that the zbus issue is closed, can this be merged?

src/langs/rust/template/Cargo.toml
```diff
@@ -9,27 +9,32 @@ edition = "2021"
...
+version = "0.8.0"
 package = "gtk4"
 features = ["gnome_45"]
 
 [dependencies.adw]
-version = "0.5"
+version = "0.6.0"
 package = "libadwaita"
-features = ["v1_4", "gtk_v4_10"]
+features = ["v1_5"]
```
Hofer-Julian, line 31: Why did you remove the gtk feature? Pretty sure that's needed

Hofer-Julian: By adding the feature we enable certain gated functions. I am not 100% sure that we are using them, but there's no harm enabling the feature since we have a new enough gtk and people using Workbench might want to use that API.

Bumping gtk fails since the latest feature depends only on gtk 4.10.

PR #879, opened 2024-02-09
Python Previewer: Initialize WebView

theCapypara:
`WebView` has no `init` function, which is odd. This might be an upstream bug after all, but I'm really not sure. 
This works at least and will continue to work, even if this is "fixed" upstream.


theCapypara:
Sure! Although the question is probably, what types should we ensure then? For GtkSource for example it's probably simple `Gtk.SourceView`.

theCapypara:
It's probably still a bit overkill to ensure all types of a library or even a big subset for some of them. I would asume all the libs that have an `init` to also make sure the types are properly registered, so personally I think using `ensure_type(s)` only for libs without an `init` is probably fine.

PR #880, opened 2024-02-10
Support adw dialog preview

sonnyp:
* Fix usage of AdwDialog in Rust
  * Shoud have been part of https://github.com/workbenchdev/Workbench/pull/877
  * Still needs https://github.com/workbenchdev/Workbench/pull/878
* Prevent preview of AdwDialog to avoid crashes


Hofer-Julian:
Looks good

PR #883, opened 2024-02-14
Support analysis / types with Rust Analyzer

### LINKED ISSUE #881: Support analysis / types in Rust LSP
In https://github.com/workbenchdev/Workbench/pull/868 we added rust-analyzer LSP support.

For some reason it only does syntax diagnostics, not types.

Hofer-Julian:
Fixes #881
It still takes a bit until diagnostics arrive, but this way error messages from `cargo` are included as well.

Also moves the target dir to a common place, similar to how we do it when we compile Rust.


sonnyp (commented):
I'm still only getting syntax diagnostics.

Anything specific I need to do?

image

image

Hofer-Julian:
> I'm still only getting syntax diagnostics.
> 
> Anything specific I need to do?
> 
> image
> 
> image

I shouldn't be necessary to do anything special, but I also noticed that it doesn't always work on my machine.

Before merging, we should also look out if this PR introduces regressions. If yes, maybe we just merge a subset of it.

PR #895, opened 2024-02-21
docs: Extend contributing section

Hofer-Julian:
`libxmljs` one of our nodejs dependencies requires a C++ compiler


Hofer-Julian:
> huh thanks
> 
> I should drop that dependency eventually

The behavior is a bit weird.

It complains that there's no binary available and falls back to compiling.

PR #916, opened 2024-03-23
Drop LLVM

sonnyp:
Fixes https://github.com/workbenchdev/Workbench/issues/611

I tried to make mold work with gcc with `-C link-arg=-fuse-ld=mold -C link-arg=-B=/usr/lib/sdk/rust-stable/` but it resulted in `note: collect2: fatal error: cannot find 'ld'`

Anyway, I can't see a difference with our without Mold - could it be the gnu linker is substantially faster now?

@Hofer-Julian wdyt?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ SHELL:=/bin/bash -O globstar
 setup:
 	flatpak remote-add --user --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo
 # flatpak remote-add --user --if-not-exists flathub-beta https://flathub.org/beta-repo/flathub-beta.flatpakrepo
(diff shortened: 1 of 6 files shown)


Hofer-Julian:
> I tried to make mold work

Mold stopped working?

> I can't see a difference with our without Mold

You mostly notice see the difference with incremental builds. 

>  could it be the gnu linker is substantially faster now?

Not that I would know

Hofer-Julian:
On my machine, I get the following results:
1.5 seconds for "Web View" with Mold
4 seconds for "Web View" with Mold

A big difference in relative numbers, but both compile times are very acceptable.
If that makes your lives easier, I am fine with dropping Mold and LLVM.

Hofer-Julian (approved):

PR #917, opened 2024-03-24
Enable Code Search

sonnyp:


BharatAtbrat:
All shortcuts are working fine, the close search button is a good addition

One small issue I noticed is that whenever closing search from the close button, the editor focus is lost. I need to click inside the editor to use shortcuts again

Here's a screengrab of the same: 

https://github.com/workbenchdev/Workbench/assets/72352158/9cea509d-b9c7-47a4-8baf-3a9b52fe154f

Pressing the Esc key closes the search but keeps the editor in focus, as before. This is probably the ideal behavior

Otherwise, Looks good to me :)

PR #921, opened 2024-03-27
Use Reverse-DNS name for developer ID

### LINKED ISSUE #906: Reverse developer ID
Appstream decided to use reverse DNS for developer IDs.

> The element should have a id property, containing a unique ID to identify the component developer / development team. It is recommended to use a reverse-DNS name, like org.gnome or io.github.ximion, or a Fediverse handle (like @user@example.org) as ID to achieve a higher chance of uniqueness.

More information: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer

You may want to change developer ID. 

https://github.com/workbenchdev/Workbench/blob/main/data/app.metainfo.xml#L6

vixalien:
fix #906 

(I finally got Workbench to build & run locally, yay!)

### DIFF
--- a/data/app.metainfo.xml
+++ b/data/app.metainfo.xml
@@ -3,7 +3,7 @@
   <id>@app_id@</id>
   <launchable type="desktop-id">@app_id@.desktop</launchable>
   <name translatable="no">Workbench</name>
-  <developer id="sonny.re">
+  <developer id="re.sonny">
     <name translatable="no">Sonny Piers</name>
   </developer>
   <update_contact>sonnyp@gnome.org</update_contact>


vixalien:
> What were the challenges to build and run Workbench locally?

None major, with flatpak I don't need to install any dependencies manually.

The only issue was that I was building the app in a cafe, the internet was slow and Workbench had to download lots of dependencies. Even checking out the repo with git without `--depth=1` was a challenge.

Secondly, It seems Workbench built some rust modules because my laptop was struggling. As much as I love Rust, I sure don't like compiling Rust projects.

Also, sometimes the CDN flathub uses hates me, so I had to use a VPN otherwise I was getting massively throttled.

PR #937, opened 2024-04-28
about: Remove translation support

yakushabb:
Workbench intentionally untranslatable.

More information: https://github.com/workbenchdev/Workbench/pull/662#issuecomment-1793727514

### DIFF
--- a/src/about.js
+++ b/src/about.js
@@ -1,5 +1,4 @@
 import Gtk from "gi://Gtk";
-import { gettext as _ } from "gettext";
 import GLib from "gi://GLib";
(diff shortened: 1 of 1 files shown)


yakushabb:
Other lines: 
- https://github.com/workbenchdev/Workbench/blob/main/src/sessions.js#L5
- https://github.com/workbenchdev/Workbench/blob/main/src/Previewer/Previewer.js#L6
- https://github.com/workbenchdev/Workbench/blob/main/src/Previewer/Previewer.js#L481-L484

PR #938, opened 2024-04-29
Add base TypeScript view to Workbench

vixalien:
This PR is just #933. I had trouble with git and had to create a new PR as Github automatically created a new one.

The issue was related to git not working correctly with "nested" git branch names, i.e. `wip/vixalien/typescript/add-to-view` couldn't coexist well with `wip/vixalien/typescript`, so I named the target branch just `typescript`.

### DIFF
--- a/build-aux/library.js
+++ b/build-aux/library.js
@@ -80,6 +80,9 @@ const demos = [];
     if (demo_dir.get_child("main.py").query_exists(null)) {
       languages.push("python");
     }
+    if (demo_dir.get_child("main.ts").query_exists(null)) {
+      languages.push("typescript");
+    }
     demo.languages = languages;
 
     await copyDemo(
--- a/build-aux/re.sonny.Workbench.json
+++ b/build-aux/re.sonny.Workbench.json
@@ -6,10 +6,12 @@
   "sdk-extensions": [
     "org.freedesktop.Sdk.Extension.vala",
     "org.freedesktop.Sdk.Extension.rust-stable",
-    "org.freedesktop.Sdk.Extension.llvm16"
+    "org.freedesktop.Sdk.Extension.llvm16",
+    "org.freedesktop.Sdk.Extension.node18",
+    "org.freedesktop.Sdk.Extension.typescript"
   ],
   "build-options": {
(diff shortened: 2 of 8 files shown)


vixalien:
Oh nice. I'll implement that too here

vixalien:
> Looks good, one thing missing is isTypeScriptEnabled.

Implemented

vixalien:
> Please address my review comments from the previous PR and try to be more thorough :)

Sorry, I had fixed them in the previous PR but it seems they didn't get cherry picked correctly.

vixalien:
The [last suggestion](https://github.com/workbenchdev/Workbench/pull/933#discussion_r1573219314) can't be applied because in the TypeScript block we check for `isTypeScriptEnabled`

vixalien:
> I'm not following. If TypeScript is not enabled that code block won't be reached - that's ok. The important thing for me is to deduplicate the code.


I'm sorry. The change

vixalien:
> I'm not following. If TypeScript is not enabled that code block won't be reached - that's ok. The important thing for me is to deduplicate the code.

https://github.com/vixalien/Workbench/blob/b2dc1df6f3e860cfbaad7ff0263702f39d93342d/src/window.js#L232C5-L238C6

I'm deeply sorry. The changes above were applied last week, and I missed them when rebasing.

> There are a bunch of unrelated changes in this PR such as submodules updates and code left untouched reformatted please revert. You can do for example git checkout main -- path-to-the-file to checkout a file in its version on main.

I have reverted the submodules changes.

> Regarding formatter, we use prettier. Maybe your editor supports it, if not we have a git pre-commit hook that should reformat commited files. Run make install please. And prettier --write src/**/*.js to make sure everything is formatted properly.

My editor does support prettier but the pre-commit hook wasn't working because `make setup` didn't work. I use pmOS/alpine so I'm thinking the `make setup` script has only been tested on glibc systems. I can create an issue if you like.

But currently, I believe everything is fixed so you can take another look if you'd like.

vixalien:
update: `make install` works when I use node v18 (not later). Maybe this should be included in the CONTRIBUTING.md file or a `.node-version` file could be added.

PR #953, opened 2024-06-28
Fix gst-plugin-gtk4 not being picked up

vixalien:
Not sure what's changed, but `gtk4paintablesink` doesn't get picked up automatically anymore. This PR makes GStreamer able to see it.

Without this, these demos are currently broken.

- Camera
- Screencast

### DIFF
--- a/build-aux/re.sonny.Workbench.Devel.json
+++ b/build-aux/re.sonny.Workbench.Devel.json
@@ -22,7 +22,8 @@
     "--socket=wayland",
     "--device=dri",
     "--share=network",
-    "--socket=pulseaudio"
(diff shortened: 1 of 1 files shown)


vixalien:
Update: This issue was caused by a `cargo-c` update, causing it to change the default libdir.

[A PR has been made](https://github.com/lu-zero/cargo-c/pull/386) to fix the `cargo-c` issue and I think we should wait to see the status of that PR (as I believe the demos are only broken on `main`)

sonnyp:
I can't reproduce with https://github.com/workbenchdev/Workbench/pull/982

Can you reopen if it still happens for you?

Issue #846, opened 2023-12-16
Rust projects cannot be run


sonnyp:
There is a regression in xdg-document-portal 1.18.2
https://github.com/flatpak/xdg-desktop-portal/issues/1234

It prevents Rust projects (the ones saved somewhere on the disk) from being Run.

Downgrading to xdg-document-portal 1.18.0 solves it.
An other option is to give filesystem access to Workbench with Flatseal or

```sh
cd my-cool-project
flatpak override --user --filesystem=$PWD re.sonny.Workbench
# restart Workbench
```

azzamsa:
I get this error when running the demo:

```bash
   Compiling proc-macro-error v1.0.4
error: linker `clang` not found
  |
  = note: No such file or directory (os error 2)

error: could not compile `slab` (build script) due to 1 previous error
```

I did:

```bash
🦄 pwd
~/.var/app/re.sonny.Workbench/data/re.sonny.Workbench/sessions/2026-01-31 03-26-55

🦄 flatpak override --user --share=network --socket=pulseaudio --device=input re.sonny.Workbench

🙊 flatpak install flathub org.freedesktop.Sdk.Extension.rust-stable//25.08 org.freedesktop.Sdk.Extension.llvm21//25.08
Looking for matches…
Skipping: org.freedesktop.Sdk.Extension.rust-stable/x86_64/25.08 is already installed
Skipping: org.freedesktop.Sdk.Extension.llvm21/x86_64/25.08 is already installed
```

Based on 
- https://github.com/flathub/org.freedesktop.Sdk.Extension.rust-stable#with-clang
- https://github.com/flathub/re.sonny.Workbench/blob/master/re.sonny.Workbench.json

```
re.sonny.Workbench 49.0
org.gnome.Platform 49

GJS 1.86.0
Adw 1.8.3
GTK 4.20.3
GLib 2.86.3
Flatpak 1.16.2
Vala 0.56.18
```

```
🦄 systeminfo
OS: Debian GNU/Linux 13 (trixie) x86_64
DE: GNOME 48.7
WM: Mutter (Wayland)
```

Am I missing something?

tristianc:
> I get this error when running the demo:
> 
>    Compiling proc-macro-error v1.0.4
> error: linker `clang` not found
>   |
>   = note: No such file or directory (os error 2)
> 
> error: could not compile `slab` (build script) due to 1 previous error
> 
> I did:
> 
> 🦄 pwd
[...]

PR #969, opened 2024-07-12
vala: Port Vala to Meson

### LINKED ISSUE #913: Vala diagnostics unknown symbols for Gtk and Workbench
Pretty sure this is new in Workbench 46

@Diego-Ivan @lw64 can one of you take a look?

image

### LINKED ISSUE #964: Consider porting Vala to Meson
Vala currently uses scripts to compile and run, which has been working great so far, but it comes with a couple of limitations:

* Vala users must specify the dependencies that they want to use in the script command. Misspelling them would cause the language server to break
* When writing demo in Vala from the ground up, shebang is not included by default, and the users must copy/paste it from other demos. The language server does not read the shebang when it is added immediately, so the code would be filled with false errors until it catches up.
* Adding/removing a dependency from the shebang is not caught by the language server immediately, which leads to false errors.

Using Meson as build system would require choosing the dependencies beforehand, just as Rust does, saving the user the need to specify them themselves. This comes at the cost that the users would not be able to add other dependencies that may have available `vapi` files, but this is something Rust has too and does not seem to be a problem. 

Adding new demos that require new dependencies would force us to keep the build systems updated too, but I think Workbench has gotten to a point where pretty much all the libraries used in GNOME projects have been covered, so this will not occur frequently.

If porting Vala to Meson is viable, I can start working on this issue :)

CC @lw64

Diego-Ivan:
This PR ports Vala from scripts to Meson, and introduces a minimal Vala template with the Workbench API file and a meson build file.

Fixes https://github.com/workbenchdev/Workbench/issues/964
Fixes #913


sonnyp (commented):
Thanks for working on this!

One thing to keep in mind is we need to keep compiling existing Vala project so please make sure that's working

I will let @lw64 do the first review

Diego-Ivan:
@sonnyp Existing sessions and all the demos work as expected. Projects don't, because meson won't accept the file paths given by a portal. I tested this with Rust/cargo and it happens too.

A possible solution might be copying the Vala sources to a temporary folder that meson accepts (like the cache directory) and compile it there. The source files are usually small so it shouldn't be expensive. What do you think?

Diego-Ivan:
Same as #846 unfortunately :(

sonnyp:
> Same as https://github.com/workbenchdev/Workbench/issues/846 unfortunately :(

I see

> A possible solution might be copying the Vala sources to a temporary folder that meson accepts (like the cache directory) and compile it there. The source files are usually small so it shouldn't be expensive. What do you think?

Sounds interesting, perhaps we can solve https://github.com/workbenchdev/Workbench/issues/846 the same way if it turns out to be a good idea.

I'm also interested to avoid having a `builddir` in the project directory. 

Wanna give it a try?

sonnyp:
I tested this and besides projects it seems to work great. I also merged main into your branch and fixed conflicts.

Diego-Ivan:
> I'm also interested to avoid having a builddir in the project directory.

I think it would be better too, since Meson's builddirs can break between major versions, and old projects may break as a consequence of this.

I'll give it a try this week :)

Diego-Ivan:
I think this PR is ready! A little summary of the changes:

* Vala projects will be built in the `/tmp/vala` directory, instead of the working directory of the session.
* The `/tmp/vala` folder will contain all the files necessary to compile a session. Every time _Run_ is clicked, the `main.vala` file in the tmp folder will be replaced by the demo's/project's `main.vala`. Probably we can do this with Rust too, so we don't have to grant filesystem permissions to compile projects.

Also updated the newer demos in https://github.com/workbenchdev/demos/pull/187, tested locally and CI passed, so this is ready for review.

src/cli/main.js
```diff
@@ -352,16 +352,30 @@ async function ci({ filenames }) {
...
+      const current_dir = Gio.File.new_for_path(GLib.get_current_dir());
+
+      const template_dir = GLib.getenv("FLATPAK_ID")
+        ? Gio.File.new_for_path(
+            `/app/share/${GLib.getenv("FLATPAK_ID")}/langs/vala/template`,
+          )
+        : current_dir.resolve_relative_path("src/langs/vala/template");
+
+      const api_file = template_dir.get_child("workbench.vala");
+      api_file.copy(
```
sonnyp, line 364:
```suggestion
      const template_dir = Gio.File.new_for_path(pkg.pkgdatadir).get_child("langs/vala/template");

      template_dir.get_child("workbench.vala").copy(
```

src/langs/vala/Compiler.js
```diff
@@ -60,11 +90,3 @@ export default function ValaCompiler({ session }) {
...
   return { compile, run };
 }
-
-// Takes a string starting with the line
-// #!/usr/bin/env -S vala workbench.vala --pkg gtk4 --pkg libadwaita-1
-// and return ["--pkg", "gtk4", "--pkg", "libadwaita-1"]
-// FIXME: consider using https://docs.gtk.org/glib/struct.OptionContext.html instead
-function getValaCompilerArguments(text) {
-  return text.split("\n")[0]?.split("-S vala ")[1]?.split(" ") || [];
-}
```
sonnyp, line 70: :tada:

src/langs/vala/Compiler.js
```diff
@@ -1,49 +1,79 @@
...
-      "-X",
-      "-fpic",
-      "--library",
-      "workbench",
-      "-o",
-      module_file.get_path(),
-      "--vapi",
-      "/dev/null",
-      ...args,
+    const meson_clean = meson_launcher.spawnv([
```
sonnyp, line 51: why cleaning?

src/langs/vala/Compiler.js
```diff
@@ -1,49 +1,79 @@
 import Gio from "gi://Gio";
 import dbus_previewer from "../../Previewer/DBusPreviewer.js";
-import { decode } from "../../util.js";
+import { createTmpDir, copy } from "../../util.js";
+import { setupValaProject } from "./vala.js";
+
+let ready_to_build = false;
```
sonnyp, line 6: I'm confused about why this is needed

src/langs/vala/template/meson.build
```diff
@@ -0,0 +1,39 @@
...
+dependencies = [
+    m_dep,
+    dependency('libadwaita-1'),
+    dependency('shumate-1.0'),
+    dependency('libportal-gtk4'),
+    dependency('libsoup-3.0'),
+    dependency('webkitgtk-6.0'),
+    dependency('gstreamer-1.0'),
+    dependency('gtksourceview-5'),
+    dependency('json-glib-1.0'),
```
sonnyp, line 33:
```suggestion
    dependency('libadwaita-1'),
    dependency('shumate-1.0'),
    dependency('libportal-gtk4'),
    dependency('libsoup-3.0'),
    dependency('webkitgtk-6.0'),
    dependency('gstreamer-1.0'),
    dependency('gtksourceview-5'),
    dependency('json-glib-1.0'),
    dependency('libspelling-1'),
```

src/langs/vala/vala.js
```diff
@@ -2,22 +2,16 @@ import Gio from "gi://Gio";
...
-  );
-  api_file.copy(
-    file.get_parent().get_child("workbench.vala"),
-    Gio.FileCopyFlags.OVERWRITE,
-    null,
-    null,
-  );
+  // VLS needs the project to be already setup once it starts,
+  // otherwise it won't pick it up later.
+  setupValaProject(file.get_parent()).catch(console.error);
```
sonnyp, line 14: is it possible to move the meson and workbench.vala files under a subdir? to avoid cluttering the project

and we might need `meson.build` for other purposes too

src/util.js
```diff
@@ -211,3 +211,25 @@ export async function copy(filename, source_dir, dest_dir, flags) {
 
   return dest_file;
 }
+
+const tmp_dirs = {};
+
+export async function createTmpDir(filename) {
```
sonnyp, line 217: checkout 

* https://gjs-docs.gnome.org/gio20~2.0/gio.file#function-new_tmp
* https://docs.gtk.org/glib/type_func.Dir.make_tmp.html

src/langs/vala/Compiler.js
Diego-Ivan: When replacing the file in the build dir, meson will not detect any changes on the file, thus not compiling anything and just running the project right away. This is a problem when a user has two open projects and working on them simultaneously, so I went for cleaning the build files and recompiling. I thought this approach was OK since we don't really compile much.

src/langs/vala/Compiler.js
Diego-Ivan: I did this to know if the Vala builddir was already setup (as it needs a meson file and the workbench API). This could be done checking if a build file (like `workbench.vala`) exists, but I went for this instead as it is one less I/O operation.

src/langs/vala/vala.js
Diego-Ivan: Sure, I think this is possible. Would that subdir still be inside the project folder? Or do you mean saving *just* the `main.vala` in the project directory and get rid of all of the build stuff? I'm guessing this last option is also possible.

src/util.js
Diego-Ivan: Thanks, I'll check them out!

PR #981, opened 2024-08-13
Add dependency for GOM demo

BharatAtbrat:
- Parent PR: https://github.com/workbenchdev/demos/pull/200


sonnyp (dismissed):
Please move to a file in `modules/gom.json` and use the http archive
 like other dependencies such as https://github.com/workbenchdev/Workbench/blob/main/build-aux/modules/libshumate.json

andyholmes (approved):

PR #982, opened 2024-09-20
GNOME 47

sonnyp:
Technically depends on https://github.com/flathub/org.freedesktop.Sdk.Extension.typescript/issues/29

Without, TypeScript support is broken.
But I'll merge anyway as keeping separate sdk branches is a mess, and we can move forward and prepare the release.

For now I have re-enabled Biome for JavaScript. 

Hopefully `org.freedesktop.Sdk.Extension.typescript//24.08` will get released soon.


andyholmes:
FYI, I pinged Chris about the TypeScript extension and he said he should get to it this week.

PR #985, opened 2024-09-26
Remove unsafe permissions and add dialog

### LINKED ISSUE #915: Inform users when a permission is required for a demo
Currently, Workbench is considered as "probably safe" by Flathub and GNOME Software.

The reason it is not considered "Safe" is because we require network and audio permission (via `--socket=pulseaudio` and `--share=network`. Those are not strictly required to use Workbench and are only needed for a handful of demos. In addition, we would also like to add a permission for gamepads https://github.com/workbenchdev/demos/pull/88

Eventually there will be portals for all of these things but until then we need static permissions.

Let's

1. remove those permissions so that Workbench is considered safe 
2. add a mechanism for demos to advertise which permissions they need - maybe `flatpak-finish-args: ["--socket=fallback-x11"]` in `main.json`
3. when a needed permission is not enabled, Workbench should show a dialog on "Run" to explain what this is about and how to enable the permission using `flatpak overrides`

Since overrides are permanent - it should be a one time thing only anyway.

sonnyp:
Fixes https://github.com/workbenchdev/Workbench/issues/915

image


andyholmes (approved):
Pretty straight-forward, LGTM!

EDIT: except the lint job I guess :sweat_smile:

bragefuglseth:
I'd:

- use `.card` styling for the command box
- use a boxed list for the permission explanations
- change the permission explanation section title to "What it Does" for the sake of simplicity

Otherwise this LGTM as well!

PR #986, opened 2024-10-03
Disable TypeScript by default

sonnyp:
Temporarily until https://github.com/flathub/org.freedesktop.Sdk.Extension.typescript/issues/29


vixalien:
Makes sense to me.

PR #988, opened 2024-10-17
Include freedesktop sdk TypeScript extension

### LINKED ISSUE #975: JavaScript diagnostics for CI
Currently `./build-aux/fun workbench-cli ci demos/src/*` doesn't do type checks for JAvaScript file because we didn't update the CLI to use typescript-language-server

sonnyp:
Closes https://github.com/workbenchdev/Workbench/issues/975

TypeScript is still disabled by default for now but it's just a gsetting.

### DIFF
--- a/build-aux/re.sonny.Workbench.json
+++ b/build-aux/re.sonny.Workbench.json
@@ -8,10 +8,11 @@
     "org.freedesktop.Sdk.Extension.vala",
     "org.freedesktop.Sdk.Extension.rust-stable",
     "org.freedesktop.Sdk.Extension.llvm18",
-    "org.freedesktop.Sdk.Extension.node20"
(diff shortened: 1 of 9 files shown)


build-aux/re.sonny.Workbench.json
vixalien, line 1: Sorry but why even do you have the non-Devel manifest here 🤔 it probably should only exist in the flathub repo

vixalien (commented):
didn't test the PR to see if it actually works but looks good for now

build-aux/re.sonny.Workbench.json
sonnyp: Occasional testing when they diverge. 

The manifest in this repo is the source of truth for flathub. I copy it over to flathub when I release.

PR #991, opened 2024-11-12
Release 47

sonnyp:


data/app.gschema.xml
```diff
@@ -19,7 +19,7 @@
       <default>true</default>
     </key>
     <key name="typescript" type="b">
-      <default>false</default>
+      <default>true</default>
```
sonnyp, line 22: @vixalien I enable TypeScript by default because it's quite unintrusive currently and marked as "experimental support" in the release notes for now

vixalien: Sure. It will also allow faster iteration and user feedback

PR #992, opened 2024-12-11
Support Flatpak < 1.15.6 for permissions

sonnyp:


src/Permissions/Permissions.blp
```diff
@@ -42,7 +42,7 @@ Adw.Dialog dialog {
           }
 
           Label {
-            label: _("Workbench needs additional permissions. Please run the following command in a terminal and restart the app.");
+            label: _("Workbench needs additional permissions. Please run the following command in a terminal and restart Workbench.");
```
sonnyp, line 45: cc @bragefuglseth

bragefuglseth: LGTM

PR #997, opened 2025-01-16
Use CSS variables instead of GTK-specific syntax

UrtsiSantsi:
I didn't notice any visual differences after this, witch is a good sign :)


UrtsiSantsi:
@sonnyp the CI failure seems unrelated, although I don't really understand it :sweat_smile:

sonnyp (approved):
Thanks!

PR #1002, opened 2025-03-22
Fix TypeScript typechecking not working

vixalien:
In this PR, I:

- Updated the `gi-types`
- Correctly placed them into the correct path (at `src/langs/typescript/template/gi-types`
- Fixed the global `workbench` type not working

This is in reference to https://github.com/workbenchdev/Workbench/issues/870#issuecomment-2715083689

image

### DIFF
--- a/src/langs/typescript/types/ambient.d.ts
+++ b/src/langs/typescript/types/ambient.d.ts
@@ -1,50 +1,17 @@
-// import Adw from "gi://Adw";
-// import Gtk from "gi://Gtk?version=4.0";
+/// <reference path="../template/gi-types/index.d.ts" />
+
+import Adw from "gi://Adw";
+import Gtk from "gi://Gtk?version=4.0";
 // import GObject from "gi://GObject";
 
-// additional type declarations for GJS
-
-// additional GJS log utils
-declare function print(...args: any[]): void;
-declare function log(...args: any[]): void;
-
-// GJS pkg global
-declare const pkg: {
-  version: string;
-  name: string;
-};
-
-// old GJS global imports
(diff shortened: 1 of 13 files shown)


src/langs/typescript/types/ambient.d.ts
```diff
@@ -1,50 +1,17 @@
...
-    vprintf(fmt: string, args: any[]): string;
+declare global {
+  const workbench: {
+    window: Adw.ApplicationWindow;
+    application: Adw.Application;
+    builder: Gtk.Builder;
+    template: string;
+    resolve(path: string): string;
+    preview(object: Gtk.Widget): void;
+    // build(params: Record<string, Function | GObject.Object>): void;
```
vixalien: Not sure if this should be included or not. Please advise

sonnyp (commented):
Thanks

Isn't it a good time to generate accurate types ourselves?

>  Add a simple script that will generate the types for us. Probably similar to https://github.com/vixalien/gi-typescript-definitions/blob/main/generate_types_in_flatpak.sh

https://github.com/workbenchdev/demos/pull/201

https://floss.social/@jumplink/114218446354825670

vixalien:
We can. But I don't know if it's the best idea, the diffs will be huge.

sonnyp:
Why is that an issue?

sonnyp:
@vixalien ping :)

src/langs/typescript/types/ambient.d.ts
sonnyp:
```suggestion
    build(params: Record<string, Function | GObject.Object>): void;
```

PR #1003, opened 2025-03-22
fix: HoverProvider `get_iter` crashing

vixalien:
Ran into this while working on #1002.

Not sure what changed upstream, may need to check documentation.


sonnyp (approved):
Probably consequence of 

> Various introspection fixes

https://gitlab.gnome.org/GNOME/gtksourceview/-/blob/2db38b96df3934a1c7c703c6b3a0aae6c60fbb39/NEWS#L20

